PR #10, opened 2017-05-28
Assign runtime components into RuntimeLibraries install component

lassoan:
This allows adding Elastix shared libraries (.dll, .dylib, .so) to packages without including all .h and .lib files.


lassoan:
Thank you!

PR #24, opened 2017-08-31
ENH: TransformixFilter exposes deformation field object

FBerendsen:
Previously, when tranformix was asked to give deformation field it was written to disk directly (deformation.mhd), even when called from the library interface. Current pull request lets transformix pass its deformation field via memory.

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.h
+++ b/Core/ComponentBaseClasses/elxTransformBase.h
@@ -298,15 +298,18 @@ public:
   /** Function to transform coordinates from fixed to moving image, given as VTK file. */
   virtual void TransformPointsSomePointsVTK( const std::string filename ) const;
 
+  /** Deprecation note: The plan is to split all Compute* and TransformPoints* functions 
+   *  into Generate* and Write* functions, since that would facilitate a proper library 
+   *  interface. To keep everything functional during the transition period we need to 
+   *  keep the orignal Compute* and TransformPoints* functions and let them just call 
+   *  Generate* and Write*. These functions should be considered marked deprecated.
+
   /** Function to transform all coordinates from fixed to moving image. */
-  /** Legacy note: the function formerly called TransformPointsAllPoints was split into 
-      GenerateDeformationFieldImage for generating and WriteDeformationFieldImage for 
-      writing to disk*/
   typename DeformationFieldImageType::Pointer GenerateDeformationFieldImage( void ) const;
 
   void WriteDeformationFieldImage( typename DeformationFieldImageType::Pointer ) const;
 
-  /** Function to transform all coordinates from fixed to moving image. */
+  /** Legacy function that calls GenerateDeformationFieldImage and WriteDeformationFieldImage. */
   virtual void TransformPointsAllPoints(void) const;
 
   /** Function to compute the determinant of the spatial Jacobian. */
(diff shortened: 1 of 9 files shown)


Core/ComponentBaseClasses/elxTransformBase.h
```diff
@@ -192,6 +192,12 @@ class TransformBase :
   typedef typename ITKBaseType::InputPointType  InputPointType;
   typedef typename ITKBaseType::OutputPointType OutputPointType;
 
+  /** Typedef's for TransformPointsAllPoints. */
+  typedef itk::Vector<
+    float, FixedImageDimension >                      VectorPixelType;
```
kaspermarstal: Is the coord type guaranteed to be float always? @stefanklein

Core/Main/elxTransformixFilter.hxx
```diff
@@ -248,7 +249,7 @@ TransformixFilter< TMovingImage >
     // Primary and all other outputs default to ResultImage.
     return TMovingImage::New().GetPointer();
   }
-}
+} // end MakeOutput()
 
 /**
  * ********************* SetInput *********************
```
kaspermarstal: `SetInput` should be `GetOutputDeformationField`

Core/ComponentBaseClasses/elxTransformBase.h
```diff
@@ -293,7 +299,15 @@ class TransformBase :
   virtual void TransformPointsSomePointsVTK( const std::string filename ) const;
 
   /** Function to transform all coordinates from fixed to moving image. */
-  virtual void TransformPointsAllPoints( void ) const;
+  /** Legacy note: the function formerly called TransformPointsAllPoints was split into 
+      GenerateDeformationFieldImage for generating and WriteDeformationFieldImage for 
+      writing to disk*/
```
kaspermarstal: Not sure we need to remind people of this. Sometimes it is just more confusing to keep legacy knowledge around.

Core/ComponentBaseClasses/elxTransformBase.h
FBerendsen: Good remark, I could have chosen double as well. I didn't check what is most in line with elastix. Preferably I would make it a template argument, but that far from trivial in the elastix framework. Other constructions would get complicated as well, so I think we just have to choose between float or double. What is your preference?

Core/ComponentBaseClasses/elxTransformBase.h
FBerendsen: Ok, the comment could be a bit clearer. Maybe I should write the bigger plan here, which points out what and why things are legacy and is important to know for developers.
My plan is to split all Compute* functions (and others such as TransformPointsAllPoints) into Generate* and Write* functions, since that would facilitate a proper library interface. To keep everything functional during the transition period we need to keep the Compute* functions and make them just call Generate* and Write*, but mark them as legacy/deprecated.

kaspermarstal: Sounds reasonable. Do you think we should merge this now as is or do you want to wait until the splitting of Compute* etc is done?

Core/ComponentBaseClasses/elxTransformBase.h
kaspermarstal: I think float is good enough since it has more than enough precision (1mm x 10^-7 ~ 1nm :D) and takes up half as much ram :)

Core/ComponentBaseClasses/elxTransformBase.h
FBerendsen: I would like to go for a quick merge and only implement (and test) the things I need most right now. The rest will follow, hopefully. I do like to put the bigger plan as a comment in the code. For this I will make a new commit including the resolutions of other remarks.

kaspermarstal: Great! I completely agree.

PR #39, opened 2018-01-02
Elx 34 remove testing external dependencies

mstaring:


kaspermarstal:
There are still two tests failing. Are you planning on fixing these or we disable these temporarily?

The travis failures looks more related to an error on travis' part.

PR #54, opened 2018-03-23
BUG: TransformixFilter GenerateOuputInformation

FBerendsen:
needed to be overridden to use image properties from TransformParameterMap instead of MovingImage


FBerendsen:
Remark: Travis CI doesn't compile transformixFilter.hxx and thus hasn't found errors.
SuperElastix does use transformixFilter.hxx, and with the last commit its Jenkins CI gives green light!
https://github.com/SuperElastix/SuperElastix/commit/72a9a03a8f48fff7719aa62f840e274b5ce2746b

Who will merge this?

mstaring:
I think I should have done a squash merge, but alas

PR #86, opened 2018-09-14
WIP: allow building with ITK_FUTURE_LEGACY_REMOVE=ON. Ref #85.

dzenanz:
This is work in progress. It would be good if someone could take over.

Issue #85, opened 2018-09-14
Build with ITK_FUTURE_LEGACY_REMOVE=ON


dzenanz:
If ITK 4.13.1 is built with option `ITK_FUTURE_LEGACY_REMOVE` set to `ON`, _elastix_ fails to build. The first error, and many others are related to `purposeful_error("itkGetObjectMacro should be replaced with itkGetModifiableObjectMacro.")`. I suspect that it would not take a lot of effort to accomplish this.

mstaring:
Can this be closed @N-Dekker ?

N-Dekker:
@mstaring Yes, this can be closed! Hereby!

PR #95, opened 2019-01-11
COMP: Allow building with ITK_FUTURE_LEGACY_REMOVE=ON, fixing issue #85

N-Dekker:
Partially redoes the work of pull request #86 by @dzenanz (sorry Dzenan, but thanks!)
Fixes elastix issue #85: "Build with ITK_FUTURE_LEGACY_REMOVE=ON", from @dzenanz


thewtex:
Thanks @N-Dekker !

mstaring:
So, we are using a lot of modifiable right now. These things could not be fixed using the const version of the functions?

dzenanz:
Some of them can, but others require refactoring besides simply changing `Modifiable` into `const`.

PR #98, opened 2019-01-14
COMP: Added compilation defines to use vnl_math legacy macro's with ITK5

N-Dekker:
Added command-line compilation defines to use `vnl_math` macro's for `abs`, `max`, `min`, `rnd`, etc., when ITK >= 5.

These macro's are also defined in "vxl\vcl\vcl_legacy_aliases.h"


mstaring:
This is code that can be removed later, once we decide to drop legacy support?

For me these can go sooner rather than later. I think we can make our next elastix release only compile with ITK 5, and nothing earlier.

PR #99, opened 2019-01-15
COMP: Fix elxElastixMain ITK5 errors, trying to use 0 for SmartPointer

N-Dekker:
Fixes compilation errors when using ITK 5 (from ITK git master branch):

    elxElastixMain.cxx(164): error C2440: 'initializing': cannot convert from 'int' to 'itk::SmartPointer<elastix::ComponentDatabase::Self>'
    elxElastixMain.cxx(165): error C2440: 'initializing': cannot convert from 'int' to 'itk::SmartPointer<elastix::ComponentLoader::Self>'
    elxElastixMain.cxx(814): error C2440: 'initializing': cannot convert from 'int' to 'itk::SmartPointer<itk::Object::Self>'
    elxElastixMain.cxx(816): error C2446: ':': no conversion from 'int' to 'elastix::ComponentDatabase::ObjectPointer'

Follow-up to revision 49421b181302c75e22158e282254c30b44631b55 (2019-01-08), "COMP: Fix error w/ ITK5, trying to initialize itk::SmartPointer by zero"

PR #100, opened 2019-01-16
ITK5 drops support 0 as smart pointer (part 3)

N-Dekker:
Three more commits regarding the use of 0 (zero) as smart pointer null value, which is no longer supported by ITK 5 (current ITK git master branch).

Follow-up to #94 and #99


N-Dekker:
Please don't merge! Another commit is underway!

PR #101, opened 2019-01-17
ENH: Remove dynamic exception specifications which are deprecated in C++11

kaspermarstal:


mstaring:
I can't really judge. THis is something we simply copied from the ITK previously. If this is removed in ITK5 now, then for me this is enough reason the also remove it in elastix. Is it?

kaspermarstal:
What does ITK do now?

N-Dekker:
@kaspermarstal 
> What does ITK do now?

As far as I know, nobody added explicit _potentially-throwing exception specifications_ (`noexcept(false)`) to ITK, so far.

kaspermarstal:
Shall we just stick to ITK style? Or do it the proper way? What do you think?

N-Dekker:
@kaspermarstal 
> Shall we just stick to ITK style? Or do it the proper way? What do you think?

I think: just merge! 

While it seems more proper to me to use `noexcept(false)`, It isn't common practice to do so.

mstaring:
please go ahead and merge then

PR #108, opened 2019-01-30
Initial commits to support ITK5 with USE_ALL_COMPONENTS

N-Dekker:
Initial commits to support ITK5 with elastix CMake option USE_ALL_COMPONENTS:

- Replace vcl macro calls by directly using std, in all components 
- Fix USE_ALL_COMPONENTS ITK5 SmartPointer compile errors 
- Added USE_ALL_COMPONENTS ITK5 C++11 vnl_math_cuberoot workaround


N-Dekker:
ITKv4

PR #112, opened 2019-02-08
COMP: Redefined DummyIPPPixelType as unsigned char

N-Dekker:
Originally `DummyIPPPixelType` was defined as `bool`, in `PolydataDummyPenalty::ReadTransformixPoints`, `MissingStructurePenalty::ReadTransformixPoints`, and `TransformBase::TransformPointsSomePoints`. This appears troublesome when using `DummyIPPPixelType` as `TElement` argument of `itk::VectorContainer<TElementIdentifier, TElement>`:

  "COMP: Fix support itk::VectorContainer<TElementIdentifier, bool>"
  InsightSoftwareConsortium/ITK#486

This commit defines `DummyIPPPixelType` as an `unsigned char` instead, in these three member functions.


N-Dekker:
For the record: The original commit (which redefined DummyIPPPixelType as empty struct) had a compile error from Ubuntu (1404, 1604) GCC:

     error: template argument for ‘itk::DefaultStaticMeshTraits’ uses local type

https://dev.azure.com/kaspermarstal/Elastix/_build/results?buildId=298

PR #117, opened 2019-02-27
COMP: fixed compilation with missing itksys/String.hxx include

dpshamonin:


N-Dekker:
For now, I think this fix is fine, but eventually, maybe we should just replace itksys::String by std::string.

Issue #21, opened 2017-08-08
Fix CMake 3.9 warnings


jcfr:
Configuring the extension using the latest CMake release reports the following warnings:

```
CMake Deprecation Warning at src/CMakeLists.txt:7 (cmake_policy):
  The OLD behavior for policy CMP0033 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


CMake Deprecation Warning at src/CMakeLists.txt:10 (cmake_policy):
  The OLD behavior for policy CMP0007 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


CMake Warning (dev) at /home/jcfr/Software/cmake-3.9.0-Linux-x86_64/share/cmake-3.9/Modules/FindOpenMP.cmake:200 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "c" will no longer be dereferenced when the policy is
  set to NEW.  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  /home/jcfr/Software/cmake-3.9.0-Linux-x86_64/share/cmake-3.9/Modules/FindOpenMP.cmake:324 (_OPENMP_GET_FLAGS)
  src/CMakeLists.txt:95 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

mstaring:
@N-Dekker would you have time to look at this?

N-Dekker:
For the record, I can reproduce the issue (warnings for policy CMP0033 and CMP0007) on CMake 3.10.2 (for generator Visual Studio 2015 Win64).

N-Dekker:
"SET CMP0033 OLD" allows using export_library_dependencies(): https://cmake.org/cmake/help/v3.10/policy/CMP0033.html
[...]

PR #123, opened 2019-03-08
CXX 11

kaspermarstal:


.gitignore
```diff
@@ -37,3 +37,4 @@ cmake*
 Makefile
 CMakeFiles/
 CTestTestfile.cmake
+.idea/
```
N-Dekker: 👍  I checked: SuperElastix does already ignore .idea files

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required( VERSION 2.8 )
 
 project( elastix )
+set(CMAKE_CXX_STANDARD 11)
```
N-Dekker: OK, but I hope it works well with `cmake_minimum_required( VERSION 2.8 )`. You see, `CMAKE_CXX_STANDARD` was introduced with CMake 3.1. But I haven't yet succeeded to upgrade `cmake_minimum_required` (issue #21).

Common/OpenCL/ITKimprovements/itkOpenCLContext.h
```diff
@@ -469,7 +469,7 @@ class ITKOpenCL_EXPORT OpenCLContext : public LightObject
 
   ITK_OPENCL_DECLARE_PRIVATE( OpenCLContext )
 
-  std::auto_ptr< OpenCLContextPimpl > d_ptr;
+  std::unique_ptr< OpenCLContextPimpl > d_ptr;
```
N-Dekker: 👍  to "ENH: Use unique_ptr instead of auto_ptr". As long as it compiles, it's fine, I guess.

Common/OpenCL/ITKimprovements/itkOpenCLGlobal.h
```diff
@@ -26,8 +26,8 @@ template< typename T >
 static inline T *
 OpenCLGetPtrHelper( T * ptr ) { return ptr; }
 template< typename TObjectType >
-static inline typename std::auto_ptr< TObjectType >::element_type * OpenCLGetPtrHelper(
-  const std::auto_ptr< TObjectType > &p )
+static inline typename std::unique_ptr< TObjectType >::element_type * OpenCLGetPtrHelper(
```
N-Dekker: Isn't `typename std::unique_ptr< TObjectType >::element_type` exactly the same as `TObjectType`? Anyway, your change from `auto_ptr` to k`unique_ptr' looks fine to me.

Testing/CMakeLists.txt
```diff
@@ -980,8 +980,10 @@ elx_add_run_test( 3DCT_lung.NC.bspline.ASGD.001d # manual estimation and no adap
   -t0 ${TestDataDir}/transformparameters.3DCT_lung.affine.txt
   -p ${TestDataDir}/parameters.3D.NC.bspline.ASGD.001d.txt )
 
-# Test transformix to check memory problem
-trx_add_test( TransformixMemoryTest
-  -in ${TestDataDir}/3DCT_lung_baseline_small.mha
-  -tp ${TestDataDir}/transformparameters.3DCT_lung.affine.txt )
+if(${ELASTIX_BUILD_EXECUTABLE})
```
N-Dekker: 👍  (but I haven't tested... I hope you did!)

Core/Main/elastixlib.cxx
```diff
@@ -46,9 +46,9 @@ namespace elastix
  * ******************* Constructor ***********************
  */
```
N-Dekker: I hate to say it but GitHub puts a hard line break in commit subjects of more 72 chars: "ENH: Initialize SmartPointers with nullptr instead of 0 which is ambi… \n …guous". Better keep the first line of a commit message at <= 72 chars, if possible.

N-Dekker: Sorry that I did not fix `elastixlib.cxx` before (as part of the upgrade to ITK5). I only tested ELASTIX_BUILD_EXECUTABLE = ON.

Core/Main/elastixlib.cxx
```diff
@@ -46,9 +46,9 @@ namespace elastix
...
 
-ELASTIX::ELASTIX() :
-  m_ResultImage( 0 )
+ELASTIX::ELASTIX()
 {
+  m_ResultImage = nullptr;
```
N-Dekker: I prefer to have the `nullptr` exactly there where it had the `0`: in the mem-initializer-list of the `ELASTIX()` constructor. With your commit,  `ELASTIX()` will first default-construct `m_ResultImage`, and then assign `nullptr` to  `m_ResultImage`. The original code directly constructed `m_ResultImage` with `0`, without an extra assignment.

Core/Main/elxElastixFilter.hxx
```diff
@@ -275,7 +275,7 @@ ElastixFilter< TFixedImage, TMovingImage >
   } // End loop over registrations
 
   // Save result image
-  if( resultImageContainer.IsNotNull() && resultImageContainer->Size() > 0 )
+  if( resultImageContainer.IsNotNull() && resultImageContainer->Size() > 0 && resultImageContainer->ElementAt( 0 ).IsNotNull() )
```
N-Dekker: Did any problem occur before you added the extra `resultImageContainer->ElementAt( 0 ).IsNotNull()` check?

Core/Main/elxElastixFilter.h
```diff
@@ -188,9 +188,6 @@ class ELASTIXLIB_API ElastixFilter : public itk::ImageSource< TFixedImage >
   /** RemoveInputsOfType. */
   void RemoveInputsOfType( const DataObjectIdentifierType & inputName );
 
-  /** Let elastix handle input verification internally */
-  virtual void VerifyInputInformation( void ) ITK_OVERRIDE {};
-
```
N-Dekker: My fault  `:-)`  I declared the ITK5 `VerifyPreconditions()` member functions `const`: https://github.com/InsightSoftwareConsortium/ITK/commit/3e6b6f5bd6772316aa0fa6b89f31b42bef562112

Of course, if these functions are empty anyway, for `elastix`, it's fine to just remove them. 👍

Core/Main/elxTransformixFilter.h
```diff
@@ -153,10 +153,7 @@ class ELASTIXLIB_API TransformixFilter : public itk::ImageSource< TMovingImage >
   void operator=( const Self & );    // purposely not implemented
 
   /** IsEmpty. */
-  virtual bool IsEmpty( const InputImagePointer inputImage );
-
-  /** Let transformix handle input verification internally. */
-  virtual void VerifyInputInformation( void ) ITK_OVERRIDE {};
+  bool IsEmpty( const InputImagePointer inputImage );
```
N-Dekker: Changing the declaration of `IsEmpty` seems unrelated to the rest of the commit, but if you want to change it, maybe you could even declare it `static`!

Core/Main/elxParameterObject.cxx
```diff
@@ -23,6 +23,7 @@
 #include "itkFileTools.h"
 #include <fstream>
 #include <iostream>
+#include <math.h>
```
N-Dekker: Instead of `#include <math.h>`, I would expect `#include <string>`!

Core/Main/elxParameterObject.h
```diff
@@ -89,17 +89,6 @@ class ParameterObject : public itk::DataObject
...
-  // C++11 has a to_string in the standard library, but we like to
-  // avoid introducing C++11 dependency for this single function
-  template< typename T >
-  static std::string ToString( const T & n )
-  {
-    std::ostringstream stm;
-    stm << n;
-    return stm.str();
-  }
-
```
N-Dekker: Nice! Less is more  👍 👍 👍

Core/Main/elastixlib.cxx
kaspermarstal: Yes I wanted to talk to you about this. SmartPointer has a constructor that takes a nullptr, so if you write m_ResultImage(nullptr) the object will get initialized so checks like if(m_ResultImage) will be true whereas we wanted it just be a nullptr so if(m_ResultImage) would be false as before (when initialized to zero). If we used if(m_ResultImage.IsNull()) everywhere in the codebase instead of if(m_ResultImage) that be fine but this is a large task to check and beyond the scope of this PR. Is there a way to initialize m_ResultImage to nullptr in the member initializer list when it has a constructor that takes a nullptr?

Core/Main/elxElastixFilter.hxx
kaspermarstal: No this is just an extra check to be absolutely certain that the object is not null. There was no specific incident that made this necessary.

Core/Main/elxTransformixFilter.h
kaspermarstal: Static is a good idea.

kaspermarstal:
Great, see my question with member initializer list and itk::SmartPointer!

Core/Main/elastixlib.cxx
N-Dekker: Sure, if you do `m_ResultImage(nullptr)` in the mem-initializer-list, it will call `SmartPointer (std::nullptr_t p)` from https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0rc01/Modules/Core/Common/include/itkSmartPointer.h#L69

`if(m_ResultImage)` calls the function `SmartPointer::operator bool()` from https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0rc01/Modules/Core/Common/include/itkSmartPointer.h#L113 which returns false when it is a `nullptr`.

`if(m_ResultImage)` is equivalent to `if( ! m_ResultImage.IsNull() )`.
It is also equivalent to `if ( m_ResultImage != nullptr )`.
I prefer the last syntax, but that's just my personal taste. I would not change it right now.

Does that answer your question?

Core/Main/elxElastixFilter.hxx
N-Dekker: OK!

PR #138, opened 2019-04-17
ENH: GTest for ImageMaskSpatialObject2

N-Dekker:
Added (optional) GoogleTest, to be enabled by advance CMake option `ELASTIX_USE_GTEST`.

Added first (rather trivial) GTest test function, for `ImageMaskSpatialObject2`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -450,6 +450,14 @@ if( BUILD_TESTING )
   add_subdirectory( Testing )
 endif()
 
+mark_as_advanced( ELASTIX_USE_GTEST )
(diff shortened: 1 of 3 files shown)


N-Dekker:
Don't merge! Still unfinished! (WIP = work in progress)

N-Dekker:
It appears unnecessary to do `find_package(GTest REQUIRED)`, as elastix can use the GoogleTest installation from ITK. 

https://cmake.org/cmake/help/v3.14/module/FindGTest.html suggested using `GTest::GTest GTest::Main`.

PR #145, opened 2019-04-29
Make BUILD_SHARED_LIBS do its job

### LINKED ISSUE #81: transformix can't be built as shared library
It uses the same macro as elastix target. This macro is converted to __declspec(dllimport) instead of export during build.
It should use another macro for exporting.

https://github.com/SuperElastix/elastix/blob/582876b698c58a0b54ef9e3cf2f0733aeb53b71c/Core/Install/elxMacro.h#L289
https://github.com/SuperElastix/elastix/blob/582876b698c58a0b54ef9e3cf2f0733aeb53b71c/Core/Main/transformixlib.h#L36

haampie:
Not 100% confident if this is correct, but it's an attempt to fix #81.

### DIFF
--- a/Common/CMakeLists.txt
+++ b/Common/CMakeLists.txt
@@ -193,7 +193,7 @@ source_group( "ImageSamplers" FILES ${ImageSamplersFiles} )
 #---------------------------------------------------------------------
 # Create the elxCommon library.
 
-add_library( elxCommon STATIC
+add_library( elxCommon
   ${CommonFiles}
   ${CostFunctionFiles}
   ${TransformFiles}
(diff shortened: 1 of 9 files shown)


mstaring:
this should be carefully reviewed, as it may break code downstream, for example for the companies using it.

haampie:
It's also failing on Windows: 

```
definition of dllimport function not allowed
```

I'm not very familiar with Windows / dll's, but it seems tricky.

N-Dekker:
Thank you Harmen @haampie for looking into this issue.  Unfortunately we don't really have the resources (both man power and computer resources) to support _both_ shared and static, for all platforms. As I discussed with Marius @mstaring earlier today. So unless there is a use case that appears very important to us, we won't actively support building shared elastix libs.

Technical considerations:

1. A static library can potentially yield better performance than a shared library, because static linking allows the linker to do extra optimizations (function inlining, etc).
2. elastix was never designed for shared libraries, and as it consists of a collection of many small subprojects, it may produce a lot of small shared libraries (if we would support building shared libs), which may be cumbersome.

Hope it's OK to you!

mstaring:
There are many intricacies involved for supporting different types of libs, upstream and downstream.

I'm going to close this PR as there are currently no good test cases, and no resources to investigate this carefully.

PR #166, opened 2019-07-16
COMP: Improve speed of github azure CI build by dropping ITK4

N-Dekker:
It was agreed at today's Elastix/SuperElastix sprint that ITK4 does not need to be supported anymore.

(The aim is now to allow elastix to build with ITK5 `ITK_LEGACY_REMOVE=ON`.)


N-Dekker:
Unfortunately, it still takes more than an hour to build on all virtual machines. Still I will merge, as it will hopefully reduce the waiting time when no VM is available.

PR #167, opened 2019-07-17
Remove ITK migration workarounds

N-Dekker:
Removes workaround from elastix that were needed while migrating from ITK4 to ITK5.

Note that these commits only remove the elastix specific ITK migration workarounds. More work needs to be done to remove the ITK_LEGACY dependency.


Common/CostFunctions/itkParzenWindowHistogramImageToImageMetric.hxx
```diff
@@ -208,8 +208,8 @@ ParzenWindowHistogramImageToImageMetric< TFixedImage, TMovingImage >
   this->m_FixedImageBinSize
     = ( this->m_FixedImageMaxLimit - this->m_FixedImageMinLimit
     + 2.0 * smallNumberFixed ) / fixedHistogramWidth;
-  this->m_FixedImageBinSize = vnl_math_max( this->m_FixedImageBinSize, 1e-10 );
-  this->m_FixedImageBinSize = vnl_math_min( this->m_FixedImageBinSize, 1e+10 );
+  this->m_FixedImageBinSize = vnl_math::max( this->m_FixedImageBinSize, 1e-10 );
+  this->m_FixedImageBinSize = vnl_math::min( this->m_FixedImageBinSize, 1e+10 );
   this->m_FixedImageNormalizedMin
```
mstaring: same here and elsewhere

mstaring:
The commits look fine, but I thought we were completely switching to the stl for min and max operations and the like?

N-Dekker:
@mstaring I just replaced `min` and `max` from VCL/vnl, thanks for the suggestion. I would prefer to now just merge this pull request (if it's OK). Other VCL/vnl functions (abs, cuberoot, rnd) could possibly also be replaced by `std` functions, if you like, but that would take more time. And it might cause regression failures. So I'd like to do so in a separate pull request.

N-Dekker:
Follow-up: https://github.com/SuperElastix/elastix/pull/168

Issue #182, opened 2019-09-10
Replace ANN library?


thewtex:
It would be nice to have something that is not LGPL. Also, there may be faster alternatives now, per:

  http://ann-benchmarks.com/
  https://github.com/erikbern/ann-benchmarks

N-Dekker:
Thanks for the suggestions, Matt! However, I'm sorry to say, we have no plans to fix this issue. Elastix has an option to disable the only component that uses the ANN library, by CMake `USE_KNNGraphAlphaMutualInformationMetric = OFF`.

I hope that that's a sufficient solution to you.

thewtex:
@N-Dekker thanks for the information! Yes, that works for me :+1:

PR #187, opened 2019-09-17
BUG: Do not force ANNlib to build as shared

thewtex:


N-Dekker:
@thewtex  Your effort is very much appreciated, Matt. I just discussed your pull request with Marius. However, I'm sorry to say, we would rather keep the ANN lib "shared", _even_ when elastix is built as a static lib! Especially because of the LGPL license of ANN, as you mentioned in issue #182. If we would link ANN statically to elastix, we would also have to make elastix LGPL, as far as I understand. But we prefer to keep elastix licensed as specified by https://github.com/SuperElastix/elastix/blob/develop/LICENSE

Hope you understand! By the way, I'm curious to hear, if you would like to share with us, what triggered your interest in ANN... 😃

thewtex:
@N-Dekker I updated the branch to keep the `SHARED` specification on the library, but still improve the export specification method so it is standard CMake, which should be more robust.

Yes, got curious about ANN since the LGPL raised an orange flag :netherlands: :-)

N-Dekker:
Sorry Matt, I'm reluctant to do modifications to the ANN, unless really necessary. I very much like the CMake `generate_export_header` command, but is there a real problem with the original `#define DLL_API`? 

 You see, the  `#define DLL_API` code is still from the original version, from https://www.cs.umd.edu/~mount/ANN/ 

    #ifdef DLL_EXPORTS
	   #define DLL_API __declspec(dllexport)
    #else
	  #define DLL_API __declspec(dllimport)
    #endif

From "ann_1.1\include\ANN\ANN.h", from https://www.cs.umd.edu/~mount/ANN/Files/1.0/ann_1.0.tar.gz

If possible, I would prefer not get too much "out of sync" with the original ANN. Hope you understand.

issakomi:
May be i understand it wrong, not 100% sure, but seems that it might be possible to link statically, important is the ability to re-link the program.
"Essentially, if it is a "work that uses the library", then it must be possible for the software to be linked with a newer version of the LGPL-covered program. The most commonly used method for doing so is to use "a suitable shared library mechanism for linking". **Alternatively, a statically linked library is allowed if either source code or linkable object files are provided**.[3]"
[3] [Gnu Lesser General Public License, Section 4](https://www.gnu.org/copyleft/lesser.html#section4)

mstaring:
on the other hand the modifications to the ANN lib are quite straightforward, only replacing this key at places. If it makes library issues more robust, we should consider it...

N-Dekker:
Marius wrote:
> on the other hand the modifications to the ANN lib are quite straightforward, only replacing this key at places.

@thewtex So is the main issue you're aiming to fix that the original ANN `DLL_API` macro does not have an `ANN` prefix? I see that that might possibly lead to ambiguities.

> If it makes library issues more robust, we should consider it...

@mstaring OK, but could we then also consider upgrading to the latest version of ANN? It feels a bit odd to do maintenance on an outdated version (version 1.1, 2005). The latest version (version 1.1.2, 2010) has a few small improvement in coding style and license text.

mstaring:
I believe the use of GenerateExportHeader is a main thing also. I guess that gives a better definition of the DLL_API thing, not only the ANN prefix (but that is a guess)

yes upgrading is a good idea!

N-Dekker:
> yes upgrading is a good idea!

Here it is: Pull request #194 "ENH: Upgrade ANN library from 1.1 (2005) to 1.1.2 (2010)"

N-Dekker:
@thewtex Now that we upgraded ANN (pull request  #194), your export specification commit 127c9ae81166469f46a2a92cb95e4d6077f38a88 cannot be merged automatically anymore. Would you possibly have time to resolve the conflicts? I think it would be rather straightforward, but it probably requires some manual editing.

thewtex:
@N-Dekker yes, I will rebase.

It seems that we need this to avoid the *itkElastixRegistrationMethodTest* test from failing on Windows. It is passing here, with the patch:

  https://dev.azure.com/ITKElastix/ITKElastix/_build/results?buildId=27

And it fails here, without the patch:

  https://dev.azure.com/ITKElastix/ITKElastix/_build/results?buildId=32

N-Dekker:
> yes, I will rebase.

Thanks Matt!

> It seems that we need this to avoid the itkElastixRegistrationMethodTest test from failing on Windows. It is passing here, with the patch

@thewtex Do you have an explanation, why or how this patch would fix a test failure? Until now, I thought it was mostly a style PR, allowing CMake `GenerateExportHeader` to generate export macro's, instead of the original "manually defined" ANN export macro's.

thewtex:
The Elastix / ITK code is #include'ing *ANN.h*, which has the export macros, but we disabled the ANNlib build in ITKElastix -- this is evidently problematic with non-CMake export macros.

N-Dekker:
> The Elastix / ITK code is #include'ing _ANN.h_, which has the export macros, but we disabled the ANNlib build in ITKElastix -- this is evidently problematic with non-CMake export macros.

Thanks for the explanation Matt. But then, when you disable  the ANNlib build, where does the `#include ANN.h` come from? Is the elastix CMake option `USE_KNNGraphAlphaMutualInformationMetric = OFF` not working properly?

thewtex:
> But then, when you disable the ANNlib build, where does the #include ANN.h come from?But then, when you disable the ANNlib build, where does the #include ANN.h come from?

Here are a few examples:

https://github.com/SuperElastix/elastix/blob/285083d01dd923e046a41528db3eadb3e4d435eb/Components/Metrics/KNNGraphAlphaMutualInformation/KNN/itkANNBinaryTreeCreator.h#L24

https://github.com/SuperElastix/elastix/blob/285083d01dd923e046a41528db3eadb3e4d435eb/Components/Metrics/KNNGraphAlphaMutualInformation/KNN/itkBinaryANNTreeBase.h#L22

https://github.com/SuperElastix/elastix/blob/465fad42a9b4f2861ee034f098c9c71074881f2c/Components/Metrics/KNNGraphAlphaMutualInformation/KNN/itkBinaryANNTreeSearchBase.h#L23

> Is the elastix CMake option USE_KNNGraphAlphaMutualInformationMetric = OFF not working properly?

The option ensures the ANNLib is not built, but the manual export specification is not compatible.

thewtex:
@N-Dekker rebased! :construction_worker_man:

thewtex:
Thank you @N-Dekker !

Issue #162, opened 2019-07-12
AffineLogStackTransform: uninitialized local variable 'rdorigin' used


N-Dekker:
`AffineLogStackTransform::ReadCenterOfRotationIndex(ReducedDimensionInputPointType&)` has a suspicious compilation warning from VS2017:

> elastix\components\transforms\affinelogstacktransform\elxaffinelogstacktransform.hxx(574): warning C4700: uninitialized local variable 'rdorigin' used

    dummyImage->SetOrigin( rdorigin );

https://github.com/SuperElastix/elastix/blob/32a1b346956cf3c250ce357249cbc16bf30310f0/Components/Transforms/AffineLogStackTransform/elxAffineLogStackTransform.hxx#L574

How shall we fix it? Is this function still in use?

N-Dekker:
I think the `dummyImage->SetOrigin( rdorigin )` call can just be removed. But I need to figure out how to test it!

mstaring:
fixed by the above commits

PR #190, opened 2019-09-24
BUG: No longer read rotation center as index (elastix transform < 3.402)

N-Dekker:
No longer read "CenterOfRotation" from an old transform parameter file, which is the rotation center of the transformation, specified by its index coordinates.

elastix version >= 3.402 no longer specifies the rotation center by its index coordinates.

As explained by Stefan @stefanklein

"Having a center of rotation in index coordinates in the TransformParameter file is bad practice, since its conversion to a world coordinate depends on the knowledge of the size/spacing/origin/directioncosines of the fixed image that was used for registration. That information is stored in the TransformParameter file as well, but it is easy to forget changing the CenterOfRotationIndex when adjusting the (fixed image) resample domain."

Also discussed with Marius @mstaring

Fixes issue #162 ("AffineLogStackTransform: uninitialized local variable 'rdorigin' used")


N-Dekker:
@stefanklein Shall I also remove the `ReadCenterOfRotationIndex` member functions from `elxAdvancedAffineTransform`, `elxEulerStackTransform`, and `elxEulerTransform`? (It seems that all of them are just for backward compatibility for elastix version < 3.402.)

mstaring:
> 
> 
> @stefanklein Shall I also remove the `ReadCenterOfRotationIndex` member functions from `elxAdvancedAffineTransform`, `elxEulerStackTransform`, and `elxEulerTransform`? (It seems that all of them are just for backward compatibility for elastix version < 3.402.)

I think that is a good idea, since we are moving to a major new version (5.0.0)

stefanklein:
The fix looks good, and I agree that we can remove this from other transforms as well. In elxEulerStackTransform the same function was implemented correctly by the way. But anyway, it's better to remove that functionality everywhere.
Quick recap of reason to deprecate/remove that functionality: Having a center of rotation in index coordinates in the TransformParameter file is bad practice, since its conversion to a world coordinate depends on the knowledge of the size/spacing/origin/directioncosines of the fixed image that was used for registration. That information is stored in the TransformParameter file as well, but it is easy to forget changing the CenterOfRotationIndex when adjusting the (fixed image) resample domain.

N-Dekker:
@stefanklein If you think it's OK, feel free to press the "merge" button 😄

PR #192, opened 2019-09-24
ENH: Bump elastix version from 4.9.0 to 5.0.0

N-Dekker:
Increased __ELASTIX_VERSION and adjusted ITKCompilationOptions.txt

Note: the link to the manual (in CONTRIBUTING.md) should still be updated, once a new manual is available.

PR #195, opened 2019-10-02
BUG: Fix Doxygen for elastix 5.0

N-Dekker:
Three commits to get proper Doxygen HTML output from the Elastix source tree, and a commit to fix a small LaTeX error in the manual.


N-Dekker:
@kaspermarstal Do you have an idea why this `configure_file` call was removed?

N-Dekker:
Note: before generating Doxygen documentation for elastix 5.0, we must still update `__ELASTIX_VERSION` in "Core\Install\elxBaseComponent.h" (pull request #192).

PR #204, opened 2019-11-05
ITK 5 threading updates

thewtex:
@N-Dekker @mstaring @stefanklein there was more to do than expected for the Windows Python packages, but we are green now :-).

I updated the sampler parallelism for ITK 5. This uses the *dynamic* threading methods -- when the TBB backend is enabled, the number of work units adjust dynamically at runtime based on the data and algorithm characteristics and CPU load. I did not have time to continue this throughout the rest of the codebase, but it serves as an example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,8 +47,6 @@ if( NOT ELASTIX_BUILD_EXECUTABLE )
   #set( CMAKE_C_VISIBILITY_PRESET hidden )
   #set( CMAKE_CXX_VISIBILITY_PRESET hidden )
 
-  mark_as_advanced( BUILD_SHARED_LIBS )
-  option( BUILD_SHARED_LIBS "Build shared libraries?" OFF )
   if( BUILD_SHARED_LIBS )
     add_definitions( -D_ELASTIX_USE_SHARED_LIBRARY )
 
--- a/Core/Main/elxParameterObject.h
+++ b/Core/Main/elxParameterObject.h
@@ -64,7 +64,7 @@ public:
   void AddParameterMap( const ParameterMapType & parameterMap );
   const ParameterMapType& GetParameterMap( const unsigned int& index ) const;
   itkGetConstReferenceMacro( ParameterMap, ParameterMapVectorType );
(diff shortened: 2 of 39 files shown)


Core/Main/elxParameterObject.h
```diff
@@ -64,7 +64,7 @@ class ParameterObject : public itk::DataObject
   void AddParameterMap( const ParameterMapType & parameterMap );
   const ParameterMapType& GetParameterMap( const unsigned int& index ) const;
   itkGetConstReferenceMacro( ParameterMap, ParameterMapVectorType );
-  unsigned int GetNumberOfParameterMaps() const { return this->m_ParameterMap.size(); }
+  unsigned int GetNumberOfParameterMaps() const { return static_cast< unsigned int >( this->m_ParameterMap.size() ); }
```
N-Dekker: Thanks for addressing this warning, Matt! I would have preferred changing the return type to `size_t` (instead of a brute-force `static_cast`),  but it appears that that changing the return type would affect other code. So as a quick fix, it seems acceptable to me. 👍

CMakeLists.txt
```diff
@@ -47,8 +47,6 @@ if( NOT ELASTIX_BUILD_EXECUTABLE )
   #set( CMAKE_C_VISIBILITY_PRESET hidden )
   #set( CMAKE_CXX_VISIBILITY_PRESET hidden )
 
-  mark_as_advanced( BUILD_SHARED_LIBS )
-  option( BUILD_SHARED_LIBS "Build shared libraries?" OFF )
```
N-Dekker: Thanks Matt. I'm sorry I haven't been able to reproduce the CMake warning so far, using CMake GUI 3.15.4 on Windows 10. So the default of BUILD_SHARED_LIBS is `OFF`, right? Then I guess indeed, these lines may go.

thewtex: > So the default of BUILD_SHARED_LIBS is OFF, right?

Yes, correct :+1:

N-Dekker:
@thewtex May we close this pull request of yours? It is at least partially implemented by now.

thewtex:
@N-Dekker :+1: yes

PR #213, opened 2019-12-04
Khronos OpenCL support

thewtex:


N-Dekker:
Thanks for your pull request, Matt. Can you please explain why you propose to `#define CL_TARGET_OPENCL_VERSION`? Note that elastix already has CMake flags to control the OpenCL version:

    mark_as_advanced( OPENCL_C_VERSION_1_1 )
    mark_as_advanced( OPENCL_C_VERSION_1_2 )
    mark_as_advanced( OPENCL_C_VERSION_2_0 )

https://github.com/SuperElastix/elastix/blob/5.0.0/CMake/FindOpenCL.cmake#L462

thewtex:
This is required when using the Khronos OpenCL headers:

 https://github.com/KhronosGroup/OpenCL-Headers

Note that the Khronos-supplied repositories are relatively new, which is why elastix did not have support for them in the past.


If it is not defined, then a warning is issued, and a version newer that we would like, 2.2, is used:

https://github.com/KhronosGroup/OpenCL-Headers/blob/de26592167b9fdea503885e40e8755393c56523d/CL/cl_version.h#L33-L34

Limit to 1.2, because this is the version supported on macOS:

 https://support.apple.com/en-us/HT202823

>  Note that elastix already has CMake flags to control the OpenCL version:

This is a preprocessor macro related to the Khronos headers vs a CMake flag.

Common/OpenCL/ITKimprovements/itkOpenCLUtil.h
```diff
@@ -40,6 +40,10 @@
 #include <sstream>
 #include <typeinfo>
 
+#ifndef CL_TARGET_OPENCL_VERSION
+#  define CL_TARGET_OPENCL_VERSION 120
```
N-Dekker: This particular header file (itkOpenCLUtil.h) does not have any OpenCL `#include`, so here `#define CL_TARGET_OPENCL_VERSION` does not seem necessary, does it?

If you still think both itkOpenCL.h and itkOpenCLUtil.h should do `#define CL_TARGET_OPENCL_VERSION`, I hope that the code redundancy can be avoided.

CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ include_directories( ${elxINCLUDE_DIRECTORIES} )
 # to the ITK the BEFORE keyword is not needed anymore.
 if( ELASTIX_USE_OPENCL )
   include_directories( BEFORE ${elxCommon_OpenCL_INCLUDE_DIRECTORIES} )
+  list( APPEND elxINCLUDE_DIRECTORIES ${elxCommon_OpenCL_INCLUDE_DIRECTORIES} )
```
N-Dekker: It looks reasonable, but can you please tell us which problem is solved by adding `elxCommon_OpenCL_INCLUDE_DIRECTORIES` to `elxINCLUDE_DIRECTORIES`? Did you get any compilation error before?

Common/OpenCL/ITKimprovements/itkOpenCLUtil.h
thewtex: @N-Dekker yes, I can remove it form *itkOpenCLUtil.h*. It was placed there because [*opencl.h* is include'd in this file in ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/7e52961776e242290f9725f273b3e3c8f16b8cf8/Modules/Core/GPUCommon/include/itkOpenCLUtil.h#L34-L42). Eventually, we should push the improvements in elastix (which are great) upstream to avoid duplicated effort.

CMakeLists.txt
thewtex: Yes this resolves missing include errors for the associated headers. It is required for [ITKElastix](https://github.com/InsightSoftwareConsortium/ITKElastix) or any other project that build against elastix with `ELASTIX_USE_OPENCL` enabled.

N-Dekker:
@thewtex I have finally been able to reproduce the `CL_TARGET_OPENCL_VERSION` problem, as VS2019 says, when building with intel_sdk_for_opencl_applications_2020.3.494:

>------ Build started: Project: elxOpenCL, Configuration: Debug x64 ------
>itkOpenCLContext.cxx
>cl_version.h: CL_TARGET_OPENCL_VERSION is not defined. Defaulting to 220 (OpenCL 2.2)
>D:\src\elastix\elxOpenCL\Common\OpenCL\ITKimprovements\itkOpenCLContext.cxx(797,13): warning C4996: 'clCreateCommandQueue': was declared deprecated

So I would like to merge your "COMP: Define CL_TARGET_OPENCL_VERSION" commit onto the develop branch.

However, I haven't yet been able to reproduce the problem addressed by the other commit, "COMP: Add elxCommon_OpenCL_INCLUDE_DIRECTORIES to elxINCLUDE_DIRECTORIES". Is that still relevant nowadays?

PR #214, opened 2019-12-04
WIP: BUG: Explicitly update the CPU buffer in OpenCLResampler component

thewtex:


thewtex:
Is this necessary?

N-Dekker:
> Is this necessary?

Thanks for asking, Matt! I asked my LKEB/LUMC colleague Denis @dpshamonin  and he told me that it never appeared necessary, and was working fine on Windows and Linux without this extra update. Why would you think it is necessary?

thewtex:
> Thanks for asking, Matt! I asked my LKEB/LUMC colleague Denis @dpshamonin and he told me that it never appeared necessary, and was working fine on Windows and Linux without this extra update.

Great, thanks for the feedback @N-Dekker @dpshamonin .

I have OpenCL Elastix Python packages building and running:

https://github.com/InsightSoftwareConsortium/ITKElastix/pull/15

but I run into an odd crash, so I explored this (it did not fix the crash).


The registration runs fine, and resampling occurs on the GPU. It completes successfully. However, when attempting to write the resulting resampled image, it crashes when writing to file. I.e.

```
# This runs fine
result = itk.elastix_registration_method(fixed, moving)
# This crashes
itk.imwrite(result, 'result.mha')
```

In Python, I can get the *TransformParameterObject*, and use it to resample with transformix, and there is no crash (and the output is consistent with the CPU version). 

The closest hint when debugging the backtrace on the `result` image is that the segfault occurs here:

https://github.com/InsightSoftwareConsortium/ITK/blob/3dcd26487bd7c9948799929d182ca08a15b0610a/Modules/Core/Common/include/itkImageBase.hxx#L224-L227

So, there is something missing, (maybe a graft or sync??) in the ITK-Elastix / elastix / GPU layers. @dpshamonin @N-Dekker @kaspermarstal @mstaring does this ring any bells?

PR #220, opened 2020-01-28
BUG: ParameterObject argument for itkTypeMacro

thewtex:
The first argument to itkTypeMacro is used to generate the output of the
GetClassName() method -- this should be the specific class name instead
of Self.

### DIFF
--- a/Core/Main/elxElastixFilter.h
+++ b/Core/Main/elxElastixFilter.h
@@ -47,7 +47,7 @@ public:
   itkNewMacro( Self );
 
   /** Run-time type information (and related methods). */
-  itkTypeMacro( Self, itk::ImageSource );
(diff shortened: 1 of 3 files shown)


N-Dekker:
@thewtex Hi Matt, it appears that the same issue occurs on two other places:

  https://github.com/SuperElastix/elastix/blob/5.0.0/Core/Main/elxElastixFilter.h#L50
  https://github.com/SuperElastix/elastix/blob/5.0.0/Core/Main/elxTransformixFilter.h#L50

Would you want to fix those two macro calls as well? Just a question, not a request!  😃

thewtex:
> Would you want to fix those two macro calls as well? 

@N-Dekker yes! Added in 5127d52

PR #223, opened 2020-02-01
STYLE : Updated manual to correctly describe checkout of 5.0.0 branch

shiruken:
The current manual instructs the user to use `git checkout v5.0.0` to get the latest release. The tags for this repository do not include the `v` prefix. The correct commmand should be `git checkout 5.0.0`.


mstaring:
thank you!

PR #224, opened 2020-02-02
Fix incompatibilites with ITK-v5.1rc01

lassoan:
FYI, we had to make these changes to make Elastix work with ITK-v5.1rc01.


lassoan:
replaced by https://github.com/SuperElastix/elastix/pull/240

PR #231, opened 2020-03-04
STYLE: Replace _ELASTIX_BUILD_LIBRARY by BaseComponent::IsElastixLibrary

N-Dekker:
Replaced `_ELASTIX_BUILD_LIBRARY` preprocessor definition by static member function, `BaseComponent::IsElastixLibrary()`.

Anticipates sharing the intermediate binary object files between the Elastix executable and the Elastix library, possibly building them within the same project, or the same (Visual Studio) solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,6 @@ mark_as_advanced( ELASTIX_BUILD_EXECUTABLE )
 option( ELASTIX_BUILD_EXECUTABLE "Generate executable or library?" ON )
 
 if( NOT ELASTIX_BUILD_EXECUTABLE )
-  add_definitions( -D_ELASTIX_BUILD_LIBRARY )
 
   # The following may make smaller and quicker loading libraries,
   # that hides unnecessary symbols. Available from CMake 3.0.0.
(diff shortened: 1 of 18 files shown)


N-Dekker:
@mstaring @stefanklein Please check carefully! 😃  I think we need to remove the dependencies on `_ELASTIX_BUILD_LIBRARY`, in order to be able to build the _exe_ and the _lib_ within one and the same "solution file", sharing the same *.obj" files with Visual C++.

N-Dekker:
@stefanklein As you suggested this morning @ LUMC, I compared the output directory of the example from "elastix\dox", as generated by the official elastix 5.0.0 release to the output directory generated from this pull request. I think they look quite comparable  😄 

I ran:
> elastix.exe -f exampleinput/fixed.mhd -m exampleinput/moving.mhd -out exampleoutput\ -p exampleinput/parameters_Rigid.txt -p exampleinput/parameters_BSpline.txt

result.0.mhd, result.0.raw, result.1.mhd, result.1.raw, and TransformParameters.0.txt are _exactly_ the same in both output directories.

elastix.log has the same number of lines. I only observed different current date and time, and slightly different durations.

mstaring:
Glancing through it, it looks good and I see no reason to worry :-)

As indeed before and after looks the same, I trust the PR and you can merge

PR #236, opened 2020-04-04
ADD: Addition of point to surface metric to elastix.

gokhangg:


N-Dekker:
Thank you Gokhan @gokhangg I just retrieved your work by cloning github.com/gokhangg/elastix.git and did an initial build. Everything builds fine, also on my laptop. 😃  However, it appears that your work it not included with the compilation! Do you still have a CMakeLists.txt that should be added, or a modification to an existing elastix CMake file?

gokhangg:
"However, it appears that your work is not included with the compilation! Do you still have a CMakeLists.txt that should be added, or a modification to an existing elastix CMake file?"

Could not get here, could you please clarify?

N-Dekker:
> Could not get here, could you please clarify?

Your "elxPointToSurfaceDistanceMetric.cxx"should be compiled when elastix is built, right? Does it actually get compiled on your computer? When you deliberately make a syntax error in your cxx file, for example by typing `#error Please get me a compilation error!`, do you get a compilation error indeed?

N-Dekker:
I think you should add a "CMakeLists.txt" to the directory of your metric, having something like this:

    ADD_ELXCOMPONENT( PointToSurfaceDistanceMetric ON
      elxPointToSurfaceDistanceMetric.h
      elxPointToSurfaceDistanceMetric.hxx
      elxPointToSurfaceDistanceMetric.cxx
      itkPointToSurfaceDistanceMetric.h
      itkPointToSurfaceDistanceMetric.hxx )

Also I think the directory of your metric should be renamed to "PointToSurfaceDistanceMetric" (by adding the word "Metric"), otherwise I still get build errors.

gokhangg:
Hi Niels, I have checked this issue and it seems that the .gitignore blocks uploading cmake related files.
Line 36 ignores "cmake*". So actually that file was inside but git ignored it while pushing. So I think you should delete that line in the gitignore file or I can send that CMakeLists.txt file to and then you can upload to the repo.

N-Dekker:
@gokhangg 
> I have checked this issue and it seems that the .gitignore blocks uploading cmake related files.
Line 36 ignores "cmake*".

Wow, well spotted! I'll remove "cmake*" from the ignore-list, thanks!

N-Dekker:
@gokhangg Could you please still try to add your CMakeLists to your branch, using the git option "--force"? Something like this, if I understand correctly:

    git add --force CMakeLists.txt

As I read at https://stackoverflow.com/questions/24844782/how-to-override-a-rule-set-in-gitignore

gokhangg:
Sure I can indeed bypass the ignore but just wanted to inform you. After the pull for the cmake fix, could you inform me then I will update my request?

N-Dekker:
OK, I removed "cmake*" from .gitignore (pull request #239). Can you please add your CMakeLists.txt to your pull request?

gokhangg:
Sure

gokhangg:
You can merge now ;)

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,172 @@
...
+
+#include "elxIncludes.h" // include first to avoid MSVS warning
+#include "itkPointToSurfaceDistanceMetric.h"
+
+namespace elastix
+{
+
+/**
+ * \class TransformRigidityPenalty
+ * \brief A penalty term based on non-rigidity.
```
N-Dekker, line 25: Is this indeed a penalty term based on non-rigidity?

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.cxx
```diff
@@ -0,0 +1,17 @@
...
+
+  Copyright (c) University Medical Center Utrecht. All rights reserved.
+  See src/CopyrightElastix.txt or http://elastix.isi.uu.nl/legal.php for
+  details.
+
+     This software is distributed WITHOUT ANY WARRANTY; without even
+     the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+     PURPOSE. See the above copyright notices for more information.
+
+======================================================================*/
```
N-Dekker, line 13: I think the license text is different now, please check the current version of other Elastix source files! Otherwise please ask Marius @mstaring

gokhangg:
Can we copy from other files? I think should be the same ??

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,172 @@
...
+  virtual int BeforeAllBase();
+
+  /**
+   * Do some things before registration:
+   * \li Load and set the pointsets.
+   */
+  virtual void BeforeRegistration();
+
+/** Function to read the corresponding points. */
+/** Function to read the corresponding points. */
```
N-Dekker, line 147: Looks like these lines are accidentally duplicated!

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,172 @@
...
+
+  /** Overwrite to silence warning. */
+  virtual void SelectNewSamples( void ){}
+
+protected:
+
+  /** The constructor. */
+  PointToSurfaceDistanceMetric(){}
+  /** The destructor. */
+  virtual ~PointToSurfaceDistanceMetric() {}
```
N-Dekker, line 160: You may consider C++11 `= default` instead of an empty function body `{}`, for both the default-constructor and the destructor.

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,126 @@
...
+protected:
+
+    PointToSurfaceDistanceMetric();
+    virtual ~PointToSurfaceDistanceMetric() {}
+    bool m_AvPointWeigh;
+
+private:
+
+    PointToSurfaceDistanceMetric( const Self & );
+    std::string m_SegFileIn,m_DTFileIn,m_DTFileOut;
```
N-Dekker, line 114: Please avoid abbreviations in C++ identifiers, if you can. "Seg", "DT", "AvPointWeigh" may be clear enough to you, but I would have to guess, or figure out the meaning.

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
gokhangg: These lines have been removed.

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
gokhangg: Fixed

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
gokhangg: I have changed the names. So more manifest names have been coined.

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
gokhangg: I missed this

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
N-Dekker: Thank you @gokhangg

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.cxx
N-Dekker: > Can we copy from other files? I think should be the same ??

@gokhangg Yes, please, ensure that the copyright text is the same as in https://github.com/SuperElastix/elastix/blob/5.0.0/Core/Main/elastix.cxx

gokhangg: Done

N-Dekker: Thanks!

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,130 @@
...
+    void SetDTImageOut(  const std::string  str );
+
+    /** Get Distance Transform Image*/
+    ConstDTimage GetDTImage() const;
+
+protected:
+
+    PointToSurfaceDistanceMetric();
+    virtual ~PointToSurfaceDistanceMetric() = default;
+    bool m_AvPointWeigh;
```
N-Dekker, line 113: Please initialize m_AvPointWeigh, for example as follows:

    bool m_AvPointWeigh { true };

Doing so should solve a Visual Studio 2019 IntelliSense warning:
> warning C26495: Variable '::m_AvPointWeigh' is uninitialized. Always initialize a member variable (type.6).

gokhangg: It is to be initialized in the derived class. But I updated the code for the sake of the compiler, in order for it does not issue warning.

gokhangg: Done

N-Dekker: Thanks @gokhangg 
> It is to be initialized in the derived class.

Thanks, I see now it would already be initialized in `elastix::PointToSurfaceDistanceMetric`. Still I think it's preferable to avoid compilation warnings, especially when they can be fixed easily. But I think we agree already  `:-)`

It would be nice, in my opinion, to have a GoogleTest that just tests the base class, `itk::PointToSurfaceDistanceMetric`. Then it should not depend on initialization being done by the derived class anyway.

See you on 15:00 😃

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,172 @@
...
+  typename MetricBase< TElastix >::MovingPointSetType >,
+  public MetricBase< TElastix >
+{
+public:
+
+  using Self = PointToSurfaceDistanceMetric;
+  /** The private constructor. */
+  PointToSurfaceDistanceMetric(const Self &) = delete;
+  /** The private copy constructor. */
+  void operator=(const Self &) = delete;
```
N-Dekker, line 69: Please note: `PointToSurfaceDistanceMetric(const Self &)` is not a _private_ constructor, and `operator=(const Self &)` is an assignment operator, _not_ a copy constructor. Personally I would do `ITK_DISALLOW_COPY_AND_ASSIGN(PointToSurfaceDistanceMetric)` instead.

gokhangg: The comments just remaining from previous code that I forgot to modify. Can we use ITK_DISALLOW_COPY_AND_ASSIGN in elastix code?

N-Dekker: @gohkangg I realize I can't blame you for the incorrect comment, because it is there already in previous elastix code! `ITK_DISALLOW_COPY_AND_ASSIGN` is part of the public interface of ITK, and elastix uses ITK already, so yes, we can use it  😃

gokhangg: Done ;)

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,130 @@
...
+    using WriterType = itk::ImageFileWriter< ImageType >;
+    using InterpolatorType = itk::BSplineInterpolateImageFunction< ImageType, float, float >;
+    using ConstDTimage = typename ImageType::ConstPointer;
+	 /** Method for creation through the object factory. */
+    itkNewMacro( Self );//OK
+
+    /** Run-time type information (and related methods). */
+    itkTypeMacro( PointToSurfaceDistanceMetric, SingleValuedPointSetToPointSetMetric );//OK
+    /**  Initialize. */
+    virtual void Initialize() noexcept(false);
```
N-Dekker, line 86: I see you marked some member functions `noexcept(false)`, indicating that they _might_ throw an exception. Why would you do so? When I use `noexcept`, it's usually just to indicate that a function does _not_ throw.

gokhangg: Fixed.

N-Dekker: @gokhangg Thanks for your patience. I still see `noexcept(false)` five times in your proposal. Would you want to remove these as well? Or do they have a special reason?

N-Dekker:
> No description provided.

Can you please add a description to your commit? Actually that's a request from Marius (@mstaring). 😃  The formatting convention for a commit message is as follows:

* The first line of the commit message starts with "ENH: " for enhancement, and is at most 72 characters (ITK convention). It is the "subject line" of the commit message.
* The second line is entirely empty.
* The next lines give a more detailed description.

mstaring:
@stefanklein can you comment on this PR? Is it ready for merging?

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
gokhangg: All removed

gokhangg: All removed

gokhangg:
> > No description provided.
> 
> Can you please add a description to your commit? Actually that's a request from Marius (@mstaring). 😃 The formatting convention for a commit message is as follows:
> 
> * The first line of the commit message starts with "ENH: " for enhancement, and is at most 72 characters (ITK convention). It is the "subject line" of the commit message.
> * The second line is entirely empty.
> * The next lines give a more detailed description.

gokhangg:
> can you comment on this PR? Is it ready for merging?

Hi Marius I gave a comment, is it possible to check if it is in the correct condition?

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.h
N-Dekker: Thanks @gokhangg Nowadays I think it's better to only use `noexcept` when the benefit is really clear (which is clearly the case for move-constructors and move-assignment operators). But other people think otherwise... For your information, I did a little lightning talk about `noexcept` at the conference C++ on Sea, two years ago https://www.youtube.com/watch?v=dVRLp-Rwg0k regarding my [noexcept benchmark](https://github.com/N-Dekker/noexcept_benchmark)

N-Dekker:
@gokhangg Can you possibly squash all (four) commits of your pull request together, to make it one single commit?

     git rebase -i HEAD~4

An then replace "pick" on the second and subsequent commits with "squash": https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

And then `git push --force-with-lease`, of course  😃

gokhangg:
@N-Dekker  Hi Niels there are some conflicts popped up after squashing the middle-commits. So seems that there is a discrepancy between my fork and the  superelastix::develop branch. So shall we resolve the conflicts manually?

N-Dekker:
Oh no... git happening 😢 Should any of those conflicting files be adjusted by your pull request?

N-Dekker:
@gokhangg  It would be wonderful if you could fix the git conflicts. But you probably need a lot of "git magic" to get it right. It might be easier to just create a new branch from the latest SuperElastix/elastix "develop", and create a new pull request. (Don't feel bad... git happens 😸) I leave it up to you.

gokhangg:
@N-Dekker ACtually those changed files are coming from other places not in this PR. I will try to solve it.

gokhangg:
@N-Dekker I have found out that it is related to neither the changes I did nor the discrepancy in the base. There was a nonlinearity in the commit line so after squashing all other commits except the one breaking the linearity it was solved. Indeed I changed the message of the commit :)

Components/Metrics/PointToSurfaceDistance/itkPointToSurfaceDistanceMetric.hxx
```diff
@@ -0,0 +1,341 @@
...
+
+{
+}
+
+template< class TFixedPointSet, class TMovingPointSet >
+void PointToSurfaceDistanceMetric< TFixedPointSet, TMovingPointSet >
+::Initialize()
+{
+  /** Initialize transform, interpolator, etc.
+  Superclass::Initialize();
```
N-Dekker, line 45: Is it the intention that this `Superclass::Initialize()` call is commented out?

Components/Metrics/PointToSurfaceDistance/elxPointToSurfaceDistanceMetric.h
```diff
@@ -0,0 +1,169 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef __elxPointToSurfaceDistanceMetric_H__
+#define __elxPointToSurfaceDistanceMetric_H__
```
N-Dekker, line 19: (Minor nitpick) In elastix, as well as in ITK, we no longer have those `__` underscores at the begin and the end of include guard macro names. See https://github.com/SuperElastix/elastix/pull/339 Of course, we can remove them from your header files later, no problem.

N-Dekker:
@gokhangg Could you possibly also include a little test (unit test) with your pull request?

PR #239, opened 2020-04-08
COMP: Remove cmake* from .gitignore

N-Dekker:
Having cmake* in there stopped Gokhan Gunay from adding his CMakeLists file to pull request #236 (Addition of PointToSurface metric).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,6 @@
 
 # Generated files and
 help/
-cmake*
 Makefile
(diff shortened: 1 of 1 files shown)


N-Dekker:
@kaspermarstal Is it still useful for you to have "cmake*" in the gitignore list? Apparently it stopped @gokhangg from adding his CMakeLists.txt !

PR #240, opened 2020-04-13
Fix build error with ITK v5.1rc03

lassoan:
Please merge this to fix build errors with latest ITK release candidate.

### DIFF
--- a/Common/Transforms/itkAdvancedRigid2DTransform.hxx
+++ b/Common/Transforms/itkAdvancedRigid2DTransform.hxx
@@ -35,7 +35,7 @@
 #define __itkAdvancedRigid2DTransform_hxx
 
 #include "itkAdvancedRigid2DTransform.h"
-#include "vnl/algo/vnl_svd.h"
+#include "vnl/algo/vnl_svd_fixed.h"
 
 namespace itk
 {
@@ -121,11 +121,9 @@ AdvancedRigid2DTransform< TScalarType >
 {
   // Extract the orthogonal part of the matrix
   //
-  vnl_matrix< TScalarType > p( 2, 2 );
-  p = this->GetMatrix().GetVnlMatrix();
-  vnl_svd< TScalarType >    svd( p );
-  vnl_matrix< TScalarType > r( 2, 2 );
-  r = svd.U() * svd.V().transpose();
+  const vnl_matrix_fixed< TScalarType, 2, 2 > p = this->GetMatrix().GetVnlMatrix();
+  const vnl_svd_fixed< TScalarType, 2, 2 >    svd( p );
+  const vnl_matrix_fixed< TScalarType, 2, 2 > r = svd.U() * svd.V().transpose();
 
   m_Angle = std::acos( r[ 0 ][ 0 ] );
 
--- a/Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h
+++ b/Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h
@@ -73,7 +73,6 @@ public:
   typedef typename Superclass1::InterpolatorPointerType InterpolatePointerType;
   typedef typename Superclass1::SizeType                SizeType;
   typedef typename Superclass1::IndexType               IndexType;
-  typedef typename Superclass1::PointType               PointType;
   typedef typename Superclass1::PixelType               PixelType;
   typedef typename Superclass1::OutputImageRegionType   OutputImageRegionType;
   typedef typename Superclass1::SpacingType             SpacingType;


Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h
```diff
@@ -73,7 +73,8 @@ class MyStandardResampler :
   typedef typename Superclass1::InterpolatorPointerType InterpolatePointerType;
   typedef typename Superclass1::SizeType                SizeType;
   typedef typename Superclass1::IndexType               IndexType;
-  typedef typename Superclass1::PointType               PointType;
+  typedef typename Superclass1::InputPointType          InputPointType;
+  typedef typename Superclass1::OutputPointType         OutputPointType;
```
N-Dekker, line 77: Thank you Andras. I encountered the same problem, when trying to compile elastix with the latest revision from the master branch of ITK:
> error C2039: 'PointType': is not a member of 'itk::ResampleImageFilter'

Personally I would prefer to simply remove the `MyStandardResampler::PointType` typedef, because that could allow supporting _both_ ITK 5.0.1 and 5.1. (At least for now.) Do you think it's really necessary to add those `InputPointType` and `OutputPointType` typedefs here?

Common/Transforms/itkAdvancedRigid2DTransform.hxx
```diff
@@ -121,10 +121,10 @@ AdvancedRigid2DTransform< TScalarType >
 {
   // Extract the orthogonal part of the matrix
   //
-  vnl_matrix< TScalarType > p( 2, 2 );
+  vnl_matrix_fixed< TScalarType, 2, 2 > p( 2, 2 );
```
N-Dekker, line 124: While I do think it could be nice to use `vnl_matrix_fixed` here (for style or performance reasons), I wonder, did the original code really cause a build error?

N-Dekker: Oh, sorry Andras, just now I see your MacOs error message! So it has to do with [the next line](https://github.com/SuperElastix/elastix/blob/5.0.0/Common/Transforms/itkAdvancedRigid2DTransform.hxx#L125), which does:

    p = this->GetMatrix().GetVnlMatrix();

I see now that `itk::Matrix::GetVnlMatrix()` returns a `vnl_matrix_fixed`, and it did so for many years already!

Personally I would strongly prefer combining the two lines of code (construction + assignment), as  in:

    vnl_matrix_fixed< TScalarType, 2, 2 > p = this->GetMatrix().GetVnlMatrix();

Would that be OK to you as well? (If so, `p` might as well be declared `const`!)

Common/Transforms/itkAdvancedRigid2DTransform.hxx
```diff
@@ -121,10 +121,10 @@ AdvancedRigid2DTransform< TScalarType >
...
   p = this->GetMatrix().GetVnlMatrix();
   vnl_svd< TScalarType >    svd( p );
-  vnl_matrix< TScalarType > r( 2, 2 );
+  vnl_matrix_fixed< TScalarType, 2, 2 > r( 2, 2 );
   r = svd.U() * svd.V().transpose();
```
N-Dekker, line 128: Again, thank you for the fix, can you please consider combining these two lines of code, as in:

    const vnl_matrix_fixed< TScalarType, 2, 2 > r = svd.U() * svd.V().transpose();

lassoan:
As a maintainer in this repository, you can push changes to my branch. Feel free to make the modifications as you suggested.

Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h
lassoan: I don't like having tons of typedefs but this is the ITK style. You can make it work with various ITK versions by using ifdefs.

Common/Transforms/itkAdvancedRigid2DTransform.hxx
lassoan: It is OK for me as long as it compiles on all three platforms.

N-Dekker:
> As a maintainer in this repository, you can push changes to my branch. Feel free to make the modifications as you suggested.

Thanks Andras, I'll give it a try!

Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h
N-Dekker: @lassoan I'm glad you don't like having tons of typedefs either 😃 Some typedefs could be useful, of course, so my main question would be: could anything be broken by removing those specific typedefs from `elastix::MyStandardResampler`. Especially because I feel that supporting multiple ITK versions by means of ifdefs is also a pain.

Now it looks like even without explicitly specifying those two typedef's with this PR, it's still possible to use `elastix::MyStandardResampler<T>::InputPointType` and `elastix::MyStandardResampler<T>::OutputPointType`, when using ITK 5.1rc03!!! Because such typedefs are publicly made accessibly already by its first base class, `itk::ResampleImageFilter`! So I don't think removing those typedefs from the PR does any harm.

Common/Transforms/itkAdvancedRigid2DTransform.hxx
N-Dekker: @lassoan I'm sorry, here at github.com/SuperElastix/elastix we only compile elastix + ITK 5.0.1, on the CI (for three platforms). I'll just merge your PR if we can get it compiled here.

N-Dekker:
@lassoan Are you OK with [the adjustments that I did to your pull request](https://github.com/SuperElastix/elastix/compare/c6d26ff852ffb64c2588f7108d23bc08c7975032..52e1dc3928046f9fbb85d4b2ecd0d5175fa9695d)? 
Please check https://github.com/SuperElastix/elastix/pull/240/commits

lassoan:
OK for me. Typedefs are just for following ITK style and make implementation derived classes simpler and safer (if the parent class changes content of a typedef then child class does not have to be changed).

PR #241, opened 2020-04-22
COMP: Remove unreachable code from Transforms

N-Dekker:
Fixes six Visual C++ 2019 Warning level 4 warnings, saying:

> warning C4702: unreachable code

These warnings occurred in Common/Transforms and Components/Transforms.

### DIFF
--- a/Components/Transforms/SplineKernelTransform/itkKernelTransform2.hxx
+++ b/Components/Transforms/SplineKernelTransform/itkKernelTransform2.hxx
@@ -331,7 +331,6 @@ KernelTransform2< TScalarType, NDimensions >
   {
     itkExceptionMacro( << "ERROR: invalid matrix inversion method ("
                        << this->m_MatrixInversionMethod << ")" );
-    this->m_LInverseComputed = false;
   }
 
 } // end ComputeLInverse()
--- a/Common/Transforms/itkAdvancedBSplineDeformableTransformBase.h
+++ b/Common/Transforms/itkAdvancedBSplineDeformableTransformBase.h
@@ -226,7 +226,6 @@ public:
   OutputVectorType TransformVector( const InputVectorType & ) const override
   {
     itkExceptionMacro( << "Method not applicable for deformable transform." );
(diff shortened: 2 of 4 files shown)


Components/Transforms/SplineKernelTransform/itkKernelTransform2.hxx
```diff
@@ -331,7 +331,6 @@ KernelTransform2< TScalarType, NDimensions >
   {
     itkExceptionMacro( << "ERROR: invalid matrix inversion method ("
                        << this->m_MatrixInversionMethod << ")" );
-    this->m_LInverseComputed = false;
```
N-Dekker, line 334: Note that this assignment never happened. Might be a bug!

N-Dekker: So shall I fix this bug, by placing the assignment (`m_LInverseComputed = false`) before the exception? Or is this line of code irrelevant anyway?

N-Dekker: For the record: the bug (doing an assignment after the exception macro call) was introduced more than nine years ago:  commit 401df2108fe8a26b29e3f1f43dab15a22f3b2eef 12 January 2011:
> MS:
>
> -PERF: many performance enhancements for the thin plate spline.
>  - QR decomposition for matrix inversion is much (x10) faster then SVD
>   - Huge speedup for Jacobian computation, using the facts that LMatrixInverse is sparse,
>    has a specific shape, and is symmetric.
>  - Cached decomposition of LMatrix, beneficial when SetParameters() is called (every> registration iteration).
> -ENH: improved TPS test. Test more, and test for failure.

N-Dekker: @mstaring I guess this bug is not a serious one, as it has been in there for so long already, so I think I'll just merge the pull request of mine this afternoon, without fixing the bug.

PR #248, opened 2020-05-01
ENH: Replace InitializeElastixLibrary() by InitializeElastixExecutable()

N-Dekker:
Kasper Marstal reported yesterday via direct messaging that the introduction of `IsElastixLibrary()`, pull request #231, commit 64ced5942604b055e3a11ee235af186811bf11dc (March 2, 2020, "Replace _ELASTIX_BUILD_LIBRARY by BaseComponent::IsElastixLibrary"), did break SimpleElastix.  SimpleElastix depends on the library interface, while it does not construct an `elastix::ELASTIX` library object.

The most elegant and reliable fix appears to be flipping the default value returned by `IsElastixLibrary()`, from `false` to `true`, and then ensuring that the value is set to `false` at the start of each run of the executable.

### DIFF
--- a/Core/Install/elxBaseComponent.cxx
+++ b/Core/Install/elxBaseComponent.cxx
@@ -22,9 +22,9 @@
 
 namespace
 {
-bool IsElastixLibrary(const bool initialValue = false)
+bool IsElastixLibrary(const bool initialValue = true)
 {
-  // By default, assume that this is not the elastix library.
+  // By default, assume that this is the elastix library (not the elastix executable).
 
   // Note that the initialization of this static variable is thread-safe,
   // as supported by C++11 "magic statics".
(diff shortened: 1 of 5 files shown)


N-Dekker:
@kaspermarstal If you think the pull request is fine (and if it compiles and passes the tests), you may just click "Rebase and merge", to get rid of the blocker 😃

kaspermarstal:
> I think you're right, thanks, I'll amend the commit, but do you think it would matter? Would it really change the behavior of the transformix executable? It depends on where, on which places `IsElastixLibrary()` is actually checked: [64ced59](https://github.com/SuperElastix/elastix/commit/64ced5942604b055e3a11ee235af186811bf11dc)

The change broke the tests with the transformix executable so it was definitely needed. Thanks!

N-Dekker:
> @kaspermarstal If you think the pull request is fine (and if it compiles and passes the tests), you may just click "Rebase and merge"

Thanks for pressing the button, Kasper! 👍 

Minor comment: it appears that your merge has resulted in _two_ commits, instead of _one_, at https://github.com/SuperElastix/elastix/commits/develop  I think that's because you clicked _"Merge pull request"_, right? When you would have clicked _"Rebase and merge"_, I think it would have resulted in a _single_ commit the develop branch, combining the code change and the merge into one. Which is my personal preference. No need to change it for this time, of course! 😃

PR #230, opened 2020-03-02
Executable linked to elastix.lib, offering subset of elastix.exe functionality

N-Dekker:

PR #266, opened 2020-08-25
COMP: Declare ConvertSecondsToDHMS and GetCurrentDateAndTime inline

N-Dekker:
Allows these functions to be called from an executable that links against the elastix library.

### DIFF
--- a/Core/Main/elastix.h
+++ b/Core/Main/elastix.h
@@ -40,6 +40,7 @@ void PrintHelp( void );
 /** ConvertSecondsToDHMS
  *
  */
+inline
 std::string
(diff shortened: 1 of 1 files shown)


N-Dekker:
This pull request follows C++ Core Guideline [SF.2: A .h file may not contain object definitions or non-inline function definitions](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf2-a-h-file-may-not-contain-object-definitions-or-non-inline-function-definitions)

N-Dekker:
This pull request is necessary for "Executable linked to elastix.lib, offering subset of elastix.exe functionality", pull request https://github.com/SuperElastix/elastix/pull/230 but also without PR 230, it's a good thing to add those two `inline` keywords.

PR #281, opened 2020-09-11
COMP: elxParameterObject size_t to unsigned int

N-Dekker:
To address:

  d:\a\1\build\_deps\elastix_fetch-src\core\main\elxParameterObject.h(67): warning C4267: 'return': conversion from 'size_t' to 'unsigned int', possible loss of data


N-Dekker:
@thewtex Better late than never... your commit from January 12, 2020 😃 

Honestly I'm not terribly fond of using `static_cast`, as it may just silence a compiler warning without solving the potential type conversion error. Although it's certainly preferable to C-style cast and function-style cast, `unsigned(this->m_ParameterMap.size())`, of course. And I agree that the warning should be silenced anyway! So thanks Matt! I'll merge when the CI allows me to!

PR #282, opened 2020-09-11
COMP: Fix CMake 3.x warnings policy CMP0033 and CMP0007 (issue #21)

N-Dekker:
Fixed the warnings reported by Jean-Christophe Fillion-Robin, issue #21 (August 8, 2017!):

> The OLD behavior for policy CMP0033 will be removed from a future version of CMake.
> The OLD behavior for policy CMP0007 will be removed from a future version of CMake.

Removed `export_library_dependencies` call, which is no longer supported now. ("The export_library_dependencies command should not be called; see CMP0033")

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,14 +11,6 @@ endif()
 
 #---------------------------------------------------------------------
 cmake_policy( SET CMP0012 NEW )
-if( POLICY CMP0033 )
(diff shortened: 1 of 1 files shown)


N-Dekker:
Supersedes pull request  #46 "Fix CMake warnings + drop old style export of library" (actually this is basically a subset of pull request  #46).

PR #287, opened 2020-09-14
STYLE: Remove internal code duplication AdvancedImageMomentsCalculator

N-Dekker:


N-Dekker:
@mstaring This pull request would remove the duplication of "post processing code" between `ComputeSingleThreaded()` and `AfterThreadedCompute()`. But then I still wonder if `Compute()` is OK. For the single threaded case, it seems to do the post processing twice, as it calls both `ComputeSingleThreaded()` and `AfterThreadedCompute()`! What do you think?

https://github.com/SuperElastix/elastix/blob/a7e4fb99954f18eb34a65d60acc8ea1f1a5eeb12/Common/Transforms/itkAdvancedImageMomentsCalculator.hxx#L268-L289

mstaring:
Hi Niels,


The PR with the removal of code duplications seems good.


Please compare with the original itk::ImageMomentsCalculator for your second question. If my memory serves me well I just added mask support with lines like these:


if (m_SpatialObjectMask.IsNull() || m_SpatialObjectMask->IsInsideInWorldSpace(physicalPosition))


Otherwise it should just be a copy of the itk version.


Best, Marius


Leiden University Medical Center | Division of Image Processing (LKEB), Department of Radiology | PO Box 9600 | 2300 RC Leiden | The Netherlands | Postzone C2-s, room C3-53 | +31 (0)71 52 62137


________________________________

PR #291, opened 2020-09-21
ENH: Enable almost all components by default

N-Dekker:
Removed default "OFF" for `ADD_ELXCOMPONENT` calls of 27 components.

Note that Optimizers/PreconditionedGradientDescent is still OFF by default because of its SuiteSparse library dependency.

Transforms/BSplineDeformableTransformWithDiffusion also remained OFF by default.

### DIFF
--- a/Components/FixedImagePyramids/FixedShrinkingPyramid/CMakeLists.txt
+++ b/Components/FixedImagePyramids/FixedShrinkingPyramid/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( FixedShrinkingPyramid OFF
+ADD_ELXCOMPONENT( FixedShrinkingPyramid # Was OFF by default before elastix 5.0.1
  elxFixedShrinkingPyramid.h
  elxFixedShrinkingPyramid.hxx
  elxFixedShrinkingPyramid.cxx )
--- a/Components/Interpolators/BSplineInterpolatorFloat/CMakeLists.txt
+++ b/Components/Interpolators/BSplineInterpolatorFloat/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( BSplineInterpolatorFloat OFF
+ADD_ELXCOMPONENT( BSplineInterpolatorFloat # Was OFF by default before elastix 5.0.1
  elxBSplineInterpolatorFloat.h
  elxBSplineInterpolatorFloat.hxx
  elxBSplineInterpolatorFloat.cxx )
--- a/Components/Interpolators/NearestNeighborInterpolator/CMakeLists.txt
+++ b/Components/Interpolators/NearestNeighborInterpolator/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( NearestNeighborInterpolator OFF
+ADD_ELXCOMPONENT( NearestNeighborInterpolator # Was OFF by default before elastix 5.0.1
  elxNearestNeighborInterpolator.h
  elxNearestNeighborInterpolator.hxx
  elxNearestNeighborInterpolator.cxx )
--- a/Components/Interpolators/RayCastInterpolator/CMakeLists.txt
+++ b/Components/Interpolators/RayCastInterpolator/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( RayCastInterpolator OFF
+ADD_ELXCOMPONENT( RayCastInterpolator # Was OFF by default before elastix 5.0.1
  elxRayCastInterpolator.h
  elxRayCastInterpolator.hxx
  elxRayCastInterpolator.cxx )
--- a/Components/Metrics/AdvancedKappaStatistic/CMakeLists.txt
+++ b/Components/Metrics/AdvancedKappaStatistic/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( AdvancedKappaStatisticMetric OFF
+ADD_ELXCOMPONENT( AdvancedKappaStatisticMetric # Was OFF by default before elastix 5.0.1
  elxAdvancedKappaStatisticMetric.h
  elxAdvancedKappaStatisticMetric.hxx
  elxAdvancedKappaStatisticMetric.cxx
--- a/Components/Metrics/DisplacementMagnitudePenalty/CMakeLists.txt
+++ b/Components/Metrics/DisplacementMagnitudePenalty/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-ADD_ELXCOMPONENT( DisplacementMagnitudePenalty OFF
+ADD_ELXCOMPONENT( DisplacementMagnitudePenalty # Was OFF by default before elastix 5.0.1
  elxDisplacementMagnitudePenalty.h
  elxDisplacementMagnitudePenalty.hxx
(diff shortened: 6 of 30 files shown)


N-Dekker:
Enables thirty components by default, that were previously OFF by default:

    FixedImagePyramids/FixedShrinkingPyramid    
    Interpolators/BSplineInterpolatorFloat    
    Interpolators/NearestNeighborInterpolator    
    Interpolators/RayCastInterpolator    
    Metrics/AdvancedKappaStatistic    
    Metrics/DisplacementMagnitudePenalty    
    Metrics/GradientDifference    
    Metrics/MissingStructurePenalty    
    Metrics/MutualInformationHistogram    
    Metrics/NormalizedGradientCorrelation    
    Metrics/PatternIntensity    
    Metrics/PolydataDummyPenalty    
    Metrics/StatisticalShapePenalty    
    Metrics/ViolaWellsMutualInformation    
    MovingImagePyramids/MovingShrinkingPyramid    
    Optimizers/CMAEvolutionStrategy    
    Optimizers/ConjugateGradientFRPR    
    Optimizers/RSGDEachParameterApart    
    Optimizers/Simplex    
    Optimizers/SimultaneousPerturbation    
    ResampleInterpolators/BSplineResampleInterpolatorFloat    
    ResampleInterpolators/LinearResampleInterpolator    
    ResampleInterpolators/NearestNeighborResampleInterpolator    
    ResampleInterpolators/RayCastResampleInterpolator    
    Transforms/AffineDTITransform    
    Transforms/AffineLogTransform    
    Transforms/BSplineDeformableTransformWithDiffusion    
    Transforms/MultiBSplineTransformWithNormal    
    Transforms/SimilarityTransform    
    Transforms/WeightedCombinationTransform

mstaring:
I would not want to do it like this. The defaults are good being OFF. For a secondary tool like ITKElastix just selecting use_all_components should work.

N-Dekker:
@mstaring Thanks for your reply. Can you explain why so many (30) components are off by default?

stefanklein:
Hi @mstaring, we discussed at the sprint that it would be better to enable more components by default. Of course we can discuss which components. But why would we want more components by default in ITK-Elastix than in normal elastix? If we want components to be available for itk-elastix users, we also would like them to be available for normal elastix users, right?
Best, Stefan

mstaring:
The purpose of the defaults is to have an elastix that by default builds fast. So we selected the important components for this.

If we need more components, then we simply have the use_all_components flag. This principle is the same for elastix and ITKElastix.

So I still disagree with this patch. Perhaps we should discuss in person.

N-Dekker:
> For a secondary tool like ITKElastix just selecting use_all_components should work.

@mstaring I tried to just add `set(USE_ALL_COMPONENTS ON)` to [ITKElastix/CMakeLists.txt](https://github.com/InsightSoftwareConsortium/ITKElastix/blob/master/CMakeLists.txt) in https://github.com/InsightSoftwareConsortium/ITKElastix/pull/65 but then I saw that `KNNGraphAlphaMutualInformation` got built, even while it does `set(USE_KNNGraphAlphaMutualInformationMetric OFF)`. 

https://github.com/InsightSoftwareConsortium/ITKElastix/pull/65/commits/1ce2dd8ae9ce149d3837693e32925ed0dfab8f99

Do you have a clue?

mstaring:
yes, that is the logic of the USE_ALL_COMPONENTS flag as defined here:

https://github.com/SuperElastix/elastix/blob/develop/Components/CMakeLists.txt

N-Dekker:
> yes, that is the logic of the USE_ALL_COMPONENTS flag as defined here:
> https://github.com/SuperElastix/elastix/blob/develop/Components/CMakeLists.txt

@mstaring  Thanks Marius, but then, how would you suggest to exclude KNNGraphAlphaMutualInformation from ITKElastix?

mstaring:
Easy but ugly option is to explicitely specify which components are off and which on. More maintenance when we add a component though.

We could also define an extra functionality perhaps like DONOTUSE_COMPONENTS_ that do not have an apache compatible license, and then modify the cmake logic in the previously mentioned file.

Or we simply include this metric (for now) also in ITKElastix

N-Dekker:
Components that were already ON by default (n = 59):

	FixedImagePyramids\FixedGenericPyramid
	FixedImagePyramids\FixedRecursivePyramid
	FixedImagePyramids\FixedSmoothingPyramid
	FixedImagePyramids\OpenCLFixedGenericPyramid
	ImageSamplers\Full
	ImageSamplers\Grid
	ImageSamplers\MultInputRandomCoordinate
	ImageSamplers\RandomCoordinate
	ImageSamplers\RandomSparseMask
	ImageSamplers\Random
	Interpolators\BSplineInterpolator
	Interpolators\LinearInterpolator
	Interpolators\ReducedDimensionBSplineInterpolator
	Metrics\AdvancedMattesMutualInformation
	Metrics\AdvancedMeanSquares
	Metrics\AdvancedNormalizedCorrelation
	Metrics\BendingEnergyPenalty
	Metrics\CorrespondingPointsEuclideanDistanceMetric
	Metrics\DistancePreservingRigidityPenalty
	Metrics\KNNGraphAlphaMutualInformation
	Metrics\NormalizedMutualInformation
	Metrics\RigidityPenalty
	Metrics\SumOfPairwiseCorrelationsMetric
	Metrics\SumSquaredTissueVolumeDifferenceMetric
	Metrics\VarianceOverLastDimension
	MovingImagePyramids\MovingGenericPyramid
	MovingImagePyramids\MovingRecursivePyramid
	MovingImagePyramids\MovingSmoothingPyramid
	MovingImagePyramids\OpenCLMovingGenericPyramid
	Optimizers\AdaGrad
	Optimizers\AdaptiveStochasticGradientDescent
	Optimizers\AdaptiveStochasticLBFGS
	Optimizers\AdaptiveStochasticVarianceReducedGradient
	Optimizers\ConjugateGradient
	Optimizers\FiniteDifferenceGradientDescent
	Optimizers\FullSearch
	Optimizers\Powell
	Optimizers\PreconditionedStochasticGradientDescent
	Optimizers\QuasiNewtonLBFGS
	Optimizers\RegularStepGradientDescent
	Optimizers\StandardGradientDescent
	Registrations\MultiMetricMultiResolutionRegistration
	Registrations\MultiResolutionRegistrationWithFeatures
	Registrations\MultiResolutionRegistration
	ResampleInterpolators\BSplineResampleInterpolator
	ResampleInterpolators\RDBSplineResampleInterpolator
	Resamplers\MyStandardResampler
	Resamplers\OpenCLResampler
	Transforms\AdvancedAffineTransform
	Transforms\AdvancedBSplineTransform
	Transforms\AffineLogStackTransform
	Transforms\BSplineStackTransform
	Transforms\DeformationFieldTransform
	Transforms\EulerStackTransform
	Transforms\EulerTransform
	Transforms\RecursiveBSplineTransform
	Transforms\SplineKernelTransform
	Transforms\TranslationStackTransform
	Transforms\TranslationTransform

N-Dekker:
After applying this PR (both commits), the following five components will be the only ones OFF by default:

    Optimizers\AdaGrad
    Optimizers\AdaptiveStochasticLBFGS
    Optimizers\AdaptiveStochasticVarianceReducedGradient
    Optimizers\PreconditionedGradientDescent
    Transforms\BSplineDeformableTransformWithDiffusion

N-Dekker:
@mstaring @stefanklein Could you please double-check that this pull request would enable exactly the right set of components "by default", to be wrapped by ITKElastix?

PR #294, opened 2020-09-23
STYLE: Modernize code in BaseComponent.h

N-Dekker:
- Declare ComponentLabel getter and setter non-virtual (In C++, the `virtual` keyword should only be used when it is really useful!)
- Declare ConvertSecondsToDHMS `static`, as it does not use `this`.
- Default default-constructor and destructor (C++11).
- Use ITK_DISALLOW_COPY_AND_ASSIGN instead of "purposely not implemented" (ITK 5 coding convention)
- Move `#include`'s from BaseComponent.h to BaseComponent.cxx (to reduce compilation times)

### DIFF
--- a/Core/Install/elxBaseComponent.cxx
+++ b/Core/Install/elxBaseComponent.cxx
@@ -18,7 +18,9 @@
 
 #include "elxBaseComponent.h"
 
-#include <cmath>
+#include <cmath> // For fmod.
+#include <iomanip> // For setprecision.
+#include <sstream> // For ostringstream.
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


Core/Install/elxBaseComponent.cxx
```diff
@@ -88,7 +90,7 @@ void BaseComponent::InitializeElastixExecutable()
 
 std::string
 BaseComponent::ConvertSecondsToDHMS(
-  const double totalSeconds, const unsigned int precision = 0 ) const
+  const double totalSeconds, const unsigned int precision )
```
mstaring, line 93: are you sure removing the default here does not give issues?

mstaring:
looks good otherwise!

Core/Install/elxBaseComponent.cxx
N-Dekker: Thanks for asking, Marius. Yes, I'm sure it's OK to remove the default (`=0`) here, because it is in the CXX file, so it does not affect the interface! (I think the default was a mistake here. For the default to have any effect, it should have been in the H file.)

PR #296, opened 2020-09-23
STYLE: Modernize code of elxTransformBase

N-Dekker:
- Used `using` for typedef's from Superclass
- Defaulted default-constructor and destructor (C++11).
- Replaced "purposely not implemented" by `ITK_DISALLOW_COPY_AND_ASSIGN` calls
- Made three protected data members `private`
- Replaced a raw pointer by `std::unique_ptr`
- Did clean up the #include's.
- Used `const auto` for variables that are initialized by `New()`
- Replaced old for-loops in `AutomaticScalesEstimation` and `AutomaticScalesEstimationStackTransform` by modern C++11 range-based for-loops.

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.h
+++ b/Core/ComponentBaseClasses/elxTransformBase.h
@@ -27,8 +27,7 @@
 #include "elxComponentDatabase.h"
 #include "elxProgressCommand.h"
 
-#include <fstream>
-#include <iomanip>
+#include <memory> // For unique_ptr.
 
 namespace elastix
 {
@@ -130,6 +129,7 @@ class TransformBase :
   public BaseComponentSE< TElastix >
 {
 public:
+  ITK_DISALLOW_COPY_AND_ASSIGN(TransformBase);
 
   /** Standard ITK stuff. */
   typedef TransformBase               Self;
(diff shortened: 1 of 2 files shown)


Core/ComponentBaseClasses/elxTransformBase.h
```diff
@@ -374,7 +370,7 @@ class TransformBase :
   }
 
   /** Boolean to decide whether or not the transform parameters are written in binary format. */
-  bool m_UseBinaryFormatForTransformationParameters;
+  bool m_UseBinaryFormatForTransformationParameters{};
```
mstaring, line 373: should this not default to false? like in the current constructor?

mstaring:
looks good !

Core/ComponentBaseClasses/elxTransformBase.h
N-Dekker: Thanks for asking, Marius.
> should this not default to false?
 Indeed, and it does! `{}` is the default way to initialize a variable in C++11.  For an integer type, that is zero, for a `bool`, it is `false`. OK?

mstaring: clear!

PR #299, opened 2020-10-02
STYLE: Remove ViolaWellsMutualInformation and MutualInformationHistogram from Metrics components

N-Dekker:
These components appear obsolete, and may be removed from the repository, as was agreed upon during our internal Elastix sprint (with Stefan Klein and Marius Staring).

Note that these component were OFF by default already. They were marked "helemaal weg" by Marius (@mstaring) at https://github.com/SuperElastix/elastix/pull/291#issuecomment-696133404 to indicate that they may be entirely removed.


N-Dekker:
@mstaring @stefanklein Is this note from https://github.com/SuperElastix/elastix/blob/5.0.0/Components/Registrations/MultiMetricMultiResolutionRegistration/elxMultiMetricMultiResolutionRegistration.h#L56-L57 still relevant?

    Note that some metrics (ViolaWellsMutualInformation for example)
    do not support the image sampler functionality (yet).

mstaring:
agreed on the removal.

the remark I am not sure. I think you would need to check all metrics if they inherit from the advancedimagetoimagemetric.

N-Dekker:
> the remark I am not sure. I think you would need to check all metrics if they inherit from the advancedimagetoimagemetric.

It appears that (at least) `elastix::CorrespondingPointsEuclideanDistanceMetric<TElastix>` does not inherit from `itk::AdvancedImageToImageMetric<TFixedImage, TMovingImage>`. 

Does that mean that CorrespondingPointsEuclideanDistanceMetric does not support _the image sampler functionality_ ?

I could just remove "(ViolaWellsMutualInformation for example)" from the comment at  https://github.com/SuperElastix/elastix/blob/5.0.0/Components/Registrations/MultiMetricMultiResolutionRegistration/elxMultiMetricMultiResolutionRegistration.h#L56-L57  OK?

mstaring:
CorrespondingPointsEuclideanDistanceMetric  does not need samples, so that one is fine. If that is the only one then all metrics now support samplers (when needed). So the entire comment can go then.

N-Dekker:
> If that is the only one then all metrics now support samplers (when needed). So the entire comment can go then.

Thanks, but I do not know if CorrespondingPointsEuclideanDistanceMetric is the only one. It's just the first one I encountered, that does not inherit from AdvancedImageToImageMetric.

Issue #305, opened 2020-10-05
Elastix 5.0.1 binaries (elastix and transformix) not installed by default cmake build


bcdarwin:
I'm building on Ubuntu 16.04 with cmake 3.15.5 and gcc 5.4 with the default cmake settings.  After running 'make install' and looking in $CMAKE_INSTALL_PREFIX, there are `lib` and `include` directories but no `bin` containing `elastix` and `transformix`, although the same Cmake settings worked with `elastix 5.0.0`.

PR #309, opened 2020-10-06
BUG: Add elastix_exe and transformix_exe to CMake install (issue #305)

N-Dekker:
Aims to fix issue #305 "Elastix 5.0.1 binaries (elastix and transformix) not installed by default cmake build", reported by Ben Darwin (@bcdarwin).


N-Dekker:
@bcdarwin Hi Ben, can you please tell us if this pull request would fix the issue you reported, issue #305, on your Ubuntu machine? I tested only on Windows so far.

bcdarwin:
Yep, that fixes it.  Thanks!

PR #317, opened 2020-10-12
Overleaf

ViktorvdValk:
Manual is now as a submodule on a separate repo that is sync with Overleaf with Github Sync


N-Dekker (approved):
Great! Shall I squash the six commits into one, before merging them?

PR #318, opened 2020-10-12
STYLE: Explicitly instantiate xout class templates, declare them extern

N-Dekker:
Aims to reduce compilation times.


N-Dekker:
Closed because I did not observe a significant reduction of compilation times. (While it does add some extra lines of code. And I prefer less lines, instead of more lines of code.)

PR #319, opened 2020-10-15
ENH: Elastix library should read the initial transform from file

N-Dekker:
When the name of the initial transform is not "NoInitialTransform", and it cannot be interpreted as an index value (unsigned integer), the Elastix library should just assume that it is the file name of the transform file, and read it, in `TransformBase::ReadFromFile()`.

The elastix library and the elastix executable should deal with this in the same way.

As discussed with @ViktorvdValk and Marius @mstaring .

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.hxx
+++ b/Core/ComponentBaseClasses/elxTransformBase.hxx
@@ -448,13 +448,18 @@ TransformBase< TElastix >
   /** Call the function ReadInitialTransformFromFile. */
   if( fileName != "NoInitialTransform" )
   {
-    if (BaseComponent::IsElastixLibrary())
-    {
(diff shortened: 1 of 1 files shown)


mstaring:
lets discuss on monday :-)

N-Dekker:
TODO: Sync with elastix executable (inc. recursive filename check)
Assume index does not have trailing spaces

mstaring:
this looks nicely clean. I would add a comment explaining that if an integer then the FromVector funtion, and otherwise the FromFile function

mstaring:
did you test if it works yet?

ViktorvdValk:
https://github.com/N-Dekker/ITKElastix/actions/runs/315867799

The wheel from this commit in ITKElastix works.

PR #330, opened 2020-10-19
COMP: Trailing return type (auto) for xoutsimple GetXOutputs/GetCOutputs

N-Dekker:
Aims to fix ITKElastix Ubuntu build-linux-python-packages error:

> _deps/elx-src/Common/xout/xoutsimple.hxx:98:3: error: prototype for ‘const typename xoutlibrary::xoutsimple<charT, traits>::XStreamMapType& xoutlibrary::xoutsimple<charT, traits>::GetXOutputs()’ does not match any in class ‘xoutlibrary::xoutsimple<charT, traits>’

https://github.com/N-Dekker/ITKElastix/runs/1263271436#step:5:1863

### DIFF
--- a/Common/xout/xoutsimple.hxx
+++ b/Common/xout/xoutsimple.hxx
@@ -94,8 +94,8 @@ xoutsimple< charT, traits >::SetOutputs( const XStreamMapType & outputmap )
  */
 
 template< class charT, class traits >
-const typename xoutsimple< charT, traits >::XStreamMapType
-& xoutsimple< charT, traits >::GetXOutputs( void )
+auto xoutsimple< charT, traits >::GetXOutputs( void )
+-> const XStreamMapType &
 {
   return this->m_XTargetCells;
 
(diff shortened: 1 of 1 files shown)


N-Dekker:
The ITKElastix compilation error was produced on Ubuntu by GNU 4.8.2. It might be an issue that could be solved by upgrading the compiler 😃 SupeElastix/elastix currently has GCC 5.5.0 on its Ubuntu VM, which does not produce this compilation error.

N-Dekker:
@thewtex Hi Matt, It appears that ITKElastix still compiles on Ubuntu with a very very old C++ compiler, being GNU GCC 4.8.2. Which is why I encountered this compilation error ("prototype...does not match any in class"). It is fixed for now, but then... would you still want to keep supporting such an old compiler?

thewtex:
Hi Neils, thanks for the note and update. I will try the build with this patch to see if it passes. For the ITK 5.1 Python packages we need to support GCC 4.8.

N-Dekker:
@thewtex You're welcome, Matt.

> For the ITK 5.1 Python packages we need to support GCC 4.8.

Are there any plans to abandon the old GCC 4.8. Or even move forward to C++14? I feel it's very pitiful to be stuck to supporting such old compiler versions.

Issue #331, opened 2020-10-21
xout on `develop` branch do not build on Linux / GCC 4.8


thewtex:
As discussed in:

https://github.com/InsightSoftwareConsortium/ITKElastix/pull/78

and reported in:

https://github.com/InsightSoftwareConsortium/ITKElastix/runs/1282030109?check_suite_focus=true

recent changes on the `develop` branch are not building on Linux / GCC 4.8, which is required for Python packages.

N-Dekker:
Hi Matt,

Was this issue not fixed by pull request #330 ?

Commit https://github.com/SuperElastix/elastix/commit/9b4d0a11164daff6b4b44fb8a8f0d5d927457576

Please also check my question at https://github.com/SuperElastix/elastix/pull/330#issuecomment-712938116

thewtex:
Hi Neils,

It looks like #330 helped, but there are more build errors:

```
_deps/elx-src/Core/Kernel/elxElastixMain.cxx: In destructor ‘elastix::xoutManager::Guard::~Guard()’:
_deps/elx-src/Core/Kernel/elxElastixMain.cxx:146:10: error: use of deleted function ‘{anonymous}::Data& {anonymous}::Data::operator=({anonymous}::Data&&)’
   g_data = {};
          ^
_deps/elx-src/Core/Kernel/elxElastixMain.cxx:49:8: note: ‘{anonymous}::Data& {anonymous}::Data::operator=({anonymous}::Data&&)’ is implicitly deleted because the default definition would be ill-formed:
 struct Data
        ^
_deps/elx-src/Core/Kernel/elxElastixMain.cxx:49:8: error: use of deleted function ‘std::basic_ofstream<char>& std::basic_ofstream<char>::operator=(const std::basic_ofstream<char>&)’
In file included from ../../../ITK-source/ITK/Modules/IO/ImageBase/include/itkImageIOBase.h:38:0,
                 from ../../../ITK-source/ITK/Modules/IO/ImageBase/include/itkImageFileReader.h:24,
                 from _deps/elx-src/Core/Kernel/elxElastixBase.h:40,
                 from _deps/elx-src/Core/Kernel/elxElastixMain.h:24,
                 from _deps/elx-src/Core/Kernel/elxElastixMain.cxx:27:
[...]

PR #333, opened 2020-10-22
COMP: Avoid GCC4 error xoutManager "use of deleted function", issue #331

N-Dekker:
Worked around a GCC 4.8.2 compile error on ITKElastix build, reported by Matt McCormick, issue #331 (xout on `develop` branch do not build on Linux / GCC 4.8):

> _deps/elx-src/Core/Kernel/elxElastixMain.cxx: In destructor ‘elastix::xoutManager::Guard::~Guard()’:
> _deps/elx-src/Core/Kernel/elxElastixMain.cxx:146:10: error: use of deleted function ‘{anonymous}::Data& {anonymous}::Data::operator=({anonymous}::Data&&)’
...
> /opt/rh/devtoolset-2/root/usr/include/c++/4.8.2/fstream:599:11: error: use of deleted function ‘std::basic_ostream<char>& std::basic_ostream<char>::operator=(const std::basic_ostream<char>&)’

### DIFF
--- a/Core/Kernel/elxElastixMain.cxx
+++ b/Core/Kernel/elxElastixMain.cxx
@@ -143,7 +143,13 @@ xoutManager::xoutManager( const std::string& logFileName, const bool setupLoggin
 
 xoutManager::Guard::~Guard()
 {
-  g_data = {};
+  // Destruct and reconstruct the data, in order to reset the output streams.
+  // Note: a simple assignment, `g_data = {}` would be preferable, but doing
+  // so triggered compilation errors on Linux / GCC 4.8:
+  // error: use of deleted function 'std::basic_ofstream<char>&
+  // std::basic_ofstream<char>::operator=(const std::basic_ofstream<char>&)'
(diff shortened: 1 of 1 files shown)


N-Dekker:
@thewtex Matt do you think this pull request would make GCC 4.8 happy?

thewtex:
Thanks, Neils! Testing it out here: https://github.com/InsightSoftwareConsortium/ITKElastix/pull/78

N-Dekker:
@thewtex Do I understand correctly from https://github.com/InsightSoftwareConsortium/ITKPythonPackage/pull/152#issuecomment-740684513 that this (rather tricky, rather ugly) workaround is no longer necessary, as GCC 4.8 no longer needs to be supported? 

We currently have GCC 7.5.0 on our SuperElastix/elastix CI (Ubuntu), which does not need this workaround.

thewtex:
@N-Dekker yes :-)

PR #336, opened 2020-10-26
Reformat source code, using ITK clang-format file

N-Dekker:


dzenanz (commented):
You might also want to apply this to *.in files (maybe manually see which ones are templates for C++ code, which ones for other things like CMake).

N-Dekker:
> extra spaties?

@mstaring Thanks for having a check, Marius. I think the spaces that break `#include` into `#  include` are intended, as well in ITK, for example:

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.1/Modules/Core/Common/include/itkAtanRegularizedHeavisideStepFunction.h#L87

N-Dekker:
> You might also want to apply this to *.in files

@dzenanz Thanks for your suggestion. Many elastix *.in files do not have C++ code, but the following do:  

 - https://github.com/SuperElastix/elastix/blob/5.0.1/Common/OpenCL/ITKimprovements/itkOpenCLKernels.h.in
 - https://github.com/SuperElastix/elastix/blob/5.0.1/Components/Metrics/KNNGraphAlphaMutualInformation/KNN/ann_1.1/include/ANN/ANN.h.in
 - https://github.com/SuperElastix/elastix/blob/5.0.1/Core/elxVersionMacros.h.in
 - https://github.com/SuperElastix/elastix/blob/5.0.1/Core/Install/elxOpenCLSupportedImageTypes.h.in
 - https://github.com/SuperElastix/elastix/blob/5.0.1/Core/Install/elxSupportedImageTypes.h.in

However, it appears that clang-format breaks the CMake configure_file `@VAR@` variables. Did you also encounter that problem? For now I prefer to skip these files, but maybe they can be clang-formatted later.

dzenanz:
Perhaps clang-format them once, then manually fix the @ breakages? That way the .in files will have the approximately correct style.

mstaring:
approved

N-Dekker:
> Perhaps clang-format them once, then manually fix the @ breakages?

@dzenanz Thanks for the suggestion but for the sake of time, I would like to just merge this PR now. However, I don't mind having another look at those in files later. I noticed the ITK *.in files also seem to be skipped by clang-format, for example: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.1/CMake/itkImageIOFactoryRegisterManager.h.in#L22 should have the `{` at the next line.

dzenanz:
Go ahead and merge. Later, having a pass at ITK's .in files would also be appreciated 😄

mstaring:
could we use wildcards so that only .in files in certain directories get processed? or only in some directories they get excluded?

N-Dekker:
> could we use wildcards so that only .in files in certain directories get processed?

@mstaring  The five C++ *.in files that were not included yet are all "*.h.in" files:
- Common/OpenCL/ITKimprovements/itkOpenCLKernels.h.in
- Components/Metrics/KNNGraphAlphaMutualInformation/KNN/ann_1.1/include/ANN/ANN.h.in
- Core/elxVersionMacros.h.in
- Core/Install/elxOpenCLSupportedImageTypes.h.in
- Core/Install/elxSupportedImageTypes.h.in

 For batch processing, it should be possible to only process "*.h.in" files. (Once we have figured out how to stop clang-format from messing up the `@VAR@` variables.)

However, I don't know yet how to tell the editor/IDE and the commit hook (to be added) that it should do so.

PR #337, opened 2020-10-27
Fix CTest options at Azure Pipelines CI 

N-Dekker:


N-Dekker:
@kaspermarstal How are you? You may find it interesting to know that `ctest --config Release` does not work as expected! Fortunately  `ctest -C Release` does the job!

PR #339, opened 2020-10-28
COMP: Fix include-guards in header files, remove umlauts

N-Dekker:
Removed two umlauts, to address a "File Load" error message box in Visual Studio 2019, saying:

> Some bytes have been replaced with the Unicode substitution character while loading file with Unicode (UTF-8) encoding. Saving the file will not preserve the original file contents.

Fixed include-guards in header files, to comply to the C++ Standard, which says:

> Each identifier that contains a double underscore `__` or begins with an underscore followed by an
uppercase letter is reserved to the implementation for any use.

(In this context, "the implementation" refers to the compiler and the Standard Library implementation.)

Following ITK commit https://github.com/InsightSoftwareConsortium/ITK/commit/dd673a7a7b32f2d7fc2278c0f8674b40a6272e48 by Sean McBride (@seanm), December 2014.


seanm (commented):
That's odd. Is the file UTF-8? If you save it, and it therefore "doesn't preserve the original", what do you actually end up with?

N-Dekker:
> That's odd. Is the file UTF-8? If you save it, and it therefore "doesn't preserve the original", what do you actually end up with?

@seanm Thanks for asking, Sean. I think the original author of the source file assumed a certain code page for 8-bits non-ASCII characters, like [Windows-1250](https://en.wikipedia.org/wiki/Windows-1250). Interestingly, the umlaut looks fine in my (Firefox) browser, looking at https://github.com/SuperElastix/elastix/blob/5.0.1/Components/Metrics/MissingStructurePenalty/elxMissingStructurePenalty.h#L38 but it looks weird in my Windows 10 Notepad. In Notepad, I got a ["black diamond replacement character"](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) at the location of the umlaut.

After saving in Visual Studio, it's still a black diamond. So I don't know if it's really an improvement, or if it just makes things worse.

I looked at it before, the Visual Studio error message box annoys me every time, and for now the workaround of replacing `ü` by `ue` looks good enough to me.

N-Dekker:
If no one objects, I'll merge this once it has passed the CI tests.

PR #340, opened 2020-10-29
STYLE: Declare purposely not implemented functions by `= delete` and fix documentation

N-Dekker:


N-Dekker:
If no one objects, I would like to merge this once it has passed the CI tests.

mstaring:
no objections, you are the master here :-)

Did ITK make the same changes in their repo? As we simply mimick ITK style here.

N-Dekker:
> no objections, you are the master here :-)
 
Thanks, Marius.

> Did ITK make the same changes in their repo? As we simply mimick ITK style here.

Actually ITK is two steps ahead here. ITK has declared these "deleted" member functions  `public`, instead of the `private`, as that is recommended nowadays (for modern C++).

And then ITK consistently declared these "deleted" member functions by means of a macro, which has just been renamed from `ITK_DISALLOW_COPY_AND_ASSIGN(TypeName)` to `ITK_DISALLOW_COPY_AND_MOVE(TypeName)` (for the next ITK release): 
https://github.com/InsightSoftwareConsortium/ITK/commit/7d177ca0c77aedb38cc8cf5561b785317617c315 

The name change is my fault 😺 As well as ITK following the convention to declare deleted member functions `public`. 

Maybe we could start using this new macro in elastix as well, once the next ITK release is out! But I think this pull request is already a step in the right direction.

PR #350, opened 2020-11-05
STYLE: Replace `v[v.size() - 1]` by `v.back()`, in elastix Core

N-Dekker:
Improved code readability and simplicity by using `std` member function `back()`.


N-Dekker:
For the record, five `std` container types have a `back()` member function: `std::string`, `std::vector`, `std::list`, `std::array`, and `std::deque`. It appears to be quite useful 😃

PR #360, opened 2020-11-18
ENH: specify required ITK components explicitly

dzenanz:
This avoids circular dependency when building ITKElastix as part of ITK.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,8 +22,32 @@ include( CTest )
   #set( CMAKE_CXX_VISIBILITY_PRESET hidden )
 
 #---------------------------------------------------------------------
+option( ELASTIX_USE_OPENCL "Use OpenCL enabled GPU" OFF )
+set(_GPU_depends "")
+if(ELASTIX_USE_OPENCL)
+  list(APPEND _GPU_depends ITKGPUCommon)
+endif()
+
 # Find ITK.
-find_package( ITK 5.1.1 REQUIRED )
+find_package( ITK 5.1.1 REQUIRED COMPONENTS
+    ITKCommon
+    ITKDistanceMap
+    ITKImageGrid
+    ITKImageIntensity
+    ITKIOImageBase
+    ITKIOMeshBase
+    ITKIOMeshOBJ
+    ITKIOMeta
+    ITKIOXML
+    ITKMathematicalMorphology
+    ITKOptimizers
+    ITKRegistrationCommon
+    ITKSmoothing
+    ITKTestKernel
+    ITKThresholding
+    ITKTransform
+    ${_GPU_depends}
+    )
 include( ${ITK_USE_FILE} )
 
 #---------------------------------------------------------------------
@@ -40,8 +64,6 @@ include(elastixExportTarget)
 
 #---------------------------------------------------------------------
 # Find OpenCL.
-option( ELASTIX_USE_OPENCL "Use OpenCL enabled GPU" OFF )
-
 if( ELASTIX_USE_OPENCL )
   find_package( OpenCL REQUIRED QUIET )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,21 +31,32 @@ endif()
 # Find ITK.
 find_package( ITK 5.1.1 REQUIRED COMPONENTS
     ITKCommon
+    ITKDisplacementField
     ITKDistanceMap
+    ITKGDCM
+    ITKImageCompose
+    ITKImageFunction
+    ITKImageGradient
     ITKImageGrid
     ITKImageIntensity
+    ITKImageStatistics
     ITKIOImageBase
     ITKIOMeshBase
     ITKIOMeshOBJ
     ITKIOMeta
+    ITKIOTransformBase
     ITKIOXML
     ITKMathematicalMorphology
+    ITKMesh
     ITKOptimizers
     ITKRegistrationCommon
     ITKSmoothing
+    ITKSpatialObjects
+    ITKStatistics
     ITKTestKernel
     ITKThresholding
     ITKTransform
+    ITKTransformFactory
     ${_GPU_depends}
     )
 include( ${ITK_USE_FILE} )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,9 @@ find_package( ITK 5.1.1 REQUIRED COMPONENTS
     ITKTransform
     ITKTransformFactory
     ${_GPU_depends}
+    ITKImageIO
+    ITKTransformIO
+    ITKMeshIO
     )
 include( ${ITK_USE_FILE} )
 


dzenanz:
This is now ready for review/merging.

mstaring:
@N-Dekker could you review?

CMakeLists.txt
```diff
@@ -23,7 +23,24 @@ include( CTest )
...
+    ITKIOMeshOBJ
+    ITKIOMeta
+    ITKIOXML
+    ITKMathematicalMorphology
+    ITKOptimizers
+    ITKRegistrationCommon
+    ITKSmoothing
+    ITKTestKernel
+    ITKThresholding
+    ITKTransform
```
N-Dekker, line 42: Thanks @dzenanz A question: are these _all_ ITK components? Or are these exactly the ones needed by elastix?

dzenanz: These are the components required by elastix to compile for me. ITK has around 100 components total. Some components will be included indirectly.

List of supported IO formats should be specified explicitly now. I only included .obj mesh format (ITKIOMeshOBJ) and .mha/.mhd image format (ITKIOMeta). You will want to expand that to include all of the [IO modules](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Modules/IO), or at least a significant subset.

I will amend the commit to optionally include ITK GPU module(s).

CMakeLists.txt
```diff
@@ -22,8 +22,32 @@ include( CTest )
   #set( CMAKE_CXX_VISIBILITY_PRESET hidden )
 
 #---------------------------------------------------------------------
+option( ELASTIX_USE_OPENCL "Use OpenCL enabled GPU" OFF )
+set(_GPU_depends "")
+if(ELASTIX_USE_OPENCL)
+  list(APPEND _GPU_depends ITKGPUCommon)
```
dzenanz, line 28: More ITKGPUXXX modules might need to go here. Or maybe this is not needed as elastix has its own separate OpenCL-related infrastructure?

N-Dekker:
I'm sorry I don't entirely understand the issue. Is the idea to have ITKElastix as remote module of ITK? If so, for what use case would that be?

dzenanz:
> ITKElastix as remote module of ITK

Yes. The use case is [another remote module depending on ITKElastix](https://github.com/dzenanz/HASI/commit/6a2f6b9fc9da61b62475eb33d10630540923207c), and that is easily accomplished only if all the dependencies are properly defined as [remotes in ITK](https://github.com/dzenanz/ITK/commit/c8e5ba1ea81061a1661c2fcc00b4e815dbedc606).

N-Dekker:
> Yes. The use case is [another remote module depending on ITKElastix](https://github.com/dzenanz/HASI/commit/6a2f6b9fc9da61b62475eb33d10630540923207c)

Can't the other remote module depend directly on the C++ SuperElastix/elastix.lib ?

dzenanz:
Of course. But having ITK remotes as only dependencies simplifies the build.

Another benefit of specifying required components explicitly is smaller build size. Perhaps not significant if one is trying to include all IOs provided by ITK.

Decide whether to reject the proposed change or suggest some improvements. I guess it is OK if you don't want the hassle of dealing with ITK's components, even if it enables this new use-case.

N-Dekker:
@dzenanz Before I can decide, I need to have a better understanding. When the components are not specified explicitly, `find_package(ITK 5.1.1 REQUIRED)` finds the remote modules as well, right? Is there no way to specify that elastix only links the "non-remote" modules?

A smaller build size is not a main aim for elastix, as far as I know. Do you think the elastix build size should be smaller?

Personally I would very much appreciate a _faster_ build though. I can imagine that your PR might reduce link time.

dzenanz:
`find_package(ITK 5.1.1 REQUIRED)` finds all enabled modules in the chosen build of ITK. There is no way to specify only non-remote modules. Smaller build size is advantageous for any software. But I don't think that this PR will significantly reduce either elastix build size (unless only a few IOs are selected) or build/link time. Majority of ITK build time and size comes from compiled third party libraries and compiled IO dependencies (VNL, GDCM, OpenJPEG, zlib etc), not ITK components themselves.

N-Dekker:
OK, thanks for explaining @dzenanz Now if we merge your PR, can you elaborate a little bit on how to maintain the `find_package( ITK 5.1.1 REQUIRED COMPONENTS...)` list of components, in the future? Is it always detectable at compile time (or link time) that the list of components is still complete? If not, how to figure out which component is missing?

dzenanz:
The errors always arise at compile/link time. The missing component can be found out by looking at the documentation of the missing include file. For example, if compile error says "cannot find itkBinShrinkImageFilter.h", take a look at [BinShrinkImageFilter](https://itk.org/Doxygen/html/classitk_1_1BinShrinkImageFilter.html)'s documentation. Near the top it says the module to which it belongs (Module ITKImageGrid in this case). So `ITKImageGrid` needs to be added to the list.

If module list is required for one, or a small number of source files, [WhatModulesITK.py](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/WhatModulesITK.py) can be used.

List of desired IOs cannot be managed this way. Absence of any concrete IO will not raise compile errors, but the corresponding file format will be unsupported. Meta-groups `ITKImageIO` and `ITKTransformIO` (described in [4.13 release notes](https://github.com/InsightSoftwareConsortium/ITK/blob/9c0cdee0b6052da4840d3cb33d20a253b3cccde9/Documentation/ReleaseNotes/4.13.md#introduction) under heading "Improved Factory Registration") can be used for external builds (the usual way elastix is built), but currently not for ITK modules. When building as part of ITK, explicit list of IO modules needs to be specified.

N-Dekker:
Can you possibly amend the PR in order to be able to use `itk::TransformFactoryBase::GetFactory()`?

ViktorvdValk:
A search of the elastix repo for include statements containing itk and IO resulted in this:
['itkImageIOBase.h', 'itkImageIORegion.h', 'itkImageIOFactory.h', 'itkMetaImageIO.h', 'itkMevisDicomTiffImageIO.h', 'itkMevisDicomTiffImageIOFactory.h']

ViktorvdValk:
Another search for all itk modules that are used in the includes statements of all files in the elastix repo, which are not already in the repo itself, resulted in the following list:

ITKCommon
ITKImageFunction
ITKMesh
ITKSpatialObjects
ITKTestKernel
ITKTransform
ITKDisplacementField
ITKDistanceMap
ITKImageCompose
ITKImageFilterBase
ITKImageGradient
ITKImageGrid
ITKImageIntensity
ITKImageStatistics
ITKMathematicalMorphology
ITKSmoothing
ITKThresholding
ITKIOImageBase
ITKIOMeshBase
ITKIOMeta
ITKIOXML
ITKOptimizers
ITKStatistics
ITKRegistrationCommon
ITKGDCM
ITKVNL

This would mean that the following modules are still missing in the CMakeList.txt:

ITKImageFunction
ITKMesh
ITKSpatialObjects
ITKDisplacementField
ITKImageCompose
ITKImageFilterBase
ITKImageGradient
ITKImageStatistics
ITKStatistics
ITKGDCM
ITKVNL

Could this be correct @dzenanz ?
And can these just be added to the CMakeList.txt or not?

dzenanz:
Some of these are indirectly included as dependencies of explicitly requested module. I am pretty sure `ITKImageFilterBase` is required by something, and therefore already included. I think `ITKVNL` is a dependency of `ITKCommon`, so it is always included. The rest should be added to the explicit COMPONENTS list.

I checked "Allow edits and access to secrets by maintainers". Does somebody want to take over this PR?

ViktorvdValk:
Yes, I'll update the cmakelist.txt

N-Dekker:
@ViktorvdValk Thanks! Before you press the "Rebase and merge" button (when it is ready), can you please locally do a "git squash", to make it one single commit? And then `git push --force-with-lease`, of course  😃 

https://blog.carbonfive.com/always-squash-and-rebase-your-git-commits/

ViktorvdValk:
@dzenanz  How should itkTransformIOFactoryRegisterManager.h be included, since it is not part of a module, but it is a dependency that is necessary since the CI now fails?

dzenanz:
I guess it should be included as a regular include file, as it gets generated in the build directory. What is the error message? Manually fiddling with that is usually not required.

dzenanz:
It should be in the include path. Do you know why [this](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.2/CMake/UseITK.cmake#L232) has no effect?

ViktorvdValk:
elastix is caching ITK v5.1.1. but the line you pointed out was already part of that version. Building elastix with ITK v5.1.2. might solve the issue, but other than that I have no clue, but I'm far from an expert in CMake. Maybe @N-Dekker, you have suggestions?

N-Dekker:
> elastix is caching ITK v5.1.1. but the line you pointed out was already part of that version. Building elastix with ITK v5.1.2. might solve the issue, but other than that I have no clue, but I'm far from an expert in CMake.

@ViktorvdValk  Honestly I have no clue either, but an upgrade to ITK v5.1.2 on the SuperElastix/elastix CI would certainly be fine to me. Please go ahead, if you think it might help  😃

CMakeLists.txt
```diff
@@ -22,8 +22,43 @@ include( CTest )
...
+endif()
+
 # Find ITK.
-find_package( ITK 5.1.1 REQUIRED )
+find_package( ITK 5.1.1 REQUIRED COMPONENTS
```
thewtex, line 32: We likely want to loosen the version requirement, i.e. `5.1.1` -> `5` or `5.1`.

CMakeLists.txt
thewtex: Was the [`WhatModulesITK.py`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/WhatModulesITK.py) script used to determine these? Modules could be missing.

Also, for the elastix CLI, we should include all the IO modules by including `ITKImageIO`, `ITKMeshIO`, and `ITKTransformIO`.

thewtex (approved):

CMakeLists.txt
N-Dekker: > We likely want to loosen the version requirement, i.e. 5.1.1 -> 5 or 5.1

@thewtex Why would you? It may change the behavior of elastix, for example when the following ITK bug fix may not be assumed https://github.com/InsightSoftwareConsortium/ITK/issues/1950

CMakeLists.txt
ViktorvdValk: I wrote another script to determine the ITK modules elastix is using

ViktorvdValk: @thewtex I can't find the modules you mentioned?
The closest I got was: ITKIOImageBase and ITKIOTransformBase (which are already included).
I did find ITKIOMesh, which I'll include.

dzenanz: Those are meta-modules. First introduced with [4.13](https://github.com/InsightSoftwareConsortium/ITK/blob/9c0cdee0b6052da4840d3cb33d20a253b3cccde9/Documentation/ReleaseNotes/4.13.md#introduction).

ViktorvdValk: Yes, but I can't find these modules (ITKImageIO and IKTTransformIO) on https://itk.org/Doxygen/html/modules.html which should list all available modules, right?

dzenanz: I added them.

dzenanz:
I squashed and rebased this branch.

CMakeLists.txt
thewtex: Yes, elastix has its own internal OpenCL code / classes.

CMakeLists.txt
thewtex: An ITK installation may be available that is newer, e.g. 5.1.2, 5.2rc01, and we can enable use of that version.

mstaring:
What is the status of this PR?

dzenanz (commented):
I think this is ready to merge.

N-Dekker:
Just tested your branch locally on VS2019 with ELASTIX_USE_OPENCL, and the three optional OpenCL components (USE_OpenCL...). (Using "C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.1/lib/x64/OpenCL.lib") Everything compiles well, two test failures but I guess they are unrelated to your pull request:

>	 30 - OpenCLSimpleTest (Failed)
>	 49 - GPUResampleImageFilterTest-BSplineBSpline (Failed)

N-Dekker:
@mstaring @stefanklein I just compared the Visual Studio 2019 project settings from this pull request to the current develop branch. While both versions build fine, it looks like the pull request would remove the following 25 library dependencies. Is that OK to you? Just want to be sure, so that we can merge then!

    ITKBiasCorrection-5.1.lib
    ITKColormap-5.1.lib
    ITKConvolution-5.1.lib
    ITKDeformableMesh-5.1.lib
    ITKDenoising-5.1.lib
    ITKDICOMParser-5.1.lib
    ITKDiffusionTensorImage-5.1.lib
    ITKFastMarching-5.1.lib
    ITKFFT-5.1.lib
    ITKIOCSV-5.1.lib
    ITKIOSiemens-5.1.lib
    ITKIOSpatialObjects-5.1.lib
    ITKKLMRegionGrowing-5.1.lib
    itklbfgs-5.1.lib
    ITKMarkovRandomFieldsClassifiers-5.1.lib
    ITKOptimizersv4-5.1.lib
    ITKPDEDeformableRegistration-5.1.lib
    ITKPolynomials-5.1.lib
    ITKQuadEdgeMeshFiltering-5.1.lib
    ITKRegionGrowing-5.1.lib
    ITKRegistrationMethodsv4-5.1.lib
    ITKVideoCore-5.1.lib
    ITKVideoIO-5.1.lib
    ITKVTK-5.1.lib
    ITKWatersheds-5.1.lib

If these 25 lib files are not meant to be used by elastix, it's fine to me, of course.

stefanklein:
Hi, I don't think any of these libraries are needed indeed. Best, Stefan

PR #362, opened 2020-11-23
Github Actions to replace Azure Pipelines

ViktorvdValk:
Besides the Azure Pipeline workflow, the Github Actions workflow includes:

- caching of ITK by version number to speed up build time
- uploading the executables (elastix and transformix) per operating system as artifacts.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,158 @@
+name: CMake
+
+on: [push, pull_request]
+  
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 3
+      matrix:
+        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        include:
+          - os: ubuntu-18.04
+            c-compiler: "gcc"
+            cxx-compiler: "g++"
+            itk-git-tag: "v5.1.1"
+            cmake-build-type: "MinSizeRel"
+            ANNLib: "libANNlib-5.0.so"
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,158 @@
...
+    strategy:
+      max-parallel: 3
+      matrix:
+        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        include:
+          - os: ubuntu-18.04
+            c-compiler: "gcc"
+            cxx-compiler: "g++"
+            itk-git-tag: "v5.1.1"
+            cmake-build-type: "MinSizeRel"
```
N-Dekker, line 22: @ViktorvdValk Is there a specific reason for "MinSizeRel", instead of simply "Release"?  As we do at Azure Pipelines: https://github.com/SuperElastix/elastix/blob/5.0.1/Testing/CI/Azure/ci.yml#L124

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,158 @@
...
+            ANNLib: "libANNlib-5.0.so"
+            ANNLib2: "libANNlib-5.0.so.1"
+          - os: windows-2019
+            c-compiler: "cl.exe"
+            cxx-compiler: "cl.exe"
+            itk-git-tag: "v5.1.1"
+            cmake-build-type: "Release"
```
N-Dekker, line 29: How does this work, here it says cmake-build-type: "Release", does that overrule the previous cmake-build-type?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,158 @@
...
+            c-compiler: "gcc"
+            cxx-compiler: "g++"
+            itk-git-tag: "v5.1.1"
+            cmake-build-type: "MinSizeRel"
+            ANNLib: "libANNlib-5.0.so"
+            ANNLib2: "libANNlib-5.0.so.1"
+          - os: windows-2019
+            c-compiler: "cl.exe"
+            cxx-compiler: "cl.exe"
+            itk-git-tag: "v5.1.1"
```
N-Dekker, line 28: Is it necessary to mention the ITK version here again, as it is already on line 21?

PR #372, opened 2020-12-11
STYLE: add clang format check

ViktorvdValk:
The proposed style changes are given as output of GitHub actions, which fails if files do not adhere to the clang format file.

### DIFF
--- a/.github/workflows/clang_format_check.yml
+++ b/.github/workflows/clang_format_check.yml
@@ -1,0 +1,17 @@
+name: test-clang-format
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.11
+      with:
+        source: '.'
+        exclude: './CMake'
+        extensions: 'hxx,h,cxx'
+        clangFormatVersion: 8
+        style: file


.github/workflows/clang_format_check.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: DoozyX/clang-format-lint-action@v0.11
+      with:
+        source: '.'
+        exclude: './CMake'
+        extensions: 'hxx,h,cxx'
+        clangFormatVersion: 8
```
N-Dekker, line 16: Did you also try  `clangFormatVersion: 8.0.1` ? I'm not sure if it matters much, but ITK uses 8.0.1, and I also use 8.0.1 locally.

N-Dekker: ( ITK has their clang version archived at https://data.kitware.com/#collection/57b5c9e58d777f126827f5a1/folder/5d635cbcd35580e6dcbd80b6 )

ViktorvdValk: this GitHub action doesn't allow for clangFormatVersion 8.0.1. but it uses the file in the repo and it seems to work the same.

N-Dekker: OK, thanks, maybe version 8 is just fine. Then I still wonder why ITK upgraded from clang-format 8 to 8.0.1.

N-Dekker: @thewtex Matt do you remember any important difference between clang-format 8.0.0 and 8.0.1? We see, you specifically chose 8.0.1, at pull request https://github.com/InsightSoftwareConsortium/ITK/pull/1203 commit https://github.com/InsightSoftwareConsortium/ITK/commit/485d1b3ee68e7f184ab25584963a0f76c1f7123b

Would it be OK for elastix to stick to clang-format 8.0.0?

N-Dekker: For the record, the GitHub Action "clang-format-lint-action" has an [MIT License](https://github.com/DoozyX/clang-format-lint-action/blob/master/LICENSE), Copyright (c) 2019 Slobodan Kletnikov (@DoozyX). I think that's just fine, I think we may directly use it without possible legal problems. (Disclaimer: I'm not a lawyer.)

DoozyX: Hello, just received the notification, versions from 5-9, are actually the patched versions. In your case if you type version 8 it will use the clang-format from clang 8.0.1

N-Dekker: > versions from 5-9 are actually the patched versions. In your case if you type version 8 it will use the clang-format from clang 8.0.1

@DoozyX That's really great, Slobodan! Thank you very much!

thewtex: @N-Dekker I recall clang-format was quite specific to versions, but I do not recall the differences between 8.0.0 and 8.0.1.

PR #373, opened 2020-12-11
STYLE: Revert workaround GCC4 error xoutManager "use of deleted func..."

N-Dekker:
This reverts commit c106327239f2d2a0ef6962f00f86f8ae41e8781f "COMP: Avoid GCC4 error xoutManager "use of deleted function", issue #331"

Matt McCormick (@thewtex) informed us that GCC 4 no longer needs to be supported, as ITKElastix is upgraded to manylinux2014, which allows using GCC 9 instead.


N-Dekker:
For the record, this pull request reverts https://github.com/SuperElastix/elastix/pull/333 as I suggested at https://github.com/SuperElastix/elastix/pull/333#issuecomment-742760679

PR #376, opened 2020-12-20
Remove bitwise AND assignment of literal bool values (true and false)

N-Dekker:
For any `bool` variable `b`, a statement of the form `b &= true` has no effect at all.

`b &= false` has exactly the same effect as a simple `b = false` assignment.

### DIFF
--- a/Common/CostFunctions/itkImageToImageMetricWithFeatures.h
+++ b/Common/CostFunctions/itkImageToImageMetricWithFeatures.h
@@ -244,7 +244,6 @@ protected:
   MovingFeatureInterpolatorVectorType m_MovingFeatureInterpolators;
 
   std::vector<bool>                    m_FeatureInterpolatorsIsBSpline;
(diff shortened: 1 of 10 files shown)


N-Dekker:
@mstaring @stefanklein Can you please check the assignments to `m_FeatureInterpolatorsAreBSpline` in the following lines of code, in `ImageToImageMetricWithFeatures::CheckForBSplineFeatureInterpolators(void)`? 
https://github.com/SuperElastix/elastix/blob/8ac0413959562957e3f389eb9cf2204acbf7fe69/Common/CostFunctions/itkImageToImageMetricWithFeatures.hxx#L309-L328

It appears that `m_FeatureInterpolatorsAreBSpline` is never set to `false` here. Was that the intention? It looks like a bug to me. If so, is it an important piece of code?

N-Dekker:
I will do Rebase-and-merge now 😃, as siscussed at the latest (internal) weekly sprint meeting, with @stefanklein and @ViktorvdValk.

Issue #329, opened 2020-10-19
Automate C++ source code formatting


N-Dekker:
Discussed with Marius @mstaring

N-Dekker:
This is what it looks like if elastix uses the clang-format configuration file from ITK: https://github.com/SuperElastix/elastix/tree/ITK-clang-format

N-Dekker:
Partially addressed by pull request  #336 (Reformat source code, using ITK clang-format file): The automation now depends on the IDE (e.g., Visual Studio) being configured to use clang-format. Or otherwise run clang-format directly, as in commit https://github.com/SuperElastix/elastix/commit/c3a254562d6801f1a34a2de0ec324c10632b5884

    find . -regex '.*\.\(hxx\|h\|cxx\)' -exec /path/to/clang-format.exe -style=file -i {} \; 

Another automation step could be by applying clang-format to a PR, as ITK does at https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.1/.github/workflows/apply-clang-format.yml

mstaring:
Yes, let us also do it automatically for PRs. Is that easy to setup?

N-Dekker:
Hans Johnson suggested me to have a look at https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.1/CMake/ITKClangFormatConfig.cmake.in (Thanks, @hjmjohnson !) To be continued!

ViktorvdValk:
This [PR](https://github.com/SuperElastix/elastix/pull/372) should close this issue @N-Dekker 
Or do we want further clang formatting such as the automated formatting instead of comments in GitHub actions?

N-Dekker:
Honestly, I think it would still be useful to have a pre-commit hook to do the formatting, especially to make life easier for elastix users who want to contribute. (For contributors it might be frustrating to see GitHub Actions reject their pull request because of formatting... even though its great that we have such a GitHub Action now  😃 )

PR #380, opened 2020-12-28
STYLE:clangFormat pre-commit hook

ViktorvdValk:
- A pre-commit hook for local git usage on both Windows and Unix systems.
- Pre-commit hook use update in CONTRIBUTING.md

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,35 @@
+#! /bin/sh
+diff_files=`git diff --cached --name-only | grep -E '\.cxx$|\.hxx$|\.h$' `
+if [ "$(uname)" == "Darwin" ]; then
+  clangFile=.githooks/clangFormatMac
+  if [ -f "$clangFile" ]; then
+    echo "clangFormat already downloaded."
+  else
+    curl -o .githooks/clangFormatMac https://data.kitware.com/api/v1/file/5d274e88877dfcc902effc47/download
+    chmod u+x .githooks/clangFormatMac
+  fi
+  for file in $diff_files; do
+    .githooks/clangFormatMac -style=file -i --verbose $file;
+  done
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+  clangFile=.githooks/clangFormatLinux
+  if [ -f "$clangFile" ]; then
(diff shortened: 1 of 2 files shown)


ViktorvdValk:
This would close #329

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -18,20 +18,20 @@ jobs:
           - os: ubuntu-18.04
             c-compiler: "gcc"
             cxx-compiler: "g++"
-            itk-git-tag: "v5.1.1"
```
N-Dekker, line 21: Should this ITK upgrade not be done in a separate commit or PR? Or is it related to clang-format?

ViktorvdValk: Yes, this was already done in commit 98aaaf7e27c7dacd271db5a2cc823b938a154392, so I don't get why it still shows up here.
But no it's not related to clang-format.

N-Dekker: git happens 😸

mstaring:
I think a rebase is needed

ViktorvdValk:
I included the clang-format binaries in the repo now, which is not really conventional.
However the other option is downloading them at a git clone,
but than you rely on an external website, which I found trickier than just including them,
since they're not supposed to be updated anymore.

N-Dekker (changes requested):
Sorry, I'm still not so sure if it's OK to store those binaries in the elastix git repo. Especially because they may need to be updated now and then, which would quickly make an elastix clone bigger and bigger. Did you also have a look at how ITK did it?

ViktorvdValk:
@N-Dekker now the binaries are downloaded from kitware first in the bash script, so they're not stored on github anymore.

N-Dekker:
> now the binaries are downloaded from kitware first in the bash script, so they're not stored on github anymore.

Cool, I'm gonna give it a try!

N-Dekker (approved):
If possibly, could you still figure out how to print a message, saying something like "Info: clang-format made some changes after the commit", whenever clang-format has actually modified the source code through this hook?

Otherwise, just merge! I like it very much :-)

PR #379, opened 2020-12-22
Add missing parameters to transform parameter maps

N-Dekker:

PR #385, opened 2021-01-08
Sync Transform WriteToFile with CreateTransformParametersMap

N-Dekker:


Components/Transforms/DeformationFieldTransform/elxDeformationFieldTransform.hxx
```diff
@@ -159,9 +159,6 @@ DeformationFieldTransform<TElastix>::WriteToFile(const ParametersType & param) c
 
   typedef itk::ChangeInformationImageFilter<DeformationFieldType> ChangeInfoFilterType;
 
-  /** Add some DeformationFieldTransform specific lines. */
-  xout["transpar"] << std::endl << "// DeformationFieldTransform specific" << std::endl;
-
```
N-Dekker, line 164: Please check that `DeformationFieldTransform::WriteToFile` cannot be entirely removed!

Components/Transforms/WeightedCombinationTransform/elxWeightedCombinationTransform.hxx
```diff
@@ -126,17 +126,6 @@ WeightedCombinationTransformElastix<TElastix>::WriteToFile(const ParametersType
...
-  xout["transpar"] << std::endl << "// WeightedCombinationTransform specific" << std::endl;
-
-  /** Write normalize-weights option */
-  std::string normalizeString = "false";
-  if (this->m_WeightedCombinationTransform->GetNormalizeWeights())
-  {
-    normalizeString = "true";
-  }
-  xout["transpar"] << "(NormalizeCombinationWeights \"" << normalizeString << "\" )" << std::endl;
-
```
N-Dekker, line 139: Please check that `WeightedCombinationTransform::WriteToFile` cannot be entirely removed!

Components/Transforms/BSplineDeformableTransformWithDiffusion/elxBSplineTransformWithDiffusion.hxx
```diff
@@ -1022,54 +1022,6 @@ BSplineTransformWithDiffusion<TElastix>::WriteToFile(const ParametersType & para
     xl::xout["error"] << excp << std::endl;
   }
 
-  /** Get the GridSize, GridIndex, GridSpacing and
```
N-Dekker, line 1025: Please check that `BSplineTransformWithDiffusion::WriteToFile` cannot be entirely removed!

N-Dekker:
Note that there are still two TODO's from related pull request https://github.com/SuperElastix/elastix/pull/379 "Add missing parameters to transform parameter maps"

Components/Transforms/BSplineDeformableTransformWithDiffusion/elxBSplineTransformWithDiffusion.hxx
mstaring: indeed

Components/Transforms/DeformationFieldTransform/elxDeformationFieldTransform.hxx
mstaring: indeed

mstaring:
lgtm

Components/Transforms/DeformationFieldTransform/elxDeformationFieldTransform.hxx
N-Dekker: @mstaring So is it OK to still have some parameters written to file that are not in the parameter map?

N-Dekker: Or would it be an option to add the parameter "DeformationFieldFileName" (which is now only written to file) to the map as well?

mstaring: I guess ideally you would push the whole deformation field to that map. Or at least the filename as you say, so that one can read it in again via disk. Lets discuss over sprint.

N-Dekker: By the way, the "making" of the DeformationFieldFileName throws an exception (`std::out_of_range`) inside `DeformationFieldTransform::WriteToFile` when the current TransformParameterFileName is not specified. Would it be OK to suppress (or avoid) the warning and just leave DeformationFieldFileName as an empty string instead?  https://github.com/SuperElastix/elastix/blob/5.0.1/Components/Transforms/DeformationFieldTransform/elxDeformationFieldTransform.hxx#L180-L182

N-Dekker: We could assume empty string for lastpart, when TransformParameterFileName  is empty.

Components/Transforms/WeightedCombinationTransform/elxWeightedCombinationTransform.hxx
N-Dekker: Update: "SubTransforms" is now added to the `WeightedCombinationTransform` parameter map (pull request https://github.com/SuperElastix/elastix/pull/391 commit ae9c0fd25aa52bd5e83456ad1129b13d1972a9d6) And this pull request is adjusted to simply remove the `WeightedCombinationTransform::WriteToFile` override.  😃

PR #387, opened 2021-01-10
Support TFM + HDF5 file format for initial transform parameter file

N-Dekker:
Added support for the TFM (Insight Transform File, .tfm) and the HDF5 (.h5) file format, for initial transform parameter files, as specified by the parameter `InitialTransformParametersFileName`.

Internally added `TransformIO::Read` and `TransformIO::CreateCorrespondingElxTransform`, which is then called by `TransformBase::ReadInitialTransformFromFile`.

Note: the currect commit only supports transform files that can be read by ITK's `itk::TransformFileReader`. And only those ITK transform types for which a corresponding elastix transform can be created.

### DIFF
--- a/Components/Transforms/ExternalTransform/elxExternalTransform.h
+++ b/Components/Transforms/ExternalTransform/elxExternalTransform.h
@@ -1,0 +1,111 @@
+/*=========================================================================
+ *
+ *  Copyright UMC Utrecht and contributors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef elxExternalTransform_h
+#define elxExternalTransform_h
+
+#include "itkAdvancedCombinationTransform.h"
+#include "elxTransformBase.h"
+
+#include <itkMacro.h>
+#include <itkSmartPointer.h>
+
+namespace elastix
+{
+
+/**
+ * \class ExternalTransform
+ *
+ * This class is represents an external ITK transform.
+ *
+ * The parameters used in this class are:
+ * \parameter Transform: Select this transform as follows:\n
+ *    <tt>(Transform "ExternalTransform")</tt>
+ * \parameter TransformFileName: \n
+ *    example: <tt>(TransformFileName "Transform.h5")</tt> \n
+ *    By default "" is assumed.
+ *
+ * \ingroup Transforms
+ */
+
+template <class TElastix>
+class ITK_TEMPLATE_EXPORT ExternalTransform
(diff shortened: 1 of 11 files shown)


N-Dekker:
Tested as follows:

    transformix.exe -in MovingImage3D.mhd -tp TransformParameters.txt -out out

With TransformParameters.txt as follows:

```
(FixedImageDimension 3)
(FixedInternalImagePixelType "float")
(InitialTransformParametersFileName "TranslationTransform.tfm")
(MovingImageDimension 3)
(MovingInternalImagePixelType "float")
(NumberOfParameters 3)
(Size 5 7 9)
(Transform "TranslationTransform")
(TransformParameters 0 0 0)
(ResultImagePixelType "float")
```

And TranslationTransform.tfm (the _initial transform parameters file_) as follows:

```
#Insight Transform File V1.0
#Transform 0
Transform: TranslationTransform_double_3_3
Parameters: 1 2 3
FixedParameters: 
```

N-Dekker:
Note that in order to specify a TFM or HDF5 file this way, the TransformParameters.txt must still specify its "own" transform as well, as I tested using:

    (Transform "TranslationTransform")
    (TransformParameters 0 0 0)

@stefanklein @mstaring  Would it be an option to allow such a "no-op" transform as default? So that users don't need to specify the transform when they already specify an `InitialTransformParametersFileName`?

mstaring:
That would not make sense. What would make sense is that the tp file itself is in the tfm or hdf5 format. That would be a good next step.

N-Dekker:
@mstaring 
> What would make sense is that the tp file itself is in the tfm or hdf5 format. That would be a good next step.

OK, but a "standard" TFM or HDF5 transform file from ITK does not have all parameters that are required by elastix. How should elastix deal with those "missing parameters"?

N-Dekker:
TODO At today's internal elastix sprint meeting, we decided that TFM/HDF5 ITK transform files should _not_ supported via the  `InitialTransformParametersFileName` parameter! Instead, such an ITK transform file could be specified _instead of_ the `(Transform "TranslationTransform")` parameter in the test example. I'll adjust the proposal accordingly!

Components/Transforms/ExternalTransform/elxExternalTransform.h
```diff
@@ -0,0 +1,111 @@
...
+  GetITKTransform(void) override
+  {
+    std::string transformFileName;
+    this->m_Configuration->ReadParameter(transformFileName, "TransformFileName", 0);
+
+    if (transformFileName != m_TransformFileName)
+    {
+      m_ITKTransform = dynamic_cast<ITKTransformType *>(TransformIO::Read(transformFileName).GetPointer());
+      m_TransformFileName = std::move(transformFileName);
+    }
```
N-Dekker, line 83: During the sprint, it was suggested to move the assignments to `m_ITKTransform` and `m_TransformFileName` to a member function override, like `BeforeAllTransformix(void)` (which should then be declared `virtual` again)

N-Dekker:
@mstaring @stefanklein @ViktorvdValk If I remember correctly, this pull request was rejected. Right? Instead of allowing to specify the HDF5 or TFM file by a `("InitialTransformParametersFileName" ...)` parameter, we added HDF5 + TFM file support by  `("Transform" "File")` and `(TransformFileName ...)`, with pull request #438 commit 1c294607e55e18018874005e57aac458b6f4e013 "ENH: Read data from an external transform file into the parameter map"

Instead of rejecting this pull request, we could also just "postpone", as it _might_ still be useful to someone, to allow specifying InitialTransformParametersFileName by a HDF5 or a TFM. But I believe it's low priority, right?

mstaring:
not sure, let's discuss on the sprint

N-Dekker:
Discussed at sprint, closed.

PR #390, opened 2021-01-13
ENH: DeformationFieldFileName when "TransformParameters." is not found

N-Dekker:
When the current transform parameter filename would not have "TransformParameters." as a substring, the creation of the "lastpart" of a `DeformationFieldFileName` would go wrong (typically causing an `std::out_of_range` exception). However, it still appears useful to make a `DeformationFieldFileName` in that case. Especially when trying to write a default-constructed transform, for test purposes.

This commit makes the "lastpart" empty, in that case. Discussed with Marius @mstaring at the weekly internal elastix sprint.


N-Dekker:
Discussed with Marius (@mstaring) in the context of pull request https://github.com/SuperElastix/elastix/pull/385 "Sync Transform WriteToFile with CreateTransformParametersMap".

PR #397, opened 2021-01-25
STYLE: Make xoutbase abstract and add xoutmain type for xout singleton

N-Dekker:
Made `xoutbase<char>` an abstract base class (as was already documented), by declaring its destructor pure (`= 0`) virtual, and its default-constructor `protected`.

Derived an `xoutmain` class as the new type of the xout singleton.

Aims to clarify the design of the xout library, by distinguishing clearly between base and leaf classes. See also Scott Meyers's book "More Effective C++", Item 33: "Make non-leaf classes abstract".

### DIFF
--- a/Common/xout/xoutbase.h
+++ b/Common/xout/xoutbase.h
@@ -57,11 +57,17 @@ public:
   typedef typename CStreamMapType::value_type   CStreamMapEntryType;
   typedef typename XStreamMapType::value_type   XStreamMapEntryType;
 
-  /** Constructors */
-  xoutbase() = default;
-
   /** Destructor */
-  virtual ~xoutbase() = default;
+  virtual ~xoutbase()
+#if !defined(_MSC_VER) || (_MSC_VER >= 1920)
+    // Preferably declare the destructor pure virtual, to ensure that this is an
+    // abstract base class. Unfortunately, doing so appears to trigger a weird
+    // compilation error with old Visual C++ versions (before Visual Studio 2019),
+    // specifically when having an `#include <valarray>`, saying:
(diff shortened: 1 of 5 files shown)


N-Dekker:
For the record, those VS2017 compile errors when declaring `xoutbase` as an abstract base class can be reproduced by https://rextester.com/WMSL75834 as follows:

```
#include <valarray>
#include <ostream>
#include <iomanip>

template <class T, class = std::basic_ostream<T>>
struct xoutbase
{
  virtual ~xoutbase() = 0;
  void operator<<(std::ostream & (*)(std::ostream&));
};

void fun(xoutbase<char>& x)
{
  x << std::endl;
}
```

Error(s):

> C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\valarray(174): error C2259: 'xoutbase<char,std::basic_ostream<char,std::char_traits<char>>>': cannot instantiate abstract class
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\valarray(174): note: due to following members:
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE\valarray(174): note: 'xoutbase<char,std::basic_ostream<char,std::char_traits<char>>>::~xoutbase(void)': is abstract
source_file.cpp(18): note: see declaration of 'xoutbase<char,std::basic_ostream<char,std::char_traits<char>>>::~xoutbase'
source_file.cpp(24): note: see reference to class template instantiation 'std::valarray<xoutbase<char,std::basic_ostream<char,std::char_traits<char>>>>' being compiled

PR #399, opened 2021-01-27
STYLE: Move static BaseComponent member functions to Conversion class

N-Dekker:


N-Dekker:
@mstaring With this pull request, I moved all the helper functions to convert parameters to string to "elastix/Core/Install/elxConversion.h" (+ ".cxx"). Is that OK to you? Or would you prefer a different directory for those two source files?

mstaring:
this looks like a great reorganization to me

N-Dekker:
> this looks like a great reorganization to me

The pull request itself is simple. In the last few months I had placed some value-to-string utility functions that I needed for the Transform classes in `elx::BaseComponent`. They did not really belong there. It was just because I did not know where else to put them.

Now that there are ten of such conversion helper functions, I think it's clearer to group them together in an `elx::Conversion` class. OK?

PR #400, opened 2021-01-28
STYLE: Remove template arguments charT, traits from xout classes

N-Dekker:
It does not appear necessary to support template arguments (charT and traits) for the xout classes, as elastix always only uses 8-bits `char` as character type of output streams.

Removing those template arguments significantly reduces compilation time (especially when adjusting the implementation of xout) and code complexity.

Discussed at our internal SuperElastix/elastix slack channel.


stefanklein (approved):
Good idea to simplify. I think I copied this style of templating from the standard c++ stream functions, but if we always use char, then let's remove the template argument!

PR #402, opened 2021-01-31
ENH: Add DeformationFieldFileName from two transforms to ParameterMap

N-Dekker:
Adds "DeformationFieldFileName" to the parameter maps of both `DeformationFieldTransform` and `BSplineTransformWithDiffusion`.

Reduced code duplication by adding a helper function, `TransformIO::MakeDeformationFieldFileName`.

Adjusted GTest `CreateTransformParametersMap` unit tests by expecting the added parameter for those two transform types.


N-Dekker:
For the sake of time, I'll just merge this to the develop branch now (unreviewed as it is), as I think it OK and it only has a small impact.

PR #407, opened 2021-02-07
BUG: ReadParameter should fail if parameter string isn't fully processed

N-Dekker:
The `ReadParameter<T>` member functions of `ParameterMapInterface` should fail when a parameter string is not entirely processed, during the conversion from `std::string` to type `T`.

For example, when the parameter has string value "0.1", an attempt to read it into an integer will only process the string until (just before) the decimal point. This should trigger an exception.

Note that a string value like "123.000" should still be converted to an integer, by `ReadParameter<int>`, as it only has trailing zero's after the decimal point.

### DIFF
--- a/Common/GTesting/elxConversionGTest.cxx
+++ b/Common/GTesting/elxConversionGTest.cxx
@@ -140,6 +140,67 @@ Expect_lossless_round_trip_of_signed_integer_parameter_values()
     { minValue, minValue + 1, -1, 0, 1, maxValue - 1, maxValue });
 }
 
+
+template <typename TInteger>
+void
+Expect_parameter_with_decimal_point_and_trailing_zeros_can_be_read_as_integer()
+{
+  using NumericLimits = std::numeric_limits<TInteger>;
+
+  const auto minValue = NumericLimits::min();
+  const auto maxValue = NumericLimits::max();
+
+  for (const TInteger integer : { minValue,
+                                  static_cast<TInteger>(minValue + 1),
+                                  TInteger{},
+                                  TInteger{ 1 },
+                                  static_cast<TInteger>(maxValue - 1),
+                                  maxValue })
+  {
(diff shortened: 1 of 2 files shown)


N-Dekker:
Not really sure which way to go, with this pull request. 

Should we allow a decimal point + trailing zero's for an integer, in order to support `(DefaultPixelValue 0.0)` for integer pixel types? Possibly even scientific notation of whole numbers, e.g., `1e+019`?

Or should we just throw an exception whenever someone tries to do `ReadParameter<int>` on any parameter value string that has a decimal point?

N-Dekker:
So the question is, which of the following strings are acceptable as parameter value (from a parameter map) of an integer type?

    "0.0"
    "42.00"

    "42.1"
    "42.9"  (truncating to 42?)

    "42  43" (ignoring the second integer, 43?)
    "42.nonsense"  (trailing garbage)

    "1.0e+001"
    "1e+019"  (representable as 64-bit unsigned.)

N-Dekker:
Discussed at today's internal elastix sprint meeting. The problem addressed by this pull request seems quite unlikely to happen, and a choice between the possible solutions (disallowing decimal point for integer values, or casting from floating point to integer, or whatever) could not be made.

So for now we decided to postpone this problem (indefinitely).

N-Dekker:
@mstaring @stefanklein Could we possibly reconsider this pull request? It would actually allow early detection of problems like the one reported by Aurelie, "elastix crash when LANG=fr_FR.UTF-8" at https://groups.google.com/g/elastix-imageregistration/c/70PXd-Hy_Xo

Aurelie reported that parameter value "0.2", was truncated to "0", and it seems to me that the parameter string just wasn't fully processed. In this case, I would prefer to have an error message during the parsing of the parameter values. OK?

N-Dekker:
Discussed again at today's sprint. We agreed to have it merged  😃

PR #409, opened 2021-02-09
Replace xout["iteration"] by access to IterationInfo via ElastixBase and BaseComponentSE 

N-Dekker:
No longer added `ElastixBase::m_IterationInfo` to `xout["iteration"]`. Accessed `IterationInfo` via newly added member functions of `ElastixBase` and `BaseComponentSE`, instead. 

Aims to improve thread-safety, as discussed at our internal weekly elastix sprint.

Related to issue #174 "elastix 4.9 static library version not thread safe"


mstaring:
does this have performance implications?

N-Dekker:
> does this have performance implications?

@mstaring  It might be slightly faster, because it avoids lookup into the `xout` data structure. But I don't expect a large performance improvement, just from this pull request. What do you think of?

> Otherwise lgtm

Thanks!

mstaring:
great, go ahead then!

PR #410, opened 2021-02-09
STYLE: Remove xout["WriteHeaders"] logic, simply call WriteHeaders()

N-Dekker:
Dropped "WriteHeaders" as special case for `xout[name]`.

Replaced "WriteHeaders" by a direct call to `xoutrow::WriteHeaders()`.

Aims to just simplify the code.


mstaring:
lgtm

Issue #411, opened 2021-02-09
elastix::TransformBase itk::Object casting


thewtex:
Building on the work of @ViktorvdValk to address the macOS ITKElastix Python package, using

https://github.com/ViktorvdValk/elastix/tree/ITK_TEMPLATE_EXPORT

https://github.com/ViktorvdValk/ITKElastix/tree/UseForkITK

with the script:

```
import itk
import faulthandler

faulthandler.enable()
fixed_image = itk.imread('data/CT_2D_head_fixed.mha', itk.F)
moving_image = itk.imread('data/CT_2D_head_moving.mha', itk.F)

parameter_object = itk.ParameterObject.New()
default_rigid_parameter_map = parameter_object.GetDefaultParameterMap('rigid')
parameter_object.AddParameterMap(default_rigid_parameter_map)

result_image, result_transform_parameters = itk.elastix_registration_method(
    fixed_image, moving_image,
    parameter_object=parameter_object,
    log_to_console=False)
```

a debug build with lldb, identifies the issue:

```
Process 40949 stopped
* thread #2, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00000001616575f1 _ElastixPython.so`elastix::TransformBase<elastix::ElastixTemplate<itk::Image<float, 2u>, itk::Image<float, 2u> > >::SetTransformParametersFileName(this=0x000000016ccacfd0, filename="output_path_not_set/TransformParameters.0.txt") at elxTransformBase.hxx:1694:17
   1691	    this->m_TransformParametersFileName = "";
   1692	  }
   1693	  ObjectType * thisAsObject = dynamic_cast<ObjectType *>(this);
-> 1694	  thisAsObject->Modified();
   1695	
   1696	} // end SetTransformParametersFileName()
   1697	
Target 0: (Python) stopped.
(lldb) p this
(elastix::TransformBase<elastix::ElastixTemplate<itk::Image<float, 2>, itk::Image<float, 2> > > *) $0 = 0x000000016ccacfd0
(lldb) p thisAsObject
(ObjectType *) $1 = 0x0000000000000000
```

https://github.com/ViktorvdValk/elastix/blob/8ef128fedc26023df0799ff2d4fe965c184b2151/Core/ComponentBaseClasses/elxTransformBase.hxx#L1676-L1696

[...]

PR #412, opened 2021-02-10
'fixed_template_export_macro_in_classes'

thewtex:


mstaring:
these commits make sense to me. We just became sloppy with the template exports, and it is good to reuse the definitions from itkMacro. 
Just one question: the commit on replacing the dynamic cast with static ones, why is that?

mstaring:
@N-Dekker @ViktorvdValk

Common/xout/xoutbase.h
```diff
@@ -44,7 +44,7 @@ using namespace std;
  */
 
 template <class charT, class traits = char_traits<charT>>
-class ITK_TEMPLATE_EXPORT xoutbase
+class xoutbase
```
N-Dekker: FYI Changes to the declaration of `xout` classes are likely to cause git conflicts now when trying to merge to `develop`, as I recently made them non-templated. With commit https://github.com/SuperElastix/elastix/commit/d0af9767512f7625dffc520322d421a9392c2527 (28 January 2021)

N-Dekker:
> Just one question: the commit on replacing the dynamic cast with static ones, why is that?

`dynamic_cast` uses runtime type information (RTTI) to check that the cast can be performed properly. `static_cast` does a more "brute force" cast, ignoring RTTI. It appears that the use of RTTI might be troublesome "over library boundaries", at least for Clang.

Having said so, I do not know if it is correct to just ignore runtime type information, in these cases.

Core/ComponentBaseClasses/elxTransformBase.hxx
```diff
@@ -1709,7 +1712,10 @@ TransformBase<TElastix>::SetReadWriteTransformParameters(const bool _arg)
   {
     this->m_ReadWriteTransformParameters = _arg;
     ObjectType * thisAsObject = dynamic_cast<ObjectType *>(this);
-    thisAsObject->Modified();
+    if (thisAsObject != nullptr)
```
N-Dekker, line 1715: @thewtex Thanks you, Matt, but have you actually seen `dynamic_cast<ObjectType *>(this)` returning a `nullptr`? Because I believe the cast should _always_ succeed.

mstaring:
@N-Dekker Would you have time to take this branch, solve merge conflicts and make it into a ready PR?

ViktorvdValk:
> these commits make sense to me. We just became sloppy with the template exports, and it is good to reuse the definitions from itkMacro.
> Just one question: the commit on replacing the dynamic cast with static ones, why is that?

This was one of my tries to prevent the segfault, but I think since Matt used my fork on which I tried everything, this.commit is still in there. But since it didn't help (also didn't make things worse btw), I would say we can ignore the dynamic -> static commit.

mstaring:
ok, so then we make a clean pr with the template exports and the use of itkMacro, right?

N-Dekker:
The current elastix `develop` branch builds fine, also on mac, even though it has an example of an executable which links to the elastix library (elastix/Core/Main/GTesting test executable ElastixLibGTest). Would it be possible to have a test in SuperElastix/elastix which reproduces the clang visibility issue?

Core/ComponentBaseClasses/elxTransformBase.hxx
mstaring: let's first make a pr without this change. If Matt actually saw a problem here it can be a separate commit, as it addresses a separate problem also.

thewtex: Yes, I was observing `nullptr` / `0x0` with lldb as described in #411. Feel free to split up this branch, rebase, etc.

ViktorvdValk:
Yes! I thinks the mac issues is fixed with this. 
I build a package using this elastix branch and I don't get the segfault.
The debugger still gives the visibility error though, but elastix does run normally.

ViktorvdValk (approved):
Yes! I thinks the mac issues is fixed with this.
I build a package using this elastix branch and I don't get the segfault.
The debugger still gives the visibility error though, but elastix does run normally.

Core/ComponentBaseClasses/elxTransformBase.hxx
N-Dekker: @thewtex  Ah, I see now! You got a nullptr! 🙀 It is really strange because unchecked use of this `dynamic_cast<ObjectType *>(this)` has been in there for a very long time! 14 October 2004:

https://github.com/SuperElastix/elastix/blob/396792f05becfe1dbe809bbab37d7758f1b5a835/src/Core/ComponentBaseClasses/elxTransformBase.hxx#L809-L810

I suspect that it has to do with the whole clang visibility/RTTI issue. So I hope that the `if (thisAsObject != nullptr)` is not necessary anymore once the visibility is fixed.

N-Dekker:
With the force-push I just did, I squashed all commits from @ViktorvdValk that added the `ITK_TEMPLATE_EXPORT ` keywords, and dropped the other commits (which replaced some `dynamic_cast`s by `static_cast`s, and added a check to `dynamic_cast<ObjectType *>(this)`). My motivation is that `dynamic_cast` should just work properly. Otherwise there must be something wrong with the visibility of class types, which should have been fixed by those `ITK_TEMPLATE_EXPORT ` keywords!

mstaring:
should we also remove our redefinition of these macro's and include itkMAcro?

Core/Install/elxMacro.h
```diff
@@ -27,6 +27,7 @@
  * includes itkWin32Header.h.
  */
 #include "itkWin32Header.h"
+#include "itkMacro.h"
```
N-Dekker, line 30: @mstaring 

> should we also remove our redefinition of these macro's and include itkMAcro?

As far as I can see, this PR does not have a redefinition of `ITK_TEMPLATE_EXPORT` (anymore?), and neither does the current develop branch.

This PR does include itkMacro.

Core/ComponentBaseClasses/elxTransformBase.hxx
thewtex: > So I hope that the if (thisAsObject != nullptr) is not necessary anymore once the visibility is fixed.

Yes. There are many other classes in the elastix codebase that need the `ITK_TEMPLATE_EXPORT` macro. Until the visibility source is unearthed, we may want to patch the version used in ITKElastix to avoid crashing on macOS.

thewtex (commented):
Thanks for the updates, @N-Dekker

N-Dekker:
I'll rebase-and-merge now, as it has passed the SuperElastix/elastix CI, but I'm interested to hear if it has entirely fixed the clang visibility problem...!

PR #415, opened 2021-02-11
CI: Elastix docker image build and push to docker hub

ViktorvdValk:
The CI now builds a docker image of every tagged elastix commit. The docker image is uploaded to [dockerhub](https://hub.docker.com/repository/docker/superelastix/elastix) and labeled by it's tag.

### DIFF
--- a/.github/workflows/ElastixGitHubActions.yml
+++ b/.github/workflows/ElastixGitHubActions.yml
@@ -46,6 +46,7 @@ jobs:
         mv ${items[*]} Elastix-source
         mv .editorconfig Elastix-source
         mv .clang-format Elastix-source
+        mv Elastix-source/Dockerfile .
       shell: bash
       
     - uses: actions/cache@v2
@@ -54,8 +55,8 @@ jobs:
         path: |
           ITK-build
           ITK-source
-        key: ${{ matrix.itk-git-tag }}-${{ matrix.os }}-${{ matrix.cmake-build-type }}
-      
+        key: ${{ matrix.itk-git-tag }}-${{ matrix.os }}-${{ matrix.cmake-build-type }} 
+        
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
       with:
@@ -93,7 +94,6 @@ jobs:
     
     - name: Build Elastix Unix
       if: matrix.os != 'windows-2019'
-
       run: |
         mkdir Elastix-build
         cd Elastix-build
@@ -154,5 +154,20 @@ jobs:
       with: 
         name: "${{ matrix.os }}"
         path: uploads
+        
+    - name: Login to Docker Hub
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && matrix.os == 'ubuntu-18.04'
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_HUB_USERNAME }}
+        password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+    - name: Build and push
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && matrix.os == 'ubuntu-18.04'     
+      run: |
+        mkdir output
+        docker build -t superelastix/elastix:${GITHUB_REF#refs/tags/} .
+        docker run -u $UID:$GROUPS --mount type=bind,source="$(pwd)"/output,target=/out -v "$(pwd)"/Elastix-source/Testing/Data:/elastix/ superelastix/elastix:${GITHUB_REF#refs/tags/} elastix -out /out/ -p /elastix/parameters.3D.NC.euler.ASGD.001.txt -f /elastix/3DCT_lung_baseline.mha -m /elastix/3DCT_lung_followup.mha
+        docker push superelastix/elastix:${GITHUB_REF#refs/tags/}
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM ubuntu:20.04
+
+# Add labels to image
+LABEL documentation="https://github.com/SuperElastix/elastix/wiki"
+LABEL license="Apache License Version 2.0"
+LABEL modelzoo="https://elastix.lumc.nl/modelzoo/"
+
+# Prepare system packages, libgomp1 is required by elastix
+RUN apt-get update && apt-get -qq install libgomp1 -y
+
+# Copy necessary files
+COPY uploads/bin/* /usr/local/bin/
+COPY uploads/lib/* /usr/lib/
+
+COPY uploads/LICENSE /
+COPY uploads/NOTICE /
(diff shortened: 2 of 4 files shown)


mstaring:
Looks good! Perhaps Sebastian can review. This only triggers on the master branch?

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -154,5 +154,20 @@ jobs:
...
+      with:
+        username: ${{ secrets.DOCKER_HUB_USERNAME }}
+        password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+    - name: Build and push
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && matrix.os == 'ubuntu-18.04'     
+      run: |
+        mkdir output
+        docker build -t superelastix/elastix:${GITHUB_REF:10} .
+        docker run -u $UID:$GROUPS --mount type=bind,source="$(pwd)"/output,target=/out -v "$(pwd)"/Elastix-source/Testing/Data:/elastix/ superelastix/elastix:${GITHUB_REF:10} elastix -out /out/ -p /elastix/parameters.3D.NC.euler.ASGD.001.txt -f /elastix/3DCT_lung_baseline.mha -m /elastix/3DCT_lung_followup.mha
```
N-Dekker, line 170: Just for my understanding, it runs a registration here, just to test if it works?

stefanklein: it seems so to me as well, and then my question would be: is there a check somewhere that this run was successful?

stefanklein (approved):
Maybe be good if Sebastian has a look as well, but it looks fine to me!

.github/workflows/ElastixGitHubActions.yml
ViktorvdValk: Yes, it runs registration to test the docker image.
I assume here that if it runs and doesn't give an error, the docker image is build successfully.
Since the elastix binary is already extensively tested in this workflow.
Or do we want more test to verify the docker image as well?

ViktorvdValk:
This only triggers on tagged branches, so in the case of elastix, that will indeed only be the master branch.
@Svdvoort this is my implementation of the docker image build and push. I decided to make 1 image for both elastix and transformix. Would you mind giving feedback on the implementation?

.github/workflows/ElastixGitHubActions.yml
stefanklein: I don't think more extensive testing is needed, it's enough indeed to see if it runs and doesn't give an error, but I didn't see an explicit check here whether this was the case (for example checking for the return value of 'docker run'). if the script automatically stops when 'docker run' gives an error then this also might be fine, but I don't know how this works with this yml files.

ViktorvdValk: The script indeed stops when 'docker run' gives an error. It's a bit hard to see this on this PR though, since the script includes a check to see if the PR/commit is tagged before it runs the docker part of the workflow. Since this PR is not tagged, the CI will not run the docker part.
The CI on [this fork](https://github.com/ViktorvdValk/elastix/runs/1878956471?check_suite_focus=true) shows the workflow run as it will be when a commit is tagged.

stefanklein: then it's fine I think!

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:20.04
+
+# Prepare system packages
+RUN apt-get update
+
+# wget is need to get Elastix package
+# libgomp1 is required by elastix
+RUN apt-get -qq install wget libgomp1 -y
```
Svdvoort, line 8: I think you can get rid of the `wget` install. It was required in my docker to get the elastix packages from github, but since you are directly adding the binaries/libs below I do not think it is needed anymore.

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -154,5 +154,20 @@ jobs:
...
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_HUB_USERNAME }}
+        password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+    - name: Build and push
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && matrix.os == 'ubuntu-18.04'     
+      run: |
+        mkdir output
+        docker build -t superelastix/elastix:${GITHUB_REF:10} .
```
Svdvoort, line 169: This tags with the commit hash, but doesn't automatically build with a new release. Or will release builds be build/pushed manually? I think those are also (perhaps more) interesting for most users.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:20.04
```
Svdvoort, line 1: You can consider adding labels to the image, for example about the authors/license etc. For more info see here: https://github.com/opencontainers/image-spec/blob/master/annotations.md

For example, to point users to the documentation: 

LABEL org.opencontainers.image.documentation="https://github.com/SuperElastix/elastix/wiki"

Svdvoort:
Looks good overall! Some very minor suggestions, but everything should work as-is. Having an 'on release' action to tag the docker images automatically when a release takes place would be nice, I think most users will only use the stable tagged releases? Of course, you can also build these manually if you want to test these images more thoroughly before pushing a stable release.

.github/workflows/ElastixGitHubActions.yml
ViktorvdValk: This ensures that a docker image will only be build and pushed to dockerhub on a tagged commit/release. This way dockerhub only contains working and relevant elastix docker images.
I should push release build automatically right?

Svdvoort: You're right, I missed the `startsWith` condition. Personally I prefer to use `${GITHUB_REF#refs/tags/}` instead of `${GITHUB_REF:10}`, because I think it's a bit clearer as to what it's doing, but that's personal preference.

Svdvoort: You can also create a seperate 'release' CI, with: 
```
on:
  push:
    tags:
    - '*.*.*'
```
to not clutter your 'nightly' CI.

ViktorvdValk:
Thanks a lot for the feedback @Svdvoort. An update is on it's way.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+# Prepare system packages
+RUN apt-get update
```
Svdvoort, line 4: Forgot to mention this earlier: it's better to run apt-get update and install on a single line (e.g. `RUN apt-get update && apt-get install libgomp1 -y`) due to caching issues that can arise otherwise. See here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

PR #417, opened 2021-02-15
COMP: Fix GetJacobianOfSpatialHessian warnings on jsh_tmp buffer overrun

N-Dekker:
Solved fifteen Visual C++ warnings from `RecursiveBSplineTransformImplementation::GetJacobianOfSpatialHessian`, which said:

> warning C4789: buffer 'jsh_tmp' of size 32 bytes will be overrun; 8 bytes will be written starting at offset 64

Note that the buffer overrun would not occur in practice, as the size of `jsh_tmp` depends on `OutputDimension`, which was checked at runtime. But the compiler still issued those warnings. Which are fixed by this commit.

### DIFF
--- a/Common/Transforms/itkRecursiveBSplineTransformImplementation.h
+++ b/Common/Transforms/itkRecursiveBSplineTransformImplementation.h
@@ -20,6 +20,10 @@
 
 #include "itkRecursiveBSplineInterpolationWeightFunction.h"
 
+// Standard C++ header files:
+#include <cassert>
+#include <cstring> // For memcpy.
+
 namespace itk
 {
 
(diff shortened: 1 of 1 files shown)


Common/Transforms/itkRecursiveBSplineTransformImplementation.h
```diff
@@ -533,22 +537,13 @@ class RecursiveBSplineTransformImplementation<OutputDimension, 0, SplineOrder, T
...
-      jsh_tmp[0] = jsh[9];
-      jsh_tmp[1] = jsh[8];
-      jsh_tmp[2] = jsh[7];
-      jsh_tmp[3] = jsh_tmp[1];
-      jsh_tmp[4] = jsh[5];
-      jsh_tmp[5] = jsh[4];
-      jsh_tmp[6] = jsh_tmp[2];
-      jsh_tmp[7] = jsh_tmp[5];
-      jsh_tmp[8] = jsh[2];
+      const double tmp[] = { jsh[9], jsh[8], jsh[7], jsh[8], jsh[5], jsh[4], jsh[7], jsh[4], jsh[2] };
```
mstaring, line 540: are these numbers correct? in the old code it reads 9 8 7 then 1 ...

N-Dekker: @mstaring Thanks for asking, Marius! Honestly I did double check  😃 The old code did `jsh_tmp[3] = jsh_tmp[1]`, knowing that  jsh_tmp[1] == jsh[8]. So the old code did 9 8 7 then _indirectly_ 8, via `jsh_tmp[1]`!

So the new code _might_ even be faster than the old one, because it does 9 8 7 8 _directly_!  (I say _might_ because I did not do any benchmarking.)

mstaring: Ah yes, i see now. Thank you!

PR #418, opened 2021-02-15
ENH: Add 4D float support

ViktorvdValk:


mstaring:
More something for @stefanklein to comment on. 

I am not sure we should change this default value. It is also possible to set this cmake flag externally when building elastix. Just override this cmake option for ITKELastix.

stefanklein (approved):
To me it seems a good idea to enable it by default, if that's what ITKElastix requires as well, in order to keep things simple (instead of having different defaults for command-line and itkelastix). The only downside of adding "float" here is that compilation time increases. The advantage is that users who have sufficient memory can use the floating point precision without having to compile elastix themselves.

mstaring:
agreed then!

PR #419, opened 2021-02-15
BUG: TransformBase call Modified when itk::Object dynamic_cast succeeds

thewtex:
This fails in the wrapping with the type of `this`:

  `elastix::TransfromBase<elastix::ElastixTemplate<itk::Image<float,
  2u>, itk::Image<float, 2u> > >`

This is currently necessary on macOS to prevent segfaults in ITKElastix
Python wrapping.

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.hxx
+++ b/Core/ComponentBaseClasses/elxTransformBase.hxx
@@ -1415,7 +1415,10 @@ TransformBase<TElastix>::SetTransformParametersFileName(const char * filename)
     this->m_TransformParametersFileName = "";
   }
   ObjectType * thisAsObject = dynamic_cast<ObjectType *>(this);
-  thisAsObject->Modified();
+  if (thisAsObject != nullptr)
+  {
+    thisAsObject->Modified();
+  }
 
 } // end SetTransformParametersFileName()
 
@@ -1433,7 +1436,10 @@ TransformBase<TElastix>::SetReadWriteTransformParameters(const bool _arg)
   {
     this->m_ReadWriteTransformParameters = _arg;
     ObjectType * thisAsObject = dynamic_cast<ObjectType *>(this);
-    thisAsObject->Modified();
+    if (thisAsObject != nullptr)
+    {
+      thisAsObject->Modified();
+    }
   }
 
 } // end SetReadWriteTransformParameters()


N-Dekker:
Thanks for trying it again, Matt! Again, I think this `dynamic_cast<ObjectType *>(this)` should never fail at all, so it seems to me that this pull request would just fix the symptom, rather than the underlying problem. But anyway, is this one change sufficient to avoid the Mac segfault? 

I wonder if it is OK for `TransformBase::SetTransformParametersFileName` not to call `thisAsObject->Modified()` when it is modified. @stefanklein @mstaring ?  

Note: It appears that the elastix source tree still has more than 180 other `dynamic_cast`s. While I'm not a fan of type casting anyway, I wonder, should all of these `dynamic_cast`s now be reconsidered? 🤔

thewtex:
> I think this dynamic_cast<ObjectType *>(this) should never fail at all, so it seems to me that this pull request would just fix the symptom, rather than the underlying problem. 

Yes, this is probably true, since it does not fail on the other platforms. However, it is difficult to nail down what classes are actually involved. There are many classes in elastix that need export improvements. This does prevent folks on macOS not experience those jolting crashes :zap: . I think they would prefer have this symptom addressed while the underlying problem is fixed.

stefanklein:
I thought calling the thisAsObject-> Modified() is necessary for any ITK object that is modified.
And avoiding that call seems not the right thing either, when you actually say that this dynamic_cast SHOULD work.
Cheers, stefan

thewtex:
> I thought calling the thisAsObject-> Modified() is necessary for any ITK object that is modified.

If there is a pipeline update that will need to follow according to the `ModifiedTime`, then yes. Is that the situation in this case? 

@stefanklein Do you know what class is failing the `dynamic_cast`?

N-Dekker:
Now I remember I had once figured out how to avoid this `dynamic_cast<ObjectType *>(this)`, and still be able to access the `itk::Object` interface of an elastix TransformBase object. Please wait...  😃

N-Dekker:
@thewtex Would pull request https://github.com/SuperElastix/elastix/pull/420 fix the macOS segfaults? You see, it replaces 11 dynamic_cast's by equivalent virtual function calls. So the question is, if RTTI does not work properly for macOS, will virtual function calls still be fine.

thewtex:
@N-Dekker great idea :bulb: Testing ...

thewtex:
Superceded by #420

PR #420, opened 2021-02-15
BUG: Replace dynamic_cast's TransformBase by GetAsCombinationTransform()

N-Dekker:
Replaced 11 `dynamic_cast`'s from the implementation of `elx::TransformBase` by calls to `this->GetAsCombinationTransform()`, which is now pure `virtual`. Did override `TransformBase::GetAsCombinationTransform()` in all of its 17 derived transform classes.

Aims to fix MacOS segfaults, mentioned by Matt McCormick (@thewtex), issue https://github.com/SuperElastix/elastix/issues/411, "elastix::TransformBase itk::Object casting".

### DIFF
--- a/Components/Transforms/AdvancedAffineTransform/elxAdvancedAffineTransform.h
+++ b/Components/Transforms/AdvancedAffineTransform/elxAdvancedAffineTransform.h
@@ -214,6 +214,18 @@ protected:
   ReadCenterOfRotationPoint(InputPointType & rotationPoint) const;
 
 private:
+  const Self &
+  GetAsCombinationTransform(void) const override
+  {
+    return *this;
+  }
+
+  Self &
+  GetAsCombinationTransform(void) override
+  {
+    return *this;
+  }
+
   /** Creates a map of the parameters specific for this (derived) transform type. */
   ParameterMapType
   CreateDerivedTransformParametersMap(void) const override;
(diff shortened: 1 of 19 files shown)


thewtex (dismissed):
Verified to address the macOS symbol issue! :1st_place_medal:

mstaring:
nice work! lgtm

N-Dekker:
@mstaring @ViktorvdValk @stefanklein This pull request should fix the macOS segfaults! As confirmed by Matt (@thewtex) 😄 

Honestly I find it a bit cumbersome to have added so many "boilerplate" `GetAsCombinationTransform()` overrides, but apparently, a pure virtual `TransformBase::GetAsCombinationTransform()` member function call works fine on MacOS, while the corresponding `dynamic_cast`would fail!

Please check if it's OK to merge this PR!

ViktorvdValk:
This is great!
I think Matt already confirmed that MacOS doesn't give the segfault with this PR, but I'm compiling right now, to see check the debugger errors/warnings

N-Dekker:
Other elastix component may need to have a similar pure `virtual` GetAs.... member function, to avoid future MacOS crashes. 🤔

stefanklein (approved):
nice work in fixing this.

ViktorvdValk:
Running ITKElastix on MacOS doesn't segfault with this PR, so the issue is indeed fixed.
The lldb debugger does give the exacts same warning/errors though, which is strange since you reduced the number of dynamic casts.
And just to reassure; even though the lldb debugger gives "errors", the code executes as expected.

PR #423, opened 2021-02-18
Replace dynamic_cast from all GetAsITKBaseType() member functions by this->GetSelf()

N-Dekker:
Added pure virtual `GetSelf()` member functions to each of the ten component base classes, returning a reference to their `ITKBaseType`. Implemented by the added macro's, `elxDeclarePureVirtualGetSelfMacro(type)` and `elxOverrideGetSelfMacro`.

Replaced `dynamic_cast`s from all `GetAsITKBaseType()` member functions by `this->GetSelf()` calls.

The main aim of this pull request is to avoid MacOS segfaults like the ones mentioned by Matt McCormick (@thewtex), issue #411, "elastix::TransformBase itk::Object casting".

This pull request supersedes pull request https://github.com/SuperElastix/elastix/pull/420 commit 6fcd90348664c4a477a73552194592de014c02a5 ("BUG: Replace dynamic_cast's TransformBase by GetAsCombinationTransform()"), which only addressed the issue for `TransformBase` (whereas this new pull request addresses the issue for all of the elastix component base classes).

### DIFF
--- a/Components/FixedImagePyramids/FixedGenericPyramid/elxFixedGenericPyramid.h
+++ b/Components/FixedImagePyramids/FixedGenericPyramid/elxFixedGenericPyramid.h
@@ -131,6 +131,8 @@ protected:
   ~FixedGenericPyramid() override = default;
 
 private:
+  elxOverrideGetSelfMacro;
+
   /** The deleted copy constructor. */
   FixedGenericPyramid(const Self &) = delete;
   /** The deleted assignment operator. */
--- a/Components/FixedImagePyramids/FixedRecursivePyramid/elxFixedRecursivePyramid.h
+++ b/Components/FixedImagePyramids/FixedRecursivePyramid/elxFixedRecursivePyramid.h
@@ -90,6 +90,8 @@ protected:
   ~FixedRecursivePyramid() override = default;
 
 private:
+  elxOverrideGetSelfMacro;
+
   /** The deleted copy constructor. */
   FixedRecursivePyramid(const Self &) = delete;
   /** The deleted assignment operator. */
--- a/Components/FixedImagePyramids/FixedShrinkingPyramid/elxFixedShrinkingPyramid.h
+++ b/Components/FixedImagePyramids/FixedShrinkingPyramid/elxFixedShrinkingPyramid.h
@@ -90,6 +90,8 @@ protected:
   ~FixedShrinkingPyramid() override = default;
 
 private:
+  elxOverrideGetSelfMacro;
+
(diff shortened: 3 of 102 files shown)


thewtex (approved):
:+1:

N-Dekker:
@thewtex Thanks for your approval, Matt. 😃  I still need to further discuss this pull request with @mstaring, @stefanklein and @ViktorvdValk as it involves so many elastix source files. Technically we wonder, it is really necessary to remove all those `dynamic_cast`s from the elastix component base classes, in order to avoid MacOS errors. Or should it be sufficient to just make sure that all library visibility and class exports are OK?

Personally I'm still OK with this pull request, although it has some "macro magic", and some more "boilerplate code".

ViktorvdValk (approved):
Closes [the macOS segfault issue](https://github.com/InsightSoftwareConsortium/ITKElastix/issues/52).
However the lldb debugger keeps throwing "soft" dynamic cast errors/warning when doing a registration with the itk-elastix python package built with this update of elastix.
A part of these errors/warning are also thrown by lldb with a script that's just calling itk.imread, so these errors/warnings are (at least partially) a result of code in itk. @thewtex @dzenanz. 

See [itk-elastix issue ](https://github.com/InsightSoftwareConsortium/ITKElastix/issues/105)
 

```

thewtex:
I think it is a good idea to use this approach, which is more robust, but also takes steps to fix the library visibility and class exports.

thewtex:
> A part of these errors/warning are also thrown by lldb with a script that's just calling itk.imread, so these errors/warnings are (at least partially) a result of code in itk

I do not think they are the result of code in ITK. The warnings are a bit flaky (inconsistent), and they will usually propose a list of 3 classes where there could be an issue. One of the proposed classes is an ITK class, but checking those classes, the export is set.

mstaring:
Should we first prepare a patch to everywhere use ITK_TEMPLATE_EXPORT?

N-Dekker:
(Marked this pull request as "draft", because it is still under discussion. Technically I think it is ready.)

ViktorvdValk:
@thewtex But why would itk.imread() then also throw these errors?

N-Dekker:
Discussed at today's weekly internal elastix sprint. We agreed to have it merged! 😄

thewtex:
> @thewtex But why would itk.imread() then also throw these errors?

The classes with warnings flagged when running `itk.imread` have their export settings set, and after resolving all the identified classes in ITK there were no longer any issues. There are many classes in elastix that do not have the export specification set.

PR #424, opened 2021-02-19
opencl

thewtex:
- COMP: Define CL_TARGET_OPENCL_VERSION
- COMP: Update itkGPUKernelClassMacro call

### DIFF
--- a/Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
+++ b/Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
@@ -25,7 +25,7 @@
 namespace itk
 {
 /** Create a helper GPU Kernel class for GPUBSplineTransform */
-itkGPUKernelClassMacro(GPUBSplineTransformKernel);
+class itkGPUKernelClassMacro(GPUBSplineTransformKernel);
 
 /** \class GPUBSplineBaseTransform
  * \brief GPU base class for the BSplineTransform.
--- a/Common/OpenCL/ITKimprovements/itkOpenCL.h
+++ b/Common/OpenCL/ITKimprovements/itkOpenCL.h
@@ -20,6 +20,11 @@
 
 #include "itkOpenCLExport.h"
 
+// For the Official Khronos OpenCL headers
+#ifndef CL_TARGET_OPENCL_VERSION
+#  define CL_TARGET_OPENCL_VERSION 120
+#endif
+
 #if defined(__APPLE__) || defined(__MACOSX)
 #  include <OpenCL/cl_platform.h>
 #  include <OpenCL/cl.h>
--- a/Common/OpenCL/Filters/itkGPUBSplineDecompositionImageFilter.h
+++ b/Common/OpenCL/Filters/itkGPUBSplineDecompositionImageFilter.h
@@ -24,7 +24,7 @@
 namespace itk
 {
 /** Create a helper GPU Kernel class for GPUBSplineDecompositionImageFilter */
-itkGPUKernelClassMacro(GPUBSplineDecompositionImageFilterKernel);
+class itkGPUKernelClassMacro(GPUBSplineDecompositionImageFilterKernel);
 
 /** \class GPUBSplineDecompositionImageFilter
  * \brief GPU version of BSplineDecompositionImageFilter.
--- a/Common/OpenCL/Filters/itkGPUBSplineInterpolateImageFunction.h
+++ b/Common/OpenCL/Filters/itkGPUBSplineInterpolateImageFunction.h
@@ -28,7 +28,7 @@
 namespace itk
 {
 /** Create a helper GPU Kernel class for GPUBSplineInterpolateImageFunction */
-itkGPUKernelClassMacro(GPUBSplineInterpolateImageFunctionKernel);
+class itkGPUKernelClassMacro(GPUBSplineInterpolateImageFunctionKernel);
 
 /** \class GPUBSplineInterpolateImageFunction
  * \brief GPU version of BSplineInterpolateImageFunction.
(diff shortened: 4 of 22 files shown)


Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
```diff
@@ -25,7 +25,7 @@
 namespace itk
 {
 /** Create a helper GPU Kernel class for GPUBSplineTransform */
-itkGPUKernelClassMacro(GPUBSplineTransformKernel);
+class itkGPUKernelClassMacro(GPUBSplineTransformKernel);
```
N-Dekker, line 28: Thanks Matt. Do I understand correctly that your proposed change would break the compilation of elastix with ITK 5.1.1 5.1.2? If so, maybe your PR should be postponed until the `itkGPUKernelClassMacro` adjustment is part of a final ITK release (either ITK 5.1.3 or 5.2.0). What do you think, Marius @mstaring?

Common/OpenCL/ITKimprovements/itkOpenCL.h
```diff
@@ -20,6 +20,11 @@
 
 #include "itkOpenCLExport.h"
 
+// For the Official Khronos OpenCL headers
+#ifndef CL_TARGET_OPENCL_VERSION
+#  define CL_TARGET_OPENCL_VERSION 120
+#endif
+
```
N-Dekker, line 27: For the record, we had already some discussion about `CL_TARGET_OPENCL_VERSION` at pull request https://github.com/SuperElastix/elastix/pull/213 "Khronos OpenCL support", December 2019

Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
thewtex: > Do I understand correctly that your proposed change would break the compilation of elastix with ITK 5.1.1 5.1.2

Yes.

>  If so, maybe your PR should be postponed until the itkGPUKernelClassMacro adjustment is part of a final ITK release (either ITK 5.1.3 or 5.2.0).

It is currently part of the ITK 5.2 Release Candidates.

Common/OpenCL/ITKimprovements/itkOpenCL.h
thewtex: Yes, that PR has not been merged, but the issue still exists.

Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
N-Dekker: > It is currently part of the ITK 5.2 Release Candidates.

Thanks @thewtex. Do I understand correctly that the definition of `itkGPUKernelClassMacro` was changed for one specific case, `GPUDenseFiniteDifferenceImageFilterKernel`, which needed to have an export specifier next to the `class` keyword?

Or should all those GPU kernel classes eventually have such an export specifier?

thewtex: @N-Dekker if you are referring to:

https://github.com/InsightSoftwareConsortium/ITK/issues/1381

https://github.com/InsightSoftwareConsortium/ITK/pull/1814

The export specifier was required for `GPUDenseFiniteDifferenceImageFilter` because of the exported `Enum` class:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/GPUFiniteDifference/src/itkGPUFiniteDifferenceFilterEnum.cxx

The other classes could have export specifications added if the associated module is built as a shared library.

N-Dekker: Thanks for your explanation, @thewtex I'm afraid I still don't understand, because `itkGPUKernelClassMacro` does not introduce any Enum class, as far as I can see. But anyway, I just tried to replace the `itkGPUKernelClassMacro` from elastix in a way that should be compatible with both ITK 5.1 and ITK 5.2 (RC1, RC2). Please check: https://github.com/SuperElastix/elastix/pull/427

Common/OpenCL/ITKimprovements/itkOpenCL.h
N-Dekker: Thanks Matt. So now I still wonder how your proposed `CL_TARGET_OPENCL_VERSION` relates to the `OPENCL_C_VERSION_1_1`, `CL_TARGET_OPENCL_VERSION_1_2`, and `OPENCL_C_VERSION_2_0` CMake variables from elastix. Should the definition of `CL_TARGET_OPENCL_VERSION` correspond to the enabled `OPENCL_C_VERSION` CMake variable(s)? Or are they entirely independent?

Is `120` just a default, or is it the only proper value for `CL_TARGET_OPENCL_VERSION` right now?

Common/OpenCL/Filters/itkGPUBSplineBaseTransform.h
thewtex: I followed up on the PR.

Common/OpenCL/ITKimprovements/itkOpenCL.h
thewtex: `OPENCL_C_VERSION*` were designed for building against Intel, Nvidia, or AMD vendor toolchains:

https://github.com/SuperElastix/elastix/blob/be0abdcbd8a47837b62310a458c56c9f1579015b/CMake/FindOpenCL.cmake#L410

https://github.com/SuperElastix/elastix/blob/be0abdcbd8a47837b62310a458c56c9f1579015b/CMake/FindOpenCL.cmake#L473-L493

`CL_TARGET_OPENCL_VERSION` is intended for the Khronos OpenCL headers per:

https://github.com/KhronosGroup/OpenCL-Headers/blob/master/README.md#compiling-for-a-specific-opencl-version

`120` is a good default / proper value because it is the supported value across most implementations. On macOS in particular, support ended at this version.

thewtex:
Closing in favor of https://github.com/InsightSoftwareConsortium/ITK/pull/2337

PR #427, opened 2021-02-20
ENH: Replace itkGPUKernelClassMacro calls by GetOpenCLSourceOf functions

N-Dekker:
Removed all (n = 22) `itkGPUKernelClassMacro` macro calls, as there is an upcoming breaking change of the macro definition with ITK 5.2 (introduced with ITK v5.2rc01).

Replaced the class definitions that were created by the macro by simple functions of the form `GetOpenCLSourceOf...Kernel()`, which just return the OpenCL source code as a string.

Allows compilation on ITK 5.1.1 and 5.1.2, as well as ITK 5.2.0 (release candidate 1 or 2).

### DIFF
--- a/CMake/elastixOpenCL.cmake
+++ b/CMake/elastixOpenCL.cmake
@@ -73,7 +73,7 @@ macro(write_opencl_kernel_to_file _opencl_file _src_var _group_name)
   set(${kernel_cxx_class_name}_KernelString
     "${${kernel_cxx_class_name}_KernelString}{\n\n")
   set(${kernel_cxx_class_name}_KernelString
-    "${${kernel_cxx_class_name}_KernelString}const char* ${kernel_cxx_class_name}::GetOpenCLSource()\n")
+    "${${kernel_cxx_class_name}_KernelString}const char* GetOpenCLSourceOf${kernel_cxx_class_name}()\n")
   set(${kernel_cxx_class_name}_KernelString
     "${${kernel_cxx_class_name}_KernelString}{\n")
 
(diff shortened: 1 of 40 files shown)


N-Dekker:
For the record, I also informed Denis (@dpshamonin) about this pull request, this morning.

thewtex:
Why is it preferred to change these names?

This macro is now not valid:

https://github.com/InsightSoftwareConsortium/ITK/blob/9ff84420e55d2f86984f40f14861bddf0812b82b/Modules/Core/Common/include/itkMacro.h#L1230-L1231

and Python API, e.g.:

https://github.com/InsightSoftwareConsortium/ITK/blob/9ff84420e55d2f86984f40f14861bddf0812b82b/Modules/Core/GPUCommon/wrapping/test/itkGPUImageTest.py#L57-L60

It also seems better to have a consistent method name without the requirement to adjust the call for every object.

N-Dekker:
Thanks for your feedback so far @thewtex  However, one way or the other, it seems too early to drop ITK 5.1 compatibility for elastix right now. Can the ITK pull request I just did possibly provide a way out of this issue? Please have a look: "COMP: Fix itkGPUKernelClassMacro(kernel) backward compatibility", https://github.com/InsightSoftwareConsortium/ITK/pull/2337

Obviously, when https://github.com/InsightSoftwareConsortium/ITK/pull/2337 would be accepted, this SuperElastix/elastix pull request (#427) isn't really necessary anymore, and may be dropped.

N-Dekker:
Update: this pull request is marked as "draft", and is intended to be dropped, once pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2337 "COMP: Fix itkGPUKernelClassMacro(kernel) backward compatibility" is merged onto ITK master branch

N-Dekker:
Closed: Superseded by pull request InsightSoftwareConsortium/ITK#2337 commit https://github.com/InsightSoftwareConsortium/ITK/commit/6c06a28d4314f0b314a9a9410317f1b942ad6d96 "COMP: Fix itkGPUKernelClassMacro(kernel) backward compatibility"

PR #429, opened 2021-02-26
Allow disabling all output from `xout`, to ease multi-threading support

N-Dekker:
Added `DisableOutput(void)` member function to both `elx::ElastixFilter` and  `elx::TransformixFilter`, which allows skipping an internal `xoutSetup` call, during `GenerateData()`. Effectively suppressed the output that is otherwise produced, when doing `xl::xout["some-stream"] << "some-output"`.

Aims to ease multi-threading support, as discussed with Konstantinos Ntatsis (@ntatsisk) .

### DIFF
--- a/Common/xout/xoutmain.cxx
+++ b/Common/xout/xoutmain.cxx
@@ -20,26 +20,14 @@
 
 namespace xoutlibrary
 {
-static xoutmain * local_xout = nullptr;
-
 xoutmain &
 get_xout(void)
 {
-  return *local_xout;
+  // Note: C++11 "magic statics" ensures that the construction of a local
+  // static variable like this is thread-safe.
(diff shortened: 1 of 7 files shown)


ntatsisk:
Hi @N-Dekker, I tried this PR locally and I confirm that it solves the issues with multi-threading in the application I am working on. The results and the performance gain is as expected. I only had to repeat the changes for `Transformix` since I use that in parallel as well. So, could you also add `TransformixFilter::DisableOutput()` together with skipping its `xoutSetup` call here? Thanks!

N-Dekker:
Discussed at the weekly internal elastix sprint. Marius (@mstaring) and Stefan (@stefanklein) think it's OK  😃

PR #433, opened 2021-03-02
STYLE: Remove 14 unnecessary dynamic_casts from T* tot T*

N-Dekker:
Removed "no-op" dynamic_casts, that just cast a pointer to its own compile-time type.

Remove unused `CheckForAdvancedTransform(void)` member functions, which have such "no-op" dynamic_casts.

Removal of dynamic_casts could ease avoiding the MacOS segfaults that we observed.

### DIFF
--- a/Components/Metrics/MissingStructurePenalty/elxMissingStructurePenalty.hxx
+++ b/Components/Metrics/MissingStructurePenalty/elxMissingStructurePenalty.hxx
@@ -154,7 +154,7 @@ MissingStructurePenalty<TElastix>::BeforeRegistration(void)
       this->ReadMesh(fixedMeshName, fixedMesh);
     }
 
-    meshPointerContainer->SetElement(meshNumber, dynamic_cast<MeshType *>(fixedMesh.GetPointer()));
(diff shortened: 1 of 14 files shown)


N-Dekker:
For the record, I detected those "no-op" `T*` to `T*` casts  by locally doing a `#define dynamic_cast MyDynamicCast`, with the following definition of `MyDynamicCast`: 


```
#include <type_traits>

template<class TPointer, class U>
TPointer
MyDynamicCast(U* source)
{
  using T = typename std::remove_pointer<TPointer>::type;
  static_assert(!std::is_same<typename std::remove_cv<T>::type, typename std::remove_cv<U>::type>::value,
      "no-op dynamic_cast<T*>!");
  return dynamic_cast<T*>(source);
}

template <class T, class U>
T &
MyDynamicCast(U& source)
{
  static_assert(!std::is_same<typename std::remove_cv<T>::type, typename std::remove_cv<U>::type>::value,
      "no-op dynamic_cast<T&>!");
  return dynamic_cast<T &>(source);
}
```

mstaring:
lgtm

PR #436, opened 2021-03-08
ENH: Add Conversion::ToOptimizerParameters(const std::vector<double> &)

N-Dekker:
Including GoogleTest unit test.


N-Dekker:
FYI: This is a helper function, intended for an upcoming commit to add support for external transform files (ITK TFM, HDF5) as input.

PR #437, opened 2021-03-08
ENH: Add RetrieveValues to ParameterMapInterface and elx::Configuration

N-Dekker:
Added convenience member functions, `itk::ParameterMapInterface::RetrieveValues(parameterName)` and `elx::Configuration::RetrieveValuesOfParameter(parameterName)`.

Included six `ParameterMapInterface` GoogleTest unit tests.

### DIFF
--- a/Common/GTesting/CMakeLists.txt
+++ b/Common/GTesting/CMakeLists.txt
@@ -6,6 +6,7 @@ add_executable(CommonGTest
   elxResamplerGTest.cxx
   elxTransformIOGTest.cxx
   itkComputeImageExtremaFilterGTest.cxx
+  itkParameterMapInterfaceTest.cxx
   )
 target_link_libraries(CommonGTest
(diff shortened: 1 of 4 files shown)


N-Dekker:
FYI: This is a helper function, intended for an upcoming commit to add support for external transform files (ITK TFM, HDF5) as input.

The proposed `ParameterMapInterface` member function `RetrieveValues<T>(parameterName)` is somewhat similar to the existing `ReadParameter(parameterValues, parameterName, entry_nr_start, entry_nr_end, printThisErrorMessage, errorMessage)` from https://github.com/SuperElastix/elastix/blob/5.0.1/Common/ParameterFileParser/itkParameterMapInterface.h#L286-L293 However, `RetrieveValues<T>` is more convenient for a common use case, where all values are retrieved at once (rather than just between `entry_nr_start` and `entry_nr_end`).

PR #438, opened 2021-03-08
ENH: Read data from an external transform file into the parameter map

N-Dekker:
Both `Configuration::Initialize` overloads now add values from an external file, when the specified parameter map has parameter `("Transform" "File")`, and a "TransformFileName" parameter, which specifies the external file.

The values are stored into the parameters "ITKTransformParameters", "ITKTransformFixedParameters", and "ITKTransformType".

The value of the "Transform" parameter is then set to the elastix transform type name that corresponds with the "ITKTransformType" value.

`TransformBase::ReadFromFile()` does retrieve these new parameters and loads their values into the transform object.


Core/Configuration/elxConfiguration.cxx
```diff
@@ -18,9 +18,49 @@
...
 
+#include "elxConversion.h"
+#include "elxTransformIO.h"
+
 namespace elastix
 {
+namespace
+{
+itk::ParameterFileParser::ParameterMapType
+AddDataFromExternalTransformFile(itk::ParameterFileParser::ParameterMapType parameterMap)
```
N-Dekker, line 29: @mstaring FYI Based on your feedback, I just looked at the possibility of passing the `parameterMap` parameter by reference, and then doing the adjustments "in-place". This does not appear possible, because the original parameter map argument in both `Configuration::Initialize` overloads is a `const` reference! So I think it's best to just leave it this way, for now.

Common/Transforms/elxTransformIO.cxx
```diff
@@ -79,6 +106,24 @@ elastix::TransformIO::Write(const itk::TransformBase & itkTransform, const std::
 }
 
 
+itk::SmartPointer<itk::TransformBase>
```
mstaring, line 109: If I remember correctly ITK has the habit to return normal pointers, which are then convertible to smartpointer. There is probably a reason behind that habit, please check.

N-Dekker:
Back to draft status: major revision coming soon  😃

Common/Transforms/elxTransformIO.cxx
N-Dekker: Thanks for looking at this, Marius. In this case I think returning a smart pointer is fine, to reduce the chance that the caller will accidentally use a normal pointer.

You see, the following would crash:

    Object* GetNormalPointer(void)
    {
        return Object::New();
    }

    int main(void)
    {
        Object* object = GetNormalPointer();
        object->Print(std::cout);  // Crash! Object is already destructed!
    }

N-Dekker:
Update: included a unit test that tests both ITK .tfm and HDF5 as external transform file, linked to by an elastix TransformParameters text file, respectively:

    (Transform "File")
    (TransformFileName "ITK-HDF5-Transform.h5")

and

    (Transform "File")
    (TransformFileName "ITK-Transform.tfm")

N-Dekker:
For the sake of time, I would like merge this pull request once the CI tests have passed successfully.

PR #442, opened 2021-03-19
ENH: Lift ParameterFileParser restrictions on chars of parameter values

N-Dekker:
Removed checks on special characters (",;!@#$%&|<>?") from `ParameterFileParser::GetParameterFromLine` for parameter values, in order to be more permissive towards free text strings, and to allow any file path name as parameter value.


N-Dekker:
Discussed with Marius at our internal slack channel.

PR #298, opened 2020-09-24
ENH: Replace __ELASTIX_VERSION by ELASTIX_VERSION_MAJOR/MINOR/PATCH

N-Dekker:
And added `GetElastixVersionString()` helper function.

(Discussed with Marius Staring at internal superelastix.slack.com channel #elastix)

PR #446, opened 2021-03-26
Fixes for building with SimpleElastix v2.0.2

kaspermarstal:
The update of ITK to v5 means SimpleElastix has to be updated to the newest of SimpleITK as well. These changes were required for SimpleElastix (based on lastest SimpleITK) to be able to build Elastix. However, the changes are quite fundamental and I would expect you to have noticed before. Can you review @N-Dekker?


N-Dekker:
@kaspermarstal Sorry, I did not mean to break the SimpleElastix build, when removing `ELASTIX_VERSION`. I did so last September, by pull request https://github.com/SuperElastix/elastix/pull/298 commit 90122eff6b37a17adc9f5bb3d7bc08608e9c14a6 "ENH: Replace __ELASTIX_VERSION by ELASTIX_VERSION_MAJOR/MINOR/PATCH" whose main aim was to start using three version numbers (major.minor.patch), instead of a floating point (or fixed point) number. The removal of `ELASTIX_VERSION` did not break any test at that time (September 2020), so I thought it would be OK. Obviously I was wrong!

Core/Install/elxConversion.h
```diff
@@ -20,7 +20,7 @@
 
 #include "itkMatrix.h"
 
-#include <vnl_vector.h>
+#include <vnl/vnl_vector.h>
```
N-Dekker, line 23: @kaspermarstal No problem with this adjustment, bu I just wonder, does SimpleElastix directly use `elx::Conversion`? Or is this header file accidentally leaking into the SimpleElastix build, via some indirect inclusion?

CMake/elastixVersion.cmake
```diff
@@ -2,3 +2,5 @@
 set(ELASTIX_VERSION_MAJOR "5")
 set(ELASTIX_VERSION_MINOR "0")
 set(ELASTIX_VERSION_PATCH "1")
+
+set(ELASTIX_VERSION "${ELASTIX_VERSION_MAJOR}.${ELASTIX_VERSION_MINOR}.${ELASTIX_VERSION_PATCH}")
```
N-Dekker, line 6: Looks like we might also use this `ELASTIX_VERSION` CMake variable for the C++ `#define ELASTIX_VERSION_STRING`, now in: https://github.com/SuperElastix/elastix/blob/5.0.1/Core/elxVersionMacros.h.in#L33  😃

N-Dekker (approved):
Thank you Kasper!

Core/Install/elxConversion.h
kaspermarstal: It's probably leaking, I don't think we use it directly. Will take a look at the SimpleElastix side!

N-Dekker: Still wondering why `#include <vnl_vector.h>` worked fine while building elastix, when it does not for SimpleElastix. But anyway, I'll just merge your PR now  😃

PR #447, opened 2021-03-26
STYLE: Remove duplicate itkAdvancedTranslationTransform from Components

N-Dekker:
Removed the duplicate definition of `itk::AdvancedTranslationTransform` from Components/Transforms/TranslationStackTransform. It is also at Common/Transforms !


N-Dekker:
@mstaring @stefanklein It appears that there are two copies of "itkAdvancedTranslationTransform.h" that are  basically identical:

https://github.com/SuperElastix/elastix/blob/5.0.1/Components/Transforms/TranslationStackTransform/itkAdvancedTranslationTransform.h
https://github.com/SuperElastix/elastix/blob/5.0.1/Common/Transforms/itkAdvancedTranslationTransform.h

Their implementations are also identical (except for possible tiny cosmetic differences):
https://github.com/SuperElastix/elastix/blob/5.0.1/Components/Transforms/TranslationStackTransform/itkAdvancedTranslationTransform.txx versus
https://github.com/SuperElastix/elastix/blob/5.0.1/Common/Transforms/itkAdvancedTranslationTransform.hxx

I hope it's OK to remove the ones from "Components/Transforms/TranslationStackTransform", to ease maintenance.

stefanklein:
Hey Niels, yes if they are functionally identical, then one could be removed

PR #449, opened 2021-03-28
BUG: Remove unreachable code in Common\OpenCL, after `itkExceptionMacro`

N-Dekker:
Removed the unreachable code after `itkExceptionMacro` calls (which throw an exception), from two cxx files in Common\OpenCL\ITKimprovements.

Fixed Visual C++ Warning Level 4 warning C4702, "unreachable code".

### DIFF
--- a/Common/OpenCL/ITKimprovements/itkOpenCLLogger.cxx
+++ b/Common/OpenCL/ITKimprovements/itkOpenCLLogger.cxx
@@ -104,10 +104,6 @@ OpenCLLogger::Initialize()
   if (this->m_FileStream->fail())
   {
     itkExceptionMacro(<< "Unable to open file: " << logFileName);
-    delete this->m_FileStream;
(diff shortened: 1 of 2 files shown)


Common/OpenCL/ITKimprovements/itkOpenCLLogger.cxx
```diff
@@ -104,10 +104,6 @@ OpenCLLogger::Initialize()
   if (this->m_FileStream->fail())
   {
     itkExceptionMacro(<< "Unable to open file: " << logFileName);
-    delete this->m_FileStream;
-    this->m_FileStream = nullptr;
-    this->m_Created = false;
```
N-Dekker, line 109: @dpshamonin Do you happen to know, should these three lines of code (starting with `delete this->m_FileStream`) be placed _before_, instead of _after_ the `itkExceptionMacro` call?

PR #426, opened 2021-02-20
CI:CTest results on cdash again

ViktorvdValk:

PR #451, opened 2021-03-30
CI:Elastix to my.cdash update

ViktorvdValk:


ViktorvdValk:
This is the working and correct version, I'll close PR #426

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -13,25 +13,25 @@ jobs:
...
       max-parallel: 3
       matrix:
-        os: [ubuntu-18.04, windows-2019, macos-10.15]
+        os: [macos-10.15, ubuntu-18.04, windows-2019]
         include:
           - os: ubuntu-18.04
             c-compiler: "gcc"
             cxx-compiler: "g++"
-            itk-git-tag: "v5.1.2"
+            itk-git-tag: "0fce3ff86b9242a298a55a42337769eca8e8a4c8"
```
N-Dekker, line 21: @ViktorvdValk Thanks so far... Regarding the upgrade of ITK, I would prefer to have that as a separate pull request. Probably "v.5.2.0", instead of "0fce3ff86b9242a298a55a42337769eca8e8a4c8", right?

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -87,40 +91,60 @@ jobs:
         mkdir ITK-build
         cd ITK-build
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
-        cmake -DCMAKE_C_COMPILER:FILEPATH="${{ matrix.c-compiler }}" -DBUILD_SHARED_LIBS:BOOL=OFF -DCMAKE_CXX_COMPILER="${{ matrix.cxx-compiler }}" -DCMAKE_BUILD_TYPE:STRING=${{ matrix.cmake-build-type }} -DBUILD_EXAMPLES=OFF -DBUILD_TESTING:BOOL=OFF -DITK_LEGACY_REMOVE=ON -GNinja ../ITK-source
+        cmake -DCMAKE_C_COMPILER:FILEPATH="${{ matrix.c-compiler }}" -DBUILD_SHARED_LIBS:BOOL=ON -DCMAKE_CXX_COMPILER="${{ matrix.cxx-compiler }}" -DCMAKE_BUILD_TYPE:STRING=${{ matrix.cmake-build-type }} -DBUILD_EXAMPLES=OFF -DBUILD_TESTING:BOOL=OFF -GNinja ../ITK-source
```
N-Dekker, line 94: Is `BUILD_SHARED_LIBS:BOOL=ON` necessary for cdash?

Testing/Dashboard/elxCommonCDash.cmake
```diff
@@ -0,0 +1,393 @@
+# ITK Common Dashboard Script
```
N-Dekker, line 1: "ITK Common Dashboard Script", should that be "_elastix_ Common Dashboard Script"? Or is it a copy from ITK? If it's a copy, could you please add a link to the original?

Testing/Dashboard/elxDashboardCommon.cmake
```diff
@@ -204,8 +200,8 @@ if( dashboard_model STREQUAL Continuous )
     endif()
 
     # Check for changes
-    ctest_update( SOURCE ${CTEST_DASHBOARD_ROOT} RETURN_VALUE res )
-    message( "Found ${res} changed files" )
+    # ctest_update( SOURCE ${CTEST_DASHBOARD_ROOT} RETURN_VALUE res )
+    # message( "Found ${res} changed files" )
```
N-Dekker, line 204: If these lines of code are no longer relevant, please just remove them. They will still be in git history anyway.

.github/workflows/ElastixGitHubActions.yml
N-Dekker: Ah, I see you adjusted the itk-git-tag to "v5.2.0"  😃

Testing/Dashboard/elxCommonCDash.cmake
N-Dekker: @ViktorvdValk Thanks for adding the link to https://raw.githubusercontent.com/InsightSoftwareConsortium/ITK/dashboard/itk_common.cmake !
Do you think it's necessary to keep maintaining our own copy of this file?  I'm somewhat concerned about future maintenance... 🤔

Testing/Dashboard/elxDashboardCommon.cmake
N-Dekker: Thanks for removing those lines!

Testing/Dashboard/elxCommonCDash.cmake
ViktorvdValk: I did change quite a lot in the file, so we should maintain our own copy I think.

N-Dekker:
Thanks for "CI: Use sha on cdash", but unfortunately I still see no revision SHA at https://my.cdash.org/index.php?project=elastix 🤔 

I discussed this at the CDash mailing list three years ago already, it should be possible by now! [CDash] Where and how to display git SHA of commit at CDash dashboard? https://public.kitware.com/pipermail/cdash/2018-January/001839.html  Maybe those CMake variables should be set _just_ before `ctest_update()` ?

N-Dekker:
Thanks @ViktorvdValk !

Issue #454, opened 2021-04-07
elastix crash when LANG=fr_FR.UTF-8


N-Dekker:
Reported by Aurelie at https://groups.google.com/g/elastix-imageregistration/c/70PXd-Hy_Xo :

> I have a problem when I run elastix computation or when I apply a transform on linux environment when the locale is set to french.
For exemple, the spacing in the transform parameters file is 0.2 0.4 6.5, but transformix crash with an error message saying that spacing value 0 is forbidden (spacing value : 0 0 6). It seems that strings are troncated
The same code on machine with locale set to en or us is ok.
Is it possible that the library converts strings in a different way according system LANG var?

N-Dekker:
Fixed by pull request  #455 commit 96a38f1fc3dfb5361311073f1ff4145628368178  "BUG: Use double-conversion instead of std::istringstream in ParameterMap"

PR #455, opened 2021-04-08
BUG: Use double-conversion instead of std::istringstream in ParameterMap

N-Dekker:
Replaced `std::istringstream` by `StringToDoubleConverter` from Google's double-conversion library <https://github.com/google/double-conversion>, inside the implementation of `ParameterMapInterface::StringCastToFloatingPoint`.

`double_conversion::StringToDoubleConverter` is not affected by locale settings (specifically the decimal separator), so this commit should fix issue https://github.com/SuperElastix/elastix/issues/454 "elastix crash when LANG=fr_FR.UTF-8".

The new `StringCastToFloatingPoint` implementation detects more types of user input errors, for example when a user tries to use a comma as decimal separator (as in "2,375").

This commit also cleans up, by removing code which is no longer necessary now, from "BUG: ParameterMapInterface::StringCast workaround Clang denormal bug", pull request https://github.com/SuperElastix/elastix/pull/404 commit a06e2323166d2ce3119ebcb650fe8ac80fe9b26b

Note that elastix already uses Google's double-conversion library for double-to-string conversion, indirectly, via `itk::NumberToString`.

### DIFF
--- a/Common/GTesting/itkParameterMapInterfaceTest.cxx
+++ b/Common/GTesting/itkParameterMapInterfaceTest.cxx
@@ -98,7 +98,15 @@ GTEST_TEST(ParameterMapInterface, RetrieveValuesThrowsExceptionWhenConversionFai
 {
   const auto        parameterMapInterface = ParameterMapInterface::New();
   const std::string parameterName("Key");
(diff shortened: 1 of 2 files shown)


mstaring:
lgtm

mstaring:
lgtm

PR #466, opened 2021-05-17
STYLE: Remove obsolete externalproject (elxExternalProject)

N-Dekker:
elxExternalProject is no longer maintained. An attempt to build the project causes compilation error, saying:

> elastix\dox\externalproject\itkTimerTest.cxx(14,10): fatal error C1083: Cannot open include file: 'elxTimer.h': No such file or directory

For the record, elxExternalProject was introduced by Stefan Klein (@stefanklein), with commit 81168bedd3bcc049ee96305f3e471edc88509648, 9 January 2011.


mstaring:
lgtm

N-Dekker:
Note: If we agree to simply remove this outdated "externalproject" directory, then we should also remove the related lines from the manual:

https://github.com/SuperElastix/manual/blob/4db8c9e3950735e14c1ee5cf86c891c4f3673145/manual.tex#L2792-L2793

mstaring:
maybe we can instead refer to one of your newer GTests, that also shows how to setup the library in C++ code?

N-Dekker:
> maybe we can instead refer to one of your newer GTests, that also shows how to setup the library in C++ code?

Nice suggestion, thanks @mstaring but the GTests are not really like an external user project. You see, their CMakeLists does not use UseElastix.cmake: https://github.com/SuperElastix/elastix/blob/develop/Core/Main/GTesting/CMakeLists.txt 

While the "externalproject" that would be removed by this pull request does use UseElastix.cmake:
https://github.com/SuperElastix/elastix/blob/5.0.1/dox/externalproject/CMakeLists.txt

Moreover, the GTests depend on GoogleTest, which is not necessary the case for external user projects, of course.

Instead of removing "externalproject" entirely, we could also just fix it and make it work again. What do you think?

mstaring:
> Instead of removing "externalproject" entirely, we could also just fix it and make it work again. What do you think?

That is a good idea, I think this still has a use case, namely when people try to integrate elastix themselves

N-Dekker:
It takes some more time than I expected, to fix externalproject, because the UseElastix.cmake mechanism from https://github.com/SuperElastix/elastix/blob/5.0.1/dox/externalproject/CMakeLists.txt does not seem to work anymore.

UseElastix.cmake suggests doing:

    set( Elastix_DIR "path/to/elastix/binary/directory" )
    find_package( Elastix REQUIRED )
    include( ${ELASTIX_USE_FILE} )

https://github.com/SuperElastix/elastix/blob/5.0.1/UseElastix.cmake.in#L5-L7

https://github.com/SuperElastix/elastix/blob/5.0.1/dox/externalproject/CMakeLists.txt does things slightly differently.

Actually I would think that the user would not need to specify the ITK directory, when the user already specifies the elastix directory. Because elastix "knows" where to find ITK.

stefanklein:
Yes, I think it would be nice to fix it. the elxTimer stuff is probably easy to fix.
Stefan

N-Dekker:
@stefanklein Thanks for your comment. Do you really want to have "elxTimer stuff" in there? Wouldn't it be more helpful to have an example "external project" that uses elastix library to do a simple registration?

mstaring:
Yes, I think that is what is meant indeed.

stefanklein:
indeed

N-Dekker:
The example externalproject now does `FIND_PACKAGE( ITK ...)`: https://github.com/SuperElastix/elastix/blob/5.0.1/dox/externalproject/CMakeLists.txt#L7

That should not be necessary, right? When elastix is "found", it seems to me that the path to ITK can automatically be deduced. I guess that via one of the CMake files in de elastix bin directory ("UseElastix.cmake" or "ElastixConfig.cmake").

stefanklein:
Yes you are probably right!

N-Dekker:
A new externalproject example is under construction:

https://github.com/SuperElastix/elastix/tree/externalproject-translation-example

Which is intended to supersede this pull request (#466 - STYLE: Remove obsolete externalproject).

N-Dekker:
Superseded by pull request  #473 "ENH: Revise externalproject example, extend ElastixConfig.cmake file" (which is still under  construction)

PR #475, opened 2021-05-30
COMP: Upgrade ITK from v5.1.1 to v5.2.0

N-Dekker:
Upgraded to ITK version 5.2.0, which was released on May 28, 2021.

ITK 5.2.0 fixed a critical issue, with commit  https://github.com/InsightSoftwareConsortium/ITK/commit/7cebc26dfae73d90dbf4702aa1ccc544e9b70a37 "COMP: Fix OpenJPEG build error with Visual Studio 16.9", by James Butler (@jamesobutler) and @spacelg

Following pull request https://github.com/SuperElastix/SimpleElastix/pull/425 and https://github.com/SuperElastix/SimpleElastix/pull/428 by David Young (@yoda-vid)

Obviously, ITK 5.2.0 has many other improvements: https://github.com/InsightSoftwareConsortium/ITK/releases/tag/v5.2.0

PR #484, opened 2021-06-22
STYLE: Use the new ITK 5.2 `OptimizerParameters` constructors

N-Dekker:
Simplified the code by using these two `explicit` constructors, which are added with ITK 5.2:

    OptimizerParameters(SizeValueType dimension, const ValueType & value);

    OptimizerParameters(const ValueType * inputData, SizeValueType dimension);

Follow-up to pull request https://github.com/SuperElastix/elastix/pull/475 commit 6803b26a73aea270cf7b75e0febc4d53ea21053e "COMP: Upgrade ITK from v5.1.1 to v5.2.0"

### DIFF
--- a/Common/Transforms/itkStackTransform.hxx
+++ b/Common/Transforms/itkStackTransform.hxx
@@ -56,13 +56,8 @@ StackTransform<TScalarType, NInputDimensions, NOutputDimensions>::SetParameters(
   const NumberOfParametersType numSubTransformParameters = this->m_SubTransformContainer[0]->GetNumberOfParameters();
   for (unsigned int t = 0; t < this->m_NumberOfSubTransforms; ++t)
   {
-    // MS, \todo: the new itk::TransformParameters only have constructors taking 1 argument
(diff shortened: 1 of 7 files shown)


Common/Transforms/itkStackTransform.hxx
```diff
@@ -56,13 +56,8 @@ StackTransform<TScalarType, NInputDimensions, NOutputDimensions>::SetParameters(
   const NumberOfParametersType numSubTransformParameters = this->m_SubTransformContainer[0]->GetNumberOfParameters();
   for (unsigned int t = 0; t < this->m_NumberOfSubTransforms; ++t)
   {
-    // MS, \todo: the new itk::TransformParameters only have constructors taking 1 argument
```
N-Dekker, line 59: @mstaring This pull request finally fixes the `\todo` that you wrote many years ago 😃  The new (ITK 5.2) `itk::OptmizerParameters` does have a few more convenient constructors than before, including one that accepts a data pointer and a size as arguments! No need to use a temporary `subarray` anymore!

mstaring: nice work and lgtm!

Issue #464, opened 2021-05-17
Support HDF5 or TFM format for all transforms in transform parameters file.


ViktorvdValk:
Support HDF5 or TFM file format by setting ("Transform" "File") and (TransformFileName "transform.tfm") or (TransformFileName "transform.hfd5") in the transform parameters .txt file.

N-Dekker:
For the record, here is an overview of ITK and elastix transform types, as well as their number of parameters: pull request https://github.com/SuperElastix/elastix/pull/358

PR #486, opened 2021-06-29
STYLE: Remove `elastix::TransformBase::m_TransformParametersPointer`

N-Dekker:
The data member `m_TransformParametersPointer` appears unnecessary. Removing this data member eases the implementation of issue https://github.com/SuperElastix/elastix/issues/464 "Support HDF5 or TFM format for all transforms..."

Discussed with Marius (@mstaring) at the internal elastix sprint.

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.h
+++ b/Core/ComponentBaseClasses/elxTransformBase.h
@@ -32,8 +32,6 @@
 #include <itkImage.h>
 #include <itkOptimizerParameters.h>
 
-#include <memory> // For unique_ptr.
-
 namespace elastix
 {
 // using namespace itk; //Not here, because a TransformBase class was added to ITK...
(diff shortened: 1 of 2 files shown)


N-Dekker:
Hmmm, could this comment about `m_FinalParameters` explain the reason for `m_TransformParametersPointer` as well? 🤔 

```
  /** Make a local copy, since some transforms do not do this,
   * like the B-spline transform.
   */
```
At https://github.com/SuperElastix/elastix/blob/4.9.0/Core/ComponentBaseClasses/elxTransformBase.hxx#L350-L352

N-Dekker:
Now I see: `AdvancedBSplineDeformableTransformBase::SetParameters(const ParametersType & parameters)` keeps a pointer to the parameters, so that's why `elastix::TransformBase` extends the lifetime of the parameters object.

https://github.com/SuperElastix/elastix/blob/5d13758a38527011b95fb36bbd3c0d568eeee931/Common/Transforms/itkAdvancedBSplineDeformableTransformBase.hxx#L285-L286

N-Dekker:
Superseded by pull request #488 "STYLE: Replace m_TransformParametersPointer by ParametersType member var"

PR #488, opened 2021-06-29
STYLE: Replace m_TransformParametersPointer by ParametersType member var

N-Dekker:
Simplified `elastix::TransformBase<TElastix>::ReadFromFile(void)` by using `ParametersType` instead of `std::unique_ptr<ParametersType>`.

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.h
+++ b/Core/ComponentBaseClasses/elxTransformBase.h
@@ -32,8 +32,6 @@
 #include <itkImage.h>
 #include <itkOptimizerParameters.h>
 
-#include <memory> // For unique_ptr.
-
 namespace elastix
 {
 // using namespace itk; //Not here, because a TransformBase class was added to ITK...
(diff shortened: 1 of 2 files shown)


N-Dekker:
Supersedes pull request  #486 "STYLE: Remove elastix::TransformBase::m_TransformParametersPointer". It appears unfeasible unfortunately to simply remove `m_TransformParametersPointer`, as the `parameters` object must be kept alive beyond the `ReadFromFile()` function call, in order to be used afterwards by `AdvancedBSplineDeformableTransformBase`.

mstaring:
lgtm. 
Did you check if the Bspline really needs this pointer, or can it perhaps be refactored to do without?

N-Dekker:
> Did you check if the Bspline really needs this pointer, or can it perhaps be refactored to do without?

I believe it could be refactored, but it does not seem easy. There are actually two transform classes that have a `m_InputParametersPointer` pointing at the argument of their `SetParameters(const ParametersType & parameters)` override: MultiBSplineDeformableTransformWithNormal and AdvancedBSplineDeformableTransform. Both of them even modify the parameters via this pointer (casting away their const-ness)!

I guess it was implemented this way to avoid copying a large list of parameters, right?

C++11 introduced "move semantics", which allows "moving" parameters from caller to callee efficiently, without copying or dynamic memory allocation. ITK does already make use of this feature, but not yet for their `TransformBaseTemplate::SetParameters`, and not yet for their `ParametersType`. Could be a nice addition to ITK 😸

PR #489, opened 2021-06-29
Support reading Euler and Similarity transforms from ITK (HDF5, TFM) into elastix

N-Dekker:
Regarding issue #464 "Support HDF5 or TFM format for all transforms"

### DIFF
--- a/Common/GTesting/elxTransformIOGTest.cxx
+++ b/Common/GTesting/elxTransformIOGTest.cxx
@@ -842,3 +842,27 @@ GTEST_TEST(Transform, TransformedPointSameAsITKEuler3D)
 
   Expect_elx_TransformPoint_yields_same_point_as_ITK<elx::EulerTransformElastix>(*itkTransform);
 }
+
+
+GTEST_TEST(Transform, TransformedPointSameAsITKSimilarity2D)
+{
+  const auto itkTransform = itk::Similarity2DTransform<double>::New();
+  itkTransform->SetScale(0.75);
+  itkTransform->SetTranslation(MakeVector(1.0, 2.0));
+  itkTransform->SetCenter(MakePoint(0.5, 1.5));
+  itkTransform->SetAngle(M_PI_4);
+
+  Expect_elx_TransformPoint_yields_same_point_as_ITK<elx::SimilarityTransformElastix>(*itkTransform);
+}
+
+
+GTEST_TEST(Transform, TransformedPointSameAsITKSimilarity3D)
+{
+  const auto itkTransform = itk::Similarity3DTransform<double>::New();
+  itkTransform->SetScale(0.75);
+  itkTransform->SetTranslation(MakeVector(1.0, 2.0, 3.0));
+  itkTransform->SetCenter(MakePoint(3.0, 2.0, 1.0));
+  itkTransform->SetRotation(itk::Vector<double, 3>(1.0), M_PI_4);
+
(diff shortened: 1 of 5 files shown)


mstaring:
lgtm

N-Dekker:
@mstaring Just added Similarity support, simply by adding an `if (!this->HasITKTransformParameters())` to the implementation of `SimilarityTransformElastix::ReadFromFile(void)`, meaning that it simply skips reading the `centerOfRotation` when the configuration has "ITKTransformParameters". Does that look OK to you?

Also added unit tests, of course  😃

mstaring:
if the tests work it should be good! The logic makes sense to me.

N-Dekker:
@mstaring Still haven't got elastix Similarity in sync with `itk::Similarity3DTransform`, when both center and scale are set. I don't know but it looks like an ITK issue. 2D and 3D seem to be inconsistent with ITK Similarity as well:

    const auto itkTransform2D = itk::Similarity2DTransform<double>::New();
    itkTransform2D->SetCenter(itk::Point<double, 2>({ 1.0, 1.0 }));
    itkTransform2D->SetScale(2.0);
    std::cout << itkTransform2D->TransformPoint(itk::Point<double, 2>({ 0.0, 0.0 })) << '\n';

Prints: _[-1, -1]_

However, in 3D:

    const auto itkTransform3D = itk::Similarity3DTransform<double>::New();
    itkTransform3D->SetCenter(itk::Point<double>({ 1.0, 1.0, 1.0 }));
    itkTransform3D->SetScale(2.0);
    std::cout << itkTransform3D->TransformPoint(itk::Point<double>({ 0.0, 0.0, 0.0 })) << '\n';

Print: _[0, 0, 0]_

While elastix `SimilarityTransformElastix` produces _[-1, -1]_ for 2D, and _[-1, -1, -1]_ for 3D.

Do you agree that ITK Similarity3DTransform appears wrong in this case?

mstaring:
I am not sure this transform is fully specified with just center and scale. Can you print GetParameters() for both? Perhaps there is an initialization difference.

In principle there are 7 parameters in 3D: 3 rotations, 3 translations, 1 scale

N-Dekker:
> I am not sure this transform is fully specified with just center and scale. Can you print GetParameters() for both? (and also fixed parameters)

2D:
```
ITK Parameters:      [2, 0, 0, 0]
ITK FixedParameters: [1, 1]
ELX Parameters:      [2, 0, 0, 0]
ELX FixedParameters: [1, 1]

```

3D:
```

ITK Parameters:      [0, 0, 0, 0, 0, 0, 2]
ITK FixedParameters: [1, 1, 1]
ELX Parameters:      [0, 0, 0, 0, 0, 0, 2]
ELX FixedParameters: [1, 1, 1]

```

mstaring:
Now that we have confirmed that both parameters and fixed parameters are the same, it is weird that the transformpoint returns different results. Can you compare their implementations, to see what the differences are? Did something change for example in the time we made the elastix copy and now?

N-Dekker:
> are intermediate matrices that the transformpoint function uses, updated correctly?

@mstaring  Ah, now I see, it appears that a call to ITK's `Similarity3DTransform::SetScale` does not update its internal `MatrixOffsetTransformBase::m_Offset`. On the other hand, `SetCenter` and `SetTranslation` do update `m_Offset`, and when the offset is updated, ITK and Elastix produce the same result. 😃 

My workaround for the unit tests is to do the `SetScale` call _before_  calling `SetCenter` and `SetTranslation`, to ensure that the offset is eventually updated. .

PR #496, opened 2021-07-08
BUG: `AdvancedSimilarity3DTransform::SetScale` should recompute m_Offset

N-Dekker:
Following ITK issue https://github.com/InsightSoftwareConsortium/ITK/issues/2629 and pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2627


mstaring:
lgtm

N-Dekker:
I'm not entirely sure if it's really necessary in elastix, for this `SetScale` member function to call `ComputeOffset()`. Do end-users possibly directly call this `SetScale`?

Anyway, ITK users would certainly directly call the corresponding ITK `SetScale`, so in ITK it was certainly necessary. So this pull request is consistent with ITK.

PR #508, opened 2021-08-07
Support MaximumNumberOfIterations = 0, allowing to just do an AutomaticTransformInitialization

N-Dekker:
Until now, optimizers appear to do a single iteration step, even when the elastix parameter "MaximumNumberOfIterations" has the value "0". It appears that those optimizers did the check `if (m_CurrentIteration >= m_NumberOfIterations) break;` when it was too late already to avoid the first iteration. 

Supporting MaximumNumberOfIterations = 0 (by doing no optimizer iteration at all) allows inspecting the result of just doing an AutomaticTransformInitialization. As hereby tested.

### DIFF
--- a/Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx
+++ b/Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx
@@ -87,6 +87,103 @@ GTEST_TEST(itkElastixRegistrationMethod, Translation)
 }
 
 
+// Tests "MaximumNumberOfIterations" value "0"
+GTEST_TEST(itkElastixRegistrationMethod, MaximumNumberOfIterationsZero)
+{
+  constexpr auto ImageDimension = 2U;
+  using ImageType = itk::Image<float, ImageDimension>;
+  using SizeType = itk::Size<ImageDimension>;
+  using IndexType = itk::Index<ImageDimension>;
+  using OffsetType = itk::Offset<ImageDimension>;
+
+  const OffsetType translationOffset{ { 1, -2 } };
+  const auto       regionSize = SizeType::Filled(2);
+  const SizeType   imageSize{ { 5, 6 } };
+  const IndexType  fixedImageRegionIndex{ { 1, 3 } };
+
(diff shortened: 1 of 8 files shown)


mstaring:
lgtm

N-Dekker:
Thanks @mstaring  Just wondering, are there already other ways to only just run the AutomaticTransformInitialization (from the elastix.exe command-line or the Elastix Filter interface)? Or is this a new feature?  😃

mstaring:
Let's actually discuss at the sprint, because now I am unsure if this is what we want. 
If I recall correctly with just 0 iteration the value and derivative are compute but never applied.

N-Dekker:
@mstaring  It appears to me that until now MaximumNumberOfIterations = 0 was treated exactly like MaximumNumberOfIterations = 1. 

Whatever we do, I think we should not treat 0 and 1 as equal!

Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx
```diff
@@ -87,6 +87,103 @@ GTEST_TEST(itkElastixRegistrationMethod, Translation)
...
+                                                       { "Metric", "AdvancedNormalizedCorrelation" },
+                                                       { "Optimizer", optimizer },
+                                                       { "Transform", "TranslationTransform" } }));
+    filter->Update();
+
+    const auto transformParameters = GetTransformParametersFromFilter(*filter);
+
+    for (const auto & transformParameter : transformParameters)
+    {
+      EXPECT_EQ(transformParameter, 0.0);
```
N-Dekker, line 133: During the sprint, Stefan (@stefanklein) suggested to add an extra test that the transformParameters are _not_ all zero when MaximumNumberOfIterations = 1. This appears to take some more time, as in the current setup, it would cause exceptions for "FiniteDifferenceGradientDescent" and "StandardGradientDescent".  (MaximumNumberOfIterations = 1 only works out of the box with "AdaptiveStochasticGradientDescent", in this specific test case.) So I would like to postpone this extra test for later.... 😃

PR #358, opened 2020-11-15
ENH: Add parameter file support for TransformOutputFileNameExtensions

N-Dekker:
Allows specifying output file types for transform files by their file name extension, for example:

    (TransformOutputFileNameExtensions ".h5" ".mat" ".tfm")

PR #525, opened 2021-09-19
ENH: Write CompositeTransform when combination has multiple transforms

N-Dekker:
When the `AdvancedCombinationTransform` to be written by the `WriteToFile` member function of `elx::TransformBase` has more than one transform, `WriteToFile` now writes an ITK CompositeTransform of all of them to TFM or HDF5 (experimentally).

Removed the BSpline test from `TransformIO.CorrespondingItkTransform`, as it is broken by this enhancement.

Added `itkElastixRegistrationMethod.WriteCompositeTransform` GoogleTest unit test.

### DIFF
--- a/Common/GTesting/elxConversionGTest.cxx
+++ b/Common/GTesting/elxConversionGTest.cxx
@@ -300,7 +300,6 @@ GTEST_TEST(Conversion, ParameterMapToString)
 (Spacing 1 1)
 (Transform "TranslationTransform")
 (TransformParameters 0 0)
-(UseBinaryFormatForTransformationParameters "false")
 (UseDirectionCosines "true")
 )";
 
@@ -318,7 +317,6 @@ GTEST_TEST(Conversion, ParameterMapToString)
                                                { "Spacing", { "1", "1" } },
                                                { "Transform", { "TranslationTransform" } },
                                                { "TransformParameters", { "0", "0" } },
-                                               { "UseBinaryFormatForTransformationParameters", { "false" } },
                                                { "UseDirectionCosines", { "true" } } }),
             expectedString);
 }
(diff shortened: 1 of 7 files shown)


N-Dekker:
This pull request supports combinations of Affine, BSpline, Euler, Similarity, and Translation transformations from an elastix registration to be written into a single ITK TFM or HDF5 file, as an ITK CompositeTransform.

(Users may specify that they want to have such output files from elastix by parameter file entry `(TransformOutputFileNameExtensions ".h5" ".mat" ".tfm")`, as implemented by pull request  #358  "ENH: Add parameter file support for TransformOutputFileNameExtensions")

Questions to be discussed (possibly at the sprint):

1. Is it already a useful feature, given the fact that it is now limited to a combination of those five transform types?
2. What if the "combination" consists of just one transform (i.e. there is no "initial transform", only a "current transform")? Is it OK to then just output the one transform, instead of a "CompositeTransform" of one.
3. What if "HowToCombineTransforms" is "Add", rather than "Compose"? The current pull request just ignores "HowToCombineTransforms".
4. What if the combination contains elastix "advanced" transforms that cannot (yet) be converted to ITK transforms? The current pull request will then just write the one elastix transform to a TFM/HDF5 file.  (Relates to question 1.)

N-Dekker:
elastix already supports configuration parameter "UseBinaryFormatForTransformationParameters". When enabled, the transform parameter "TransformParameters" will have the ".dat" file name, instead of the ITK-style list from `itk::Transform::GetParameters()`. 

Personally I found it clearer to use a separate "TransformFileName" parameter, to refer to an external ITK Transform file (HDF5 or TFM). But if the two options are so similar, should the parameter conversion be synchronized, one way or the other?

N-Dekker:
TODO: 
- Document this! (wiki, example ITKElastix notebook)
- Remove redundant transform parameters from txt file, when using ITK format
- .dat may go.

PR #556, opened 2021-11-25
PERF: Make AdvancedNormalizedCorrelationImageToImageMetric faster

N-Dekker:
Added a new member function to `AdvancedImageToImageMetric`, `FastEvaluateMovingImageValueAndDerivative`, which calls the multi-threaded overload of ITK's `itk::BSplineInterpolateImageFunction::EvaluateValueAndDerivativeAtContinuousIndex`, indirectly. (It does so via `EvaluateMovingImageValueAndDerivativeWithOptionalThreadId`, another new member function.)

Made `AdvancedNormalizedCorrelationImageToImageMetric::ThreadedGetValueAndDerivative` faster, by calling this new `FastEvaluateMovingImageValueAndDerivative`.

A large performance improvement was observed for GoogleTest unit test `itkElastixRegistrationMethod.EulerDiscRotation2D` (which uses the "AdvancedNormalizedCorrelation" metric), from ~1.5 second before this commit down to ~0.9 second after this commit, using Visual Studio 2019, Release configuration. (For a Debug configuration even from ~15 seconds before, down to ~4 seconds after this commit.)

### DIFF
--- a/Common/CostFunctions/itkAdvancedImageToImageMetric.h
+++ b/Common/CostFunctions/itkAdvancedImageToImageMetric.h
@@ -507,7 +507,21 @@ protected:
   virtual bool
   EvaluateMovingImageValueAndDerivative(const MovingImagePointType & mappedPoint,
                                         RealType &                   movingImageValue,
-                                        MovingImageDerivativeType *  gradient) const;
+                                        MovingImageDerivativeType *  gradient) const
+  {
+    return EvaluateMovingImageValueAndDerivativeWithOptionalThreadId(mappedPoint, movingImageValue, gradient);
+  }
+
+  /* A faster version of `EvaluateMovingImageValueAndDerivative`: Non-virtual, using multithreading, and doing less
+   * dynamic memory allocation/decallocation operations, internally. */
+  bool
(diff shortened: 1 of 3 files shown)


N-Dekker:
@mstaring @stefanklein Here is the `EvaluateValueAndDerivativeAtContinuousIndex` overload that was called before, by the metric: 

"Slower, but safer"
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h#L239-L257

Here is the faster overload (which has a `threadId` parameter): 

"FIXME -- Review this "fix" and ensure it works."
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.hxx#L276-L284

The performance improvement comes from skipping the memory allocation of [local `vnl_matrix` variables](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h#L248-L250), not just from using multiple threads!

PR #559, opened 2021-11-30
STYLE: Use angle brackets (`<`, `>`) for `#include`'s of VNL headers

N-Dekker:
It is common practice to use angle brackets, rather than double-quotes, for `#include` directives that include headers of external libraries. Doing so may slightly improve the compilation speed, as well as the performance of the Visual Studio IDE, when opening the elastix project.

By Notepad++ v8.1.9.2, Find in Files:

    Find what: include "vnl/(.+)"
    Replace with: include <vnl/$1>
    Filter: *.hxx;*.h;*.cxx
    [v] Match case
    (*) Regular expression

### DIFF
--- a/Common/CostFunctions/itkAdvancedImageToImageMetric.h
+++ b/Common/CostFunctions/itkAdvancedImageToImageMetric.h
@@ -28,7 +28,7 @@
 #include "itkLimiterFunctionBase.h"
 #include "itkFixedArray.h"
 #include "itkAdvancedTransform.h"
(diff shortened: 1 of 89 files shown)


N-Dekker:
Follows https://fuchsia.dev/fuchsia-src/development/languages/c-cpp/cpp-style#includes which says:

> If the header being included is a system, global, or library header (see [Naming C/C++ objects](https://fuchsia.dev/fuchsia-src/development/languages/c-cpp/naming) for precise definitions), use `<angle brackets>` and the complete name of the header.

PR #560, opened 2021-12-10
ENH: Add TransformedPointLimitedEulerSameAsUnlimitedEuler3D unit test

N-Dekker:
Tests that the new limited Euler transform by Žiga Špiclin (@zigaso) yields the same point-to-point transformation as the unlimited Euler transform, when there is no limit specified.

### DIFF
--- a/Common/GTesting/elxTransformIOGTest.cxx
+++ b/Common/GTesting/elxTransformIOGTest.cxx
@@ -46,6 +46,8 @@
 #include "TranslationTransform/elxTranslationTransform.h"
 #include "WeightedCombinationTransform/elxWeightedCombinationTransform.h"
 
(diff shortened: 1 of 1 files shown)


N-Dekker:
@zigaso This unit test checks that for an `AdvancedLimitedEulerTransform` instance (your PR #511) any input point, `limitedTransform->TransformPoint(inputPoint)` returns the same output point as the corresponding "unlimited" Euler transform `unlimitedTransform->TransformPoint(inputPoint)`. (It does not yet compile on the current "develop" branch because we did not yet merge your PR #511.)

It appears to me that `limitedTransform->TransformPoint(inputPoint)` calls are not affected by upper/lower limits, right?

PR #596, opened 2022-01-31
BUG: BSplineTransform should not read config params if it has ITK params

N-Dekker:
`RecursiveBSplineTransform::ReadFromFile()` should not do `Configuration::ReadParameter` if it has "ITKTransformParameters".

On the other hand, the old `AdvancedBSplineTransform` should not get its data from an ITK transform.


N-Dekker:
Note that m_Cyclic is not stored in FixedParameters. TODO To be documented!

Issue #647, opened 2022-05-13
[GPU] Compilation fails with ELASTIX_USE_OPENCL: ON


georg-walther:
Hi,
I am trying to compile elastix with ELASTIX_USE_OPENCL: ON to make use of the GPU. I followed the steps at [[https://simpleelastix.readthedocs.io/GettingStarted.html#manually-building-on-linux](https://simpleelastix.readthedocs.io/GettingStarted.html#manually-building-on-linux)] to build ITK with ITK_USE_GPU: OFF (like it was suggested here: https://github.com/SuperElastix/elastix/issues/15).
To build elastix I used the following variables:
ITK_DIR: /home/username/ITK/build
ELASTIX_INCLUDE_DIR: include
ELASTIX_USE_OPENCL: ON
OPENCL_INCLUDE_DIRS: /usr/local/cuda/include
OPENCL_LIBRARIES: /usr/local/cuda/lib64/libOpenCL.so

To configure the above variables I used "cmake-gui" followed by the "make -j64" command inside the /elastix/build directory. I used the default ITK branch "master" and default elastix branch "develop".

The "make -j64" command failed at multiple tests. But the first occuring error is the following:

```
[ 46%] Linking CXX executable ../bin/itkMevisDicomTiffImageIOTest
In file included from /home/username/elastix/Common/OpenCL/Filters/itkGPUAdvancedCombinationTransformCopier.hxx:39,
                 from /home/username/elastix/Common/OpenCL/Filters/itkGPUAdvancedCombinationTransformCopier.h:259,
                 from /home/username/elastix/Testing/itkGPUResampleImageFilterTest.cxx:25:
/home/username/elastix/Common/OpenCL/Filters/itkGPUAdvancedSimilarity3DTransform.h: In instantiation of ‘const CPUInverseMatrixType& itk::GPUAdvancedSimilarity3DTransform<TScalarType, TParentTransform>::GetCPUInverseMatrix() const [with TScalarType = float; TParentTransform = itk::AdvancedSimilarity3DTransform<float>; itk::GPUAdvancedSimilarity3DTransform<TScalarType, TParentTransform>::CPUInverseMatrixType = itk::Matrix<float, 3, 3>]’:
/home/username/elastix/Common/OpenCL/Filters/itkGPUAdvancedSimilarity3DTransform.h:70:3:   required from here
[...]

PR #653, opened 2022-05-17
COMP: Fix OPENCL errors GetInverseMatrix() and GetOffset() (issue #647)

N-Dekker:
Addressed the compile errors reported by Georg Walther (@georg-walther), issue #647 "[GPU] Compilation fails with ELASTIX_USE_OPENCL: ON", including:

> error: ‘GetInverseMatrix()’ is private within this context
> error: ‘GPUAdvancedEuler3DTransform’ has no member named ‘GetOffset’

`GetInverseMatrix()` was used only to pass the inverse matrix to the `inverse_matrix` data members of `GPUMatrixOffsetTransformBase` (the OpenCL implementation), but these data members were never actually used, so they could simply be removed. All `GetCPUInverseMatrix()` member functions then also appeared unnecessary.

`AdvancedMatrixOffsetTransformBase::GetOffset()` needed to be restored (albeit only `protected`), as it was removed by pull request https://github.com/SuperElastix/elastix/pull/544 commit 15a3c3f07669fa811fc5896b9b01863a33c4b73b "STYLE: Remove unused AdvancedMatrixOffsetTransformBase member functions", October 29, 2021.

With help from Denis P. Shamonin (@dpshamonin).

### DIFF
--- a/Common/OpenCL/Filters/itkGPUAdvancedEuler2DTransform.h
+++ b/Common/OpenCL/Filters/itkGPUAdvancedEuler2DTransform.h
@@ -65,13 +65,6 @@ public:
     return this->GetMatrix();
   }
 
-  /** Get CPU inverse matrix of an MatrixOffsetTransformBase. */
-  const CPUInverseMatrixType &
-  GetCPUInverseMatrix() const override
-  {
-    return this->GetInverseMatrix();
-  }
-
   /** Get CPU offset of an MatrixOffsetTransformBase. */
   const CPUOutputVectorType &
   GetCPUOffset() const override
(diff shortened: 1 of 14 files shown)


dpshamonin:
The GPUMatrixOffsetTransformBase.h was designed based on the ITK [itkMatrixOffsetTransformBase](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.1/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h)
The MatrixOffsetTransformBase class has private members:
```
private:
  MatrixType                m_Matrix{ MatrixType::GetIdentity() };               // Matrix of the transformation
  OutputVectorType          m_Offset{};                                          // Offset of the transformation
  mutable InverseMatrixType m_InverseMatrix{ InverseMatrixType::GetIdentity() }; // Inverse of the matrix
```
Thats why OpenCL structs have the same in GPUMatrixOffsetTransformBase.cl
But indeed was never used. I discussed with @N-Dekker and we are removing it.

PR #655, opened 2022-05-25
Install ElastixTargets.cmake file

blowekamp:
Created variables in ElastixConfig.cmake to specify installation path
and build path uniformly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,9 @@ set( ELASTIX_LIBRARY_DIR "lib"      CACHE STRING
   "Directory for installing library files; path is relative to CMAKE_INSTALL_PREFIX" )
 set( ELASTIX_RUNTIME_DIR "bin"      CACHE STRING
   "Directory for installing runtime files; path is relative to CMAKE_INSTALL_PREFIX" )
+if(NOT ELASTIX_INSTALL_PACKAGE_DIR)
+  set(ELASTIX_INSTALL_PACKAGE_DIR "${ELASTIX_LIBRARY_DIR}/cmake/elastix")
+endif()
 
 #---------------------------------------------------------------------
 # Check if Mevis DicomTiff support is desired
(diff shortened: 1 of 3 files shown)


blowekamp:
The example external project built when I defined ITK_DIR on the cmake configuration line. i.e.: `cmake -G Ninja -DElastix_DIR=/scratch/elx -DITK_DIR=/scratch/itk/ ~/src/elastix/dox/externalproject/`

With this change the find_packge(ITK) has moved into the ElastixConfig.cmake file. So these lines are not longer needed in an external project:
https://github.com/SuperElastix/elastix/blob/02489736455ad9893856ccb376f616b7305d0a6a/dox/externalproject/CMakeLists.txt#L16-L19

The above is require so that the target defined in ElastixTargets ITK dependencies are targets defined in ITKTargets.

The "ELASTIX_ITK_DIR" variable definition was left in the ElastixConfig.cmake file for compatibility, but it is problematic to save the ITK_DIR directory used to build. 

So, this is a change in behavior requiring ITK_DIR to be defined on the command line.

CMakeLists.txt
```diff
@@ -598,15 +601,27 @@ endforeach()
 string(REPLACE "\\" "/" ELASTIX_ITK_DIR ${ITK_DIR})
 
 # Create the Config.cmake file for the build tree:
+set(ElastixConfig_TREE "build")
+set(ElasstixConfig_CODE "")
```
N-Dekker, line 605: Please mind the typo: "ElasstixConfig_CODE" should only have one `s`!

blowekamp: done

blowekamp:
> > this is a change in behavior requiring ITK_DIR to be defined on the command line.
> 
> Thanks for investigating this issue, Bradley. However, I would like an external project that links to the elastix build tree to be able to link to specifically the ITK_DIR that was used by the elastix build. By default, a `find_package(ITK)` in such an external project should find the ITK binaries used by the specific elastix build. Do you think that could still be made possible?

I just updated the PR so that the build config has code to set ITK_DIR.

PR #659, opened 2022-06-01
ENH: Resampler use size of input image when "Size" parameter is missing

N-Dekker:
Assumed that the size of the output image (the number of pixels along each axis) is the same as the input image, when the user does not explicitly specify the "Size" parameter.

Eases manually writing and reusing a transform parameter file. Especially useful for tests and experiments.


N-Dekker:
Rejected, as discussed at the sprint last Monday. Users should still explicitly specify the requested size of the output image, even if it is just the same as the input image.

PR #673, opened 2022-06-28
BUG: Add Direction, Index, Origin, Spacing to transform parameter file

N-Dekker:
To avoid exceptions from `itk::TransformixFilter::GenerateOutputInformation()`.


N-Dekker:
Discussed with @mstaring at our sprint. These parameters should indeed be required for transformix input

PR #681, opened 2022-07-07
ENH: Report any exception from elastix and transformix `main` functions

N-Dekker:
Caught and reported exceptions from the `main` function of elastix and transformix, which would otherwise just cause the program to terminate unexpectedly.

Addressed LLVM 14.03 LLVM clang-tidy check `bugprone-exception-escape` regarding `main`:

> Finds functions which may throw an exception directly or indirectly, but they should not.

https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-exception-escape.html

### DIFF
--- a/Core/CMakeLists.txt
+++ b/Core/CMakeLists.txt
@@ -146,6 +146,8 @@ endif()
 add_executable(elastix_exe
   Main/elastix.cxx
   Main/elastix.h
+  Main/elxMainExeUtilities.cxx
+  Main/elxMainExeUtilities.h
   Kernel/elxElastixMain.cxx
   Kernel/elxElastixMain.h
   ${InstallFilesForExecutables}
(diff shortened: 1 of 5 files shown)


N-Dekker:
Ignore white-space when comparing the changes: https://github.com/SuperElastix/elastix/pull/681/files?w=1

Core/CMakeLists.txt
```diff
@@ -175,6 +177,8 @@ target_link_libraries(elastix_lib ${ELASTIX_TARGET_LINK_LIBRARIES})
 add_executable(transformix_exe
   Main/transformix.cxx
   Main/elastix.h
+  Main/elxMainExeUtilities.cxx
```
N-Dekker, line 180: It is somewhat unfortunate that Main/elxMainExeUtilities.cxx is built twice now: once for elastix_exe and once for transformix_exe.

PR #707, opened 2022-08-19
STYLE: Use `DummyIPPPixelType = unsigned char` everywhere

N-Dekker:
Removed "todo test DummyIPPPixelType=bool", and consistently used `unsigned char` as dummy pixel type for mesh template instantiations (rather than `float` or `double`).

`DummyIPPPixelType = unsigned char` was being used on three places already.


N-Dekker:
Sprint: Use float

PR #714, opened 2022-08-26
ENH: Add `NumberOfThreads` property to itk::TransformixFilter

N-Dekker:
Addressed issue https://github.com/InsightSoftwareConsortium/ITKElastix/issues/163 "No option to set number of threads for Transformix", submitted by Sebastian van der Voort (@svdvoort).


thewtex (approved):
:+1:

N-Dekker:
Sorry, after submitting this pull request, I started doubting. If `TransformixFilter.SetNumberOfThreads` is just equivalent to `itk.MultiThreaderBase.SetGlobalMaximumNumberOfThreads`, I think users should just call the `itk.MultiThreaderBase` member function directly.

N-Dekker:
> There is a per-instance [SetNumberOfWorkUnits](https://itk.org/Doxygen/html/classitk_1_1ProcessObject.html#a07e1d0d435747ae65bfb213bd12053e8) on all `ProcessObject's`. When using ResampleImageFilter directly, for example, this can be set. Ideally, `itk::TransformixFilter` should expose this interface and pass the value to component classes.

@thewtex Thanks for the suggestion Matt. I did give it an initial try before, with https://github.com/SuperElastix/elastix/compare/NumberOfWorkUnits It sets the NumberOfWorkUnits for each elastix component that is derived from itk::ProcessObject. However, I have not finished this approach, because it does not really limit the number of threads used by each individual component.  The NumberOfWorkUnits is not (yet?) passed to each sub-object of each component. I don't know if it's already worth merging this incomplete commit. If users really want to reduce the number of threads, they should just call `itk.MultiThreaderBase.SetGlobalMaximumNumberOfThreads` anyway.

thewtex:
@N-Dekker that's great, it looks like with some more work to propagate NumberOfWorkUnits to the sub-objects developers will be able to request the number of threads used on a per-filter basis.

PR #243, opened 2020-04-26
BUG: Try fixing #70 CL_OUT_OF_RESOURCES exception

squll1peter:
In this PR, I tried to resolved issue #70 by adding GPUImage counterpart of template rebind and adding recursive grafting in OpenCL Pyramids.
Besides, an memory allocation limit is added to GPURecursiveGaussianImageFilter to reduce unnecessary large amount of GPU global memory allocation in the kernel.  
Detailed discussion is in #70.

Issue #70, opened 2018-06-18
OpenCL memory error CL_OUT_OF_RESOURCES


urlicht:
Hello,
I'm having a difficulty running registration with OpenCL, getting this error message CL_OUT_OF_RESOURCES.

The moving and fixed images are about 4MB in size. Strangely, the memory usage maxes at about ~2GB (the GPU on the machine has 8GB of memory, GTX 1080) and the GPU utilization maxes at 5~7%. Even if I decrease the number of resolutions to make the pyramid smaller, I still get the same error message:
`
Description: CL_OUT_OF_RESOURCES
WARNING: Unable to configure the GPU.
 The OpenCLFixedGenericImagePyramid is switching back to CPU mode.

Error: in function: opencl_context_notify
Details: OpenCL error during context creation or runtime:
CL_OUT_OF_RESOURCES error executing CL_COMMAND_WRITE_BUFFER on GeForce GTX 1080 (Device 0).
`

Any insights on the issue?

mstaring:
I have no idea. I think this requires debugging.

chunlc:
Hi ,
I also had difficulty running registration with OpenCl.The errors described by "urlicht" are very similar to those I encountered.

I tried to run the **“elastix.exe”** program ,the moving and fixed images are about 10MB in size.  the memory usage maxes at about 1.2GB (the GPU on the machine has 6GB of memory, GTX 1060) and the GPU utilization is almost 0%.  Registration iteration didn't really start, because the last characters displayed in the console are: _Initialization of all components (before registration) took: 27 ms_.   What's worse is that the “elastix.exe”program can't exit. 

I use ITK 4.13.1 compiled with Windows 64 bit binaries | MSVS 2013 | CMake 3.7.2 :
Module_ITKReview: ON
ITK_USE_64BITS_IDS: ON
ITK_LEGACY_REMOVE: ON
ITK_USE_GPU: ON
I use elastix 4.9.0 compiled with  Windows 64 bit binaries | MSVS 2013 | CMake 3.7.2 :
ELASTIX_USE_OPENCL: ON
USE_OpenCLFixedGenericPyramid: ON
USE_OpenCLMovingGenericPyramid: ON
USE_OpenCLResampler: ON
OPENCL_C_VERSION_1_2: ON
OPENCL_INCLUDE_DIRS: /usr/local/cuda/include
OPENCL_LIBRARIES: /usr/local/cuda/lib64/libOpenCL.so
[...]

PR #734, opened 2022-09-12
BUG: Fix OpenCL pyramids (original: https://github.com/SuperElastix/elastix/pull/243)

ntatsisk:
This PR provides the cosmetic changes needed for PR https://github.com/SuperElastix/elastix/pull/243 together with a small addition. Hence, full credits should be given to @squll1peter for writing up the original PR and resolving the CL_OUT_OF_RESOURCES issue discussed in PR https://github.com/SuperElastix/elastix/pull/243 and issue https://github.com/SuperElastix/elastix/issues/70!  Two different bugs are actually solved. In short summary:

1. The CL_OUT_OF_RESOURCES error. This bug was created because `itk::SmoothingRecursiveGaussianImageFilter` was overriding (here: https://github.com/InsightSoftwareConsortium/ITK/blob/3c6de8f009aadfab8f801070237cb699f5c895cb/Modules/Filtering/Smoothing/include/itkSmoothingRecursiveGaussianImageFilter.h#L80-L87) the internal pixel type of `RecursiveGaussianImageFilter` member objects using the templated struct `Rebind`. However, `Rebind` was only implemented for `itk::Image` and not `itk::GPUImage`, making the latter call the implementation of the former and leading to mixing of the two types. The solution here was to create the equivalent `Rebind` for `itk::GPUImage`

2. A crash was appearing during the start of the second registration resolution because only the first output of the pyramid member object was communicated/grafted to the FixedGenericPyramid / MovingGenericPyramid objects. A simple for-loop to graft all outputs was added as solution.

-------------------------------------------------------------------------------------
In addition to the above bug-fixing changes, the class signature of itkGPUResampleImageFilter was updated to match the one of itkResampleImageFilter where an extra templated parameter was added (`TTransformPrecisionType`). The `itkGPUResampleImageFilterFactory` was updated accordingly.
 
An important note: Several of the related tests were passing successfully before this changes, where clearly they shouldn't. So, some work needs to be done there to make them effective.  

A second note is that memory allocation enhancement, provided in the original PR, is not implemented here, but it should be looked into the future.

### DIFF
--- a/Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
+++ b/Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
@@ -73,72 +73,144 @@ public:
     using GPUOutputImageType = GPUImage<TTypeOut, VImageDimension>;
 
     // Override default
-    this->RegisterOverride(typeid(ResampleImageFilter<InputImageType, OutputImageType, float>).name(),
-                           typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float>).name(),
-                           "GPU ResampleImageFilter override default, interpolator float",
+    this->RegisterOverride(typeid(ResampleImageFilter<InputImageType, OutputImageType, float, float>).name(),
+                           typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float, float>).name(),
+                           "GPU ResampleImageFilter override default, interpolator float and transform float",
                            true,
                            CreateObjectFunction<GPUResampleImageFilter<InputImageType, OutputImageType, float>>::New());
 
     // Override default
-    // With interpolator precision type as double, on GPU as float
-    this->RegisterOverride(typeid(ResampleImageFilter<InputImageType, OutputImageType, double>).name(),
-                           typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float>).name(),
-                           "GPU ResampleImageFilter override default, interpolator double",
-                           true,
-                           CreateObjectFunction<GPUResampleImageFilter<InputImageType, OutputImageType, float>>::New());
+    // With interpolator precision type as double and transform precision type as double, on GPU as float
+    this->RegisterOverride(
+      typeid(ResampleImageFilter<InputImageType, OutputImageType, double, double>).name(),
+      typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override default, interpolator double  and transform double",
+      true,
+      CreateObjectFunction<GPUResampleImageFilter<InputImageType, OutputImageType, float, float>>::New());
+
+    // Override default
+    // With interpolator precision type as float and transform precision type as double, on GPU as float
+    this->RegisterOverride(
+      typeid(ResampleImageFilter<InputImageType, OutputImageType, float, double>).name(),
+      typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override default, interpolator float and transform double",
+      true,
+      CreateObjectFunction<GPUResampleImageFilter<InputImageType, OutputImageType, float, float>>::New());
+
+    // Override default
+    // With interpolator precision type as double and transform precision type as float, on GPU as float
+    this->RegisterOverride(
+      typeid(ResampleImageFilter<InputImageType, OutputImageType, double, float>).name(),
+      typeid(GPUResampleImageFilter<InputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override default, interpolator double and transform float",
+      true,
+      CreateObjectFunction<GPUResampleImageFilter<InputImageType, OutputImageType, float, float>>::New());
 
     // Override when itkGPUImage is first template argument
     this->RegisterOverride(
-      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, float>).name(),
-      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float>).name(),
-      "GPU ResampleImageFilter override GPUImage first, interpolator float",
+      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, float, float>).name(),
+      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage first, interpolator float and transform float",
       true,
-      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float>>::New());
+      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>>::New());
 
     // Override when itkGPUImage is first template argument
-    // With interpolator precision type as double, on GPU as float
+    // With interpolator precision type as double and transform as double, on GPU as float
     this->RegisterOverride(
-      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, double>).name(),
-      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float>).name(),
-      "GPU ResampleImageFilter override GPUImage first, interpolator double",
+      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, double, double>).name(),
+      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage first, interpolator double and transform double",
       true,
-      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float>>::New());
+      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>>::New());
+
+    // Override when itkGPUImage is first template argument
+    // With interpolator precision type as double and transform as float, on GPU as float
+    this->RegisterOverride(
+      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, double, float>).name(),
+      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage first, interpolator double and transform float",
+      true,
+      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>>::New());
+
+    // Override when itkGPUImage is first template argument
+    // With interpolator precision type as float and transform as double, on GPU as float
+    this->RegisterOverride(
+      typeid(ResampleImageFilter<GPUInputImageType, OutputImageType, float, double>).name(),
+      typeid(GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage first, interpolator float and transform double",
+      true,
+      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, OutputImageType, float, float>>::New());
 
     // Override when itkGPUImage is second template argument
     this->RegisterOverride(
-      typeid(ResampleImageFilter<InputImageType, GPUOutputImageType, float>).name(),
-      typeid(GPUResampleImageFilter<InputImageType, GPUOutputImageType, float>).name(),
-      "GPU ResampleImageFilter override GPUImage second, interpolator float",
+      typeid(ResampleImageFilter<InputImageType, GPUOutputImageType, float, float>).name(),
+      typeid(GPUResampleImageFilter<InputImageType, GPUOutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage second, interpolator float and transform float",
       true,
-      CreateObjectFunction<GPUResampleImageFilter<InputImageType, GPUOutputImageType, float>>::New());
+      CreateObjectFunction<GPUResampleImageFilter<InputImageType, GPUOutputImageType, float, float>>::New());
 
     // Override when itkGPUImage is second template argument
-    // With interpolator precision type as double, on GPU as float
+    // With interpolator precision type as double and transform as double, on GPU as float
     this->RegisterOverride(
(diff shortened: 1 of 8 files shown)


mstaring:
Great work!

> A second note is that memory allocation enhancement, provided in the original PR, is not implemented here, but it should be looked into the future.

I think a new test will make this bug visible, when run on a modern GPU.

squll1peter:
This PR & comment summarized the reasoning and code elegantly, I really appreciate it!

Common/OpenCL/Filters/itkGPUResampleImageFilter.hxx
```diff
@@ -1135,10 +1184,11 @@ GPUResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType>::G
  * ***************** PrintSelf ***********************
  */
 
-template <typename TInputImage, typename TOutputImage, typename TInterpolatorPrecisionType>
+template <typename TInputImage, typename TOutputImage, typename TInterpolatorPrecisionType, TTransformPrecisionType>
```
N-Dekker, line 1187: Missing typename keyword! Visual C++ says:
> itkGPUResampleImageFilter.hxx(1187,116): error C2061: syntax error: identifier 'TTransformPrecisionType'

Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
```diff
@@ -73,89 +73,164 @@ class ITK_TEMPLATE_EXPORT GPUResampleImageFilterFactory2 : public GPUObjectFacto
...
-      typeid(ResampleImageFilter<GPUInputImageType, GPUOutputImageType, double>).name(),
-      typeid(GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float>).name(),
-      "GPU ResampleImageFilter override GPUImage first and second, interpolator double",
+      typeid(ResampleImageFilter<GPUInputImageType, GPUOutputImageType, double, double>).name(),
+      typeid(GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float, float>).name(),
+      "GPU ResampleImageFilter override GPUImage first and second, interpolator double and transform double",
       true,
-      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float>>::New());
+      CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float, float>>::New());
   }
```
N-Dekker, line 195: This `}` should be moved down, beyond the last `this->RegisterOverride` call!

Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
```diff
@@ -73,89 +73,164 @@ class ITK_TEMPLATE_EXPORT GPUResampleImageFilterFactory2 : public GPUObjectFacto
...
+
+// Override when itkGPUImage is first and second template arguments
+// With interpolator precision type as double and transform float, on GPU as float
+this->RegisterOverride(
+  typeid(ResampleImageFilter<GPUInputImageType, GPUOutputImageType, double, float>).name(),
+  typeid(GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float, float>).name(),
+  "GPU ResampleImageFilter override GPUImage first and second, interpolator double and transform float",
+  true,
+  CreateObjectFunction<GPUResampleImageFilter<GPUInputImageType, GPUOutputImageType, float, float>>::New());
+} // namespace itk
```
N-Dekker, line 215: The two closing curly braces `}` should be removed from here!

N-Dekker (changes requested):
Obviously a few syntax errors still need to be fixed. Otherwise, very promising!

Common/OpenCL/Filters/itkGPUResampleImageFilter.hxx
N-Dekker: @ntatsisk  I see, you fixed it, by https://github.com/SuperElastix/elastix/compare/86468c87bef8c29cf7236159cff91d232fb87680..5901d51a30c5a67604a5081287a282e2075c9eff  👍

Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
N-Dekker: @ntatsisk  I see, you fixed those `}` as well, by https://github.com/SuperElastix/elastix/compare/86468c87bef8c29cf7236159cff91d232fb87680..5901d51a30c5a67604a5081287a282e2075c9eff  👍

Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h
N-Dekker: I see, fixed 👍

ntatsisk:
I added some changes to the pyramid tests because I noticed they were all passing before this PR, which clearly showed that the bug/crash in the code was not really covered in the tests. 

The relevant test here is the `GPUGenericMultiResolutionPyramid` test. The initial reason that it was passing was that in fact the CPU version of the Resampler was being called instead of the GPU, which indicates that the corresponding Factory+overloaded function (this one: https://github.com/SuperElastix/elastix/blob/fdd71315d26ef61b814f77ec939c0e10951e537b/Common/OpenCL/Factories/itkGPUResampleImageFilterFactory.h#L84-L88) was not doing its job. Looking into it further, it appears that `PrecisionType` was not specified for the `GenericMultiResolutionPyramid` inside the test, which in turn meant that default `double` was  used. Still, the corresponding Factory should have worked, but it seems that it doesn't, which is a bug. Here, I just added the `PrecisionType=float` in order to call a different overloaded function of the Factory, which in fact works (the same used inside the code). So, in this way we avoid using the problematic factory but still the bug is there (maybe it is in the ITK side, since all the factor magic happens inside the `itkObjectFactoryBase.h` file).

As expected, this change allowed the GPUResampler to get called as it should. However, the test was still passing successfully! Reminder: the equivalent part of the code crashes. Why? I don't understand fully all the details, but the high-level difference is that inside the code we call `GenericMultiResolutionPyramidImageFilter<GPUInputImageType, GPUOutputImageType, PrecisionType>` while inside the test: `GenericMultiResolutionPyramidImageFilter<InputImageType, OutputImageType, PrecisionType>`. So, I added an extra test to cover also for the case of calling the pyramid filter with `itk::GPUImage`s instead of `itk::Image`s. This new test fails as expected before the changes introduced here, and passes after the changes in this PR.

ntatsisk:
To explain a bit better the problem with the factory registration, I created a minimal example: 

```
#include "itkProcessObject.h"
#include "itkObjectFactoryBase.h"
#include "itkImage.h"

namespace itk
{

// MyFilter
template <typename TInputImage, typename TParameter=double>
class MyFilter : public ProcessObject 
{
public:
  using Self = MyFilter;
  using Superclass = ProcessObject;
  using Pointer = SmartPointer<Self>;
  using ConstPointer = SmartPointer<const Self>;
		  
  itkNewMacro(Self);
  itkTypeMacro(MyFilter, ProcessObject);
  
}; // end MyFilter

// SubFilter
template <typename TInputImage, typename TParameter=double>
class SubFilter : public MyFilter<TInputImage, TParameter> 
{
public:
  using Self = SubFilter;
  using Superclass = MyFilter;
  using Pointer = SmartPointer<Self>;
  using ConstPointer = SmartPointer<const Self>;
		  
  itkNewMacro(Self);
  itkTypeMacro(SubFilter, MyFilter);
  
}; // end SubFilter

// SubFilterFactory
class SubFilterFactory : public itk::ObjectFactoryBase
{
public:
  using Self = SubFilterFactory;
  using Superclass = itk::ObjectFactoryBase;
  using Pointer = itk::SmartPointer<Self>;
  using ConstPointer = itk::SmartPointer<const Self>;

  itkFactorylessNewMacro(Self);
  itkTypeMacro(SubFilterFactory, itk::ObjectFactoryBase);

  static void
  RegisterOneFactory()
  {
    auto factory = SubFilterFactory::New();

    itk::ObjectFactoryBase::RegisterFactory(factory);
  }
  
  const char *
  GetITKSourceVersion() const override
  {
    return "itk version 5.3.0";
  }
  const char *
  GetDescription() const override
  {
    return "A Factory";
  }
  
private:
  SubFilterFactory()
  {
    using InputImage = Image<float, 3>;

    // double -> float
    this->RegisterOverride(typeid(MyFilter<InputImage, double>).name(),
							typeid(SubFilter<InputImage, float>).name(),
							" ",
							true,
							CreateObjectFunction<SubFilter<InputImage, float>>::New());

    // float -> float
    this->RegisterOverride(typeid(MyFilter<InputImage, float>).name(),
							typeid(SubFilter<InputImage, float>).name(),
							" ",
							true,
							CreateObjectFunction<SubFilter<InputImage, float>>::New());

  }	    
}; // end SubFilterFactory

} // end itk

int
main(int argc, char * argv[])
{
  using InputImage = itk::Image<float, 3>;
  using MyFilterFloatType = itk::MyFilter<InputImage, float>;
  using MyFilterDoubleType = itk::MyFilter<InputImage, double>;

  // Instantiate MyFilter class as usual
  auto myFilterFloat = MyFilterFloatType::New();
  auto myFilterDouble = MyFilterDoubleType::New();

  std::cout << "Before factory registration:" << std::endl;
  std::cout << typeid(*myFilterFloat).name() << std::endl;
  std::cout << typeid(*myFilterDouble).name() << "\n" << std::endl;

  // Register factory
  itk::SubFilterFactory::RegisterOneFactory();
  
  // From now on, only SubFilters should be instantiated
  auto subFilterFloat = MyFilterFloatType::New();
  auto subFilterDouble = MyFilterDoubleType::New();

  std::cout << "After factory registration:" << std::endl;
  std::cout << typeid(*subFilterFloat).name() << std::endl;
  std::cout << typeid(*subFilterDouble).name() << std::endl;


} // end main()
```

The output is: 
```
Before factory registration:
class itk::MyFilter<class itk::Image<float,3>,float>
class itk::MyFilter<class itk::Image<float,3>,double>

After factory registration:
class itk::SubFilter<class itk::Image<float,3>,float>
class itk::MyFilter<class itk::Image<float,3>,double>
```

As you can see the conversion from double->float fails, while the float->float succeeds. @N-Dekker, what are your thoughts? How should we move forward?

N-Dekker:
I see, the dynamic_cast from `SubFilter<itk::Image<float,3>,float>*` to `MyFilter<itk::Image<float,3>,double>>*` fails, in `ObjectFactory<itk::MyFilter<itk::Image<float,3>,double>>::Create()`.

ntatsisk:
After some discussions with @N-Dekker, we agreed to have simply one test for `GPUGenericMultiResolutionPyramid` instead of two i.e. to keep the one which covers how the filter is instantiated in the actual elastix code. In addition, some duplicate code was simplified in that test. Lastly, based on the discussion regarding the factories above , I added a commit that removes all GPUResampler factories that were not functional (=they were attempting without success to override double with float) and, because of that, the GPUFactories test had to be updated as well.

If there are no further comments/questions/discussion points, I think the PR can be merged. :)

Components/FixedImagePyramids/OpenCLFixedGenericPyramid/elxOpenCLFixedGenericPyramid.hxx
```diff
@@ -196,8 +196,12 @@ OpenCLFixedGenericPyramid<TElastix>::GenerateData()
 
   if (computedUsingOpenCL)
   {
-    // Graft output
-    this->GraftOutput(this->m_GPUPyramid->GetOutput());
+    // Graft outputs
+    const auto numberOfLevels = this->GetNumberOfLevels();
+    for (unsigned int i = 0; i < numberOfLevels; i++)
```
N-Dekker, line 201: FYI: No show-stopper, but in general, in C++ code, `++i` is preferred to `i++`, see pull request #467. Just a minor style issue in this particular case. No action needed.

Testing/itkGPUGenericMultiResolutionPyramidImageFilterTest.cxx
```diff
@@ -45,6 +45,29 @@
 // The filter takes an input image and produces an output image.
 // We compare the CPU and GPU output image write RMSE and speed.
 
+template <typename FilterType>
+void
+UpdateFilterNTimes(typename FilterType::Pointer filter, const unsigned int N, const bool computeOnlyForCurrentLevel)
```
N-Dekker, line 50: Really nice! Not just because your helper function reduces the amount of duplicate code, but also because it makes it very clear that the test does exactly the same thing for the GPU filter as for the CPU filter!

Minor comments (no show stopper):
 - Our (ITK based) naming conversion is to have a `T` as first letter of a template type parameter: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/51b0eb02d945ab7f5f24f16139cdf00fd0c783a9/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1083 So in this case it could be `TFilter`.
 - Personally I would rather pass a function argument _by reference_ than by (smart or raw) pointer. In this case, I would declare the filter parameter something like `TFilter& filter`. However, I haven't been able not convince ITK developers in a similar case: https://github.com/InsightSoftwareConsortium/ITK/pull/2102#discussion_r519021533 

No action needed! 👍

Common/OpenCL/ITKimprovements/itkGPUImage.h
```diff
@@ -96,6 +96,17 @@ class ITK_TEMPLATE_EXPORT ITKOpenCL_EXPORT GPUImage : public Image<TPixel, VImag
...
 
+  /** Override Rebind and RebindImageType of itk::Image class */
+  template <typename UPixelType, unsigned int VUImageDimension = VImageDimension>
+  struct Rebind
+  {
+    using Type = itk::GPUImage<UPixelType, VUImageDimension>;
+  };
+
+  template <typename UPixelType, unsigned int VUImageDimension = VImageDimension>
+  using RebindImageType = itk::GPUImage<UPixelType, VUImageDimension>;
```
N-Dekker, line 107: Note to ourselves: remember to consider to propose a similar specialization for ITK's GPU image!

N-Dekker (commented):
Yes, I think the PR can be merged as well :)  Thanks @ntatsisk !

N-Dekker (approved):

PR #737, opened 2022-09-28
ENH: Add warnings on schedule adjustments in pyramids

N-Dekker:
Pair-programmed w/ Konstantinos (@ntatsisk).

### DIFF
--- a/Components/elxGenericPyramidHelper.h
+++ b/Components/elxGenericPyramidHelper.h
@@ -109,6 +109,16 @@ public:
     {
       /** Set the rescale schedule into this class. */
       pyramid.SetRescaleSchedule(rescaleSchedule);
+
+      const auto newSchedule = pyramid.GetRescaleSchedule();
+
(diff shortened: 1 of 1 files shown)


N-Dekker:
For the record, elastix `GenericMultiResolutionPyramidImageFilter::SetSmoothingSchedule` clamps between zero and the previous value, at:

https://github.com/SuperElastix/elastix/blob/86e3d2cfdc5158b1ccf5f2c876acb36db75ef020/Common/itkGenericMultiResolutionPyramidImageFilter.hxx#L220-L233

On the other hand, ITK's `MultiResolutionPyramidImageFilter::SetSchedule` clamps between _one_ and the previous value:

https://github.com/InsightSoftwareConsortium/ITK/blob/e2b9068d8aa525b7bca410b717e2ef8cb2290763/Modules/Registration/Common/include/itkMultiResolutionPyramidImageFilter.hxx#L178-L190

PR #747, opened 2022-10-21
Add `TransformixFilter::SetTransform`

N-Dekker:
Added support for single and composite transforms (itk::CompositeTransform). Tested with itk::TranslationTransform, itk::AffineTransform, itk::Euler2DTransform, itk::Euler3DTransform, itk::Similarity2DTransform, itk::Similarity3DTransform, and itk::BSplineTransform.

### DIFF
--- a/Core/Main/itkTransformixFilter.h
+++ b/Core/Main/itkTransformixFilter.h
@@ -37,6 +37,7 @@
 
 #include "itkImageSource.h"
 #include "itkMesh.h"
+#include "itkTransformBase.h"
 
 #include "elxTransformixMain.h"
 #include "elxParameterObject.h"
@@ -220,6 +221,11 @@ public:
     return m_OutputMesh;
   }
 
+  /** Sets the transformation. If null, the transformation is entirely specified by the transform
+   * parameter object that is set by SetTransformParameterObject. Otherwise, the transformation is specified by this
+   * transform object, with additional information from the specified transform parameter object. */
+  itkSetConstObjectMacro(Transform, TransformBase);
+
 protected:
   TransformixFilter();
 
@@ -264,6 +270,8 @@ private:
 
   typename MeshType::ConstPointer m_InputMesh{ nullptr };
   typename MeshType::Pointer      m_OutputMesh{ nullptr };
+
+  TransformBase::ConstPointer m_Transform;
 };
 
 } // namespace itk
--- a/Core/Main/itkTransformixFilter.hxx
+++ b/Core/Main/itkTransformixFilter.hxx
@@ -39,7 +39,11 @@
 #include "itkTransformixFilter.h"
 #include "elxPixelTypeToString.h"
 #include "elxTransformBase.h"
+#include "elxTransformIO.h"
 #include "elxDefaultConstruct.h"
+
+#include <itkCompositeTransform.h>
+
 #include <memory> // For unique_ptr.
 
 namespace itk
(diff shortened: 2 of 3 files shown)


N-Dekker:
The most important part of the implementation is in `itk::TransformixFilter<TMovingImage>::GenerateData()`, in itkTransformixFilter.hxx:
https://github.com/SuperElastix/elastix/pull/747/files?file-filters%5B%5D=.hxx&show-viewed-files=true

Core/Main/itkTransformixFilter.h
```diff
@@ -220,6 +221,19 @@ class ITK_TEMPLATE_EXPORT TransformixFilter : public ImageSource<TMovingImage>
     return m_OutputMesh;
   }
 
+  /** Sets the transformation. If null, the transformation is entirely specified by the transform
+   * parameter object that is set by SetTransformParameterObject. Otherwise, the transformation is specified by this
+   * transform object, with additional information from the specified transform parameter object. */
```
mstaring, line 226: Is this implementation not the same as the itkSetConstObjectMacro ?

N-Dekker: @mstaring Welll... it's similar, looking at https://github.com/InsightSoftwareConsortium/ITK/blob/49150c8a701f629a122646f708a802f961d1ace7/Modules/Core/Common/include/itkMacro.h#L1185-L1195
```
#define itkSetConstObjectMacro(name, type)             \
  virtual void Set##name(const type * _arg)            \
  {                                                    \
    itkDebugMacro("setting " << #name " to " << _arg); \
    if (this->m_##name != _arg)                        \
    {                                                  \
      this->m_##name = _arg;                           \
      this->Modified();                                \
    }                                                  \
  }                                                    \
```

Three differences:
- TransformixFilter::SetTransform is non-virtual. I don't think it should be overridden, otherwise it just might not work anymore. My proposal to remove the virtual keywords from those ITK Get/Set macro's is postponed until ITK 6.0.0: https://github.com/InsightSoftwareConsortium/ITK/pull/2785
- The parameter of TransformixFilter::SetTransform is a SmartPointer, indicating that the transform object will be "shared" between the caller and TransformixFilter. I think that makes sense.
- TransformixFilter::SetTransform does not call `itkDebugMacro`. Should it do so?

Anyway, if you prefer `itkSetConstObjectMacro`, I can adjust the code, no problem.

N-Dekker: OK, now using the itkSetConstObjectMacro, although I'm not very happy about the current implementation of the macro.

Core/Main/itkTransformixFilter.hxx
```diff
@@ -190,6 +196,77 @@ TransformixFilter<TMovingImage>::GenerateData()
...
+          transformParameterMapVector.resize(numberOfTransforms);
+        }
+        else
+        {
+          // The last TransformParameterMap is special, as it needs to be used for the final transformation.
+          const auto lastTransformParameterMap = transformParameterMapVector.back();
+          transformParameterMapVector.resize(numberOfTransforms);
+          transformParameterMapVector.back() = lastTransformParameterMap;
+        }
+      }
```
N-Dekker, line 247: @mstaring @stefanklein Please check if you like this logic:
 - The user-specified compositeTransform mus have _at least_ one subtransform (otherwise throw an exception)
 - If the number of parameter maps is zero, resize the vecor of maps to the number of subtransforms.
 - Otherwise, if the number of subtransforms of the compositeTransform is different from the number of parameter maps, add maps to the end, or remove maps from the end, to make these numbers the same, and then, copy the original "last parameter map" to the last entry of the vector of maps (the "back").

Is that the intended behavior?

Core/Main/itkTransformixFilter.hxx
```diff
@@ -190,6 +196,77 @@ TransformixFilter<TMovingImage>::GenerateData()
...
+        transformToMap(*transform, transformParameterMap);
+      }
+    }
+    else
+    {
+      for (auto & transformParameterMap : transformParameterMapVector)
+      {
+        // Use the same transform for all parameter maps.
+        transformToMap(*m_Transform, transformParameterMap);
+      }
```
N-Dekker, line 266: @mstaring @stefanklein Also please check if you like this: When the specified transform is a single transform (not a composite), its Parameters/FixedParameters/etc are copied to each map. Also if there are multiple maps in the vector. Is that OK?

Should there be a check if there is at least one map? Or if there is _exactly_ one, when the transform is also a single one.

N-Dekker: Discussed with Marius (@mstaring) at the office: when there is only one single transform, there should be only one parameter map. If the specified `TransformParameterObject` has more than one map, only the last map should be taken into account.

Core/Main/itkTransformixFilter.hxx
N-Dekker: Discussed with Marius (@mstaring) at the office: If the number of parameter maps is zero, an exception should be thrown. For the other cases, the behavior was described as intended.

N-Dekker:
@thewtex FYI: `TransformixFilter::SetTransform(const itk::TransformBase*)` is underway! 🎉

PR #767, opened 2022-12-04
ElastixRegistrationMethod transform objects

N-Dekker:
Added public functions to retrieve the transformation(s) produced by elastix through the `itk::ElastixRegistrationMethod` interface:

 - ElastixRegistrationMethod::GetCombinationTransform()
 - TransformixFilter::SetCombinationTransform(transform)
 -  ~~ElastixRegistrationMethod::GetCompositeTransform()~~
 - ElastixRegistrationMethod::GetNumberOfTransforms()
 - ElastixRegistrationMethod::GetNthTransform(n)
 - ElastixRegistrationMethod::ConvertToItkTransform(transform)
 
The transform return by `ConvertToItkTransform(transform)` is a standard ITK transform. Note that this function may throw an exception, if there is no corresponding ITK transform for the actual elastix transform.

The transforms returned by `GetCombinationTransform()` and `GetNthTransform(n)` are specific to elastix. However, the elastix combination transform may be used as argument to `TransformixFilter.SetCombinationTransform(transform)`, and the _nth_ transform may be used as argument to `ConvertToItkTransform(transform)`..

### DIFF
--- a/Core/Main/itkTransformixFilter.h
+++ b/Core/Main/itkTransformixFilter.h
@@ -103,6 +103,8 @@ public:
 
   using MeshType = Mesh<OutputImagePixelType, MovingImageDimension>;
 
+  using TransformType = Transform<double, MovingImageDimension, MovingImageDimension>;
+
   /** Typedefs for images of determinants of spatial Jacobian matrices, and images of spatial Jacobian matrices */
   using SpatialJacobianDeterminantImageType = itk::Image<float, MovingImageDimension>;
   using SpatialJacobianMatrixImageType =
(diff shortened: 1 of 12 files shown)


Core/Main/itkTransformixFilter.h
```diff
@@ -103,6 +103,8 @@ class ITK_TEMPLATE_EXPORT TransformixFilter : public ImageSource<TMovingImage>
 
   using MeshType = Mesh<OutputImagePixelType, MovingImageDimension>;
 
+  using TransformType = Transform<double, MovingImageDimension, MovingImageDimension>;
```
N-Dekker, line 106: Is it OK for the CombinationTransform to assume that FixedImageDimension == MovingImageDimension?

N-Dekker:
For now, remove GetCompositeTransform()

PR #770, opened 2022-12-08
Artifact no install

thewtex:


thewtex:
Re:

- https://github.com/InsightSoftwareConsortium/ITKElastix/pull/186
- https://github.com/InsightSoftwareConsortium/ITKElastix/issues/182
- https://github.com/InsightSoftwareConsortium/ITKElastix/issues/183

This reduces the wheel size from 95 MB to 18 MB.

N-Dekker:
@thewtex The commit is a bit overwhelming to me. Do I understand correctly that it supports `2^3` = eight combinations of those flags, (`ELASTIX_NO_INSTALL_EXECUTABLES, ELASTIX_NO_INSTALL_RUNTIME_LIBRARIES ELASTIX_NO_INSTALL_DEVELOPMENT`)? I assume that only a subset of those eight combinations is of interest, right? Which particular combinations would that be, for which particular use cases?

How many use cases really need to be supported?

thewtex:
@N-Dekker yes, this make corrections and additions to the installation behavior. There are some subsets that I need now, some subsets that I will need later, and there are other subsets that other folks may need. In any case, the categorization of the subsets follows a logical and useful organization of executables, runtime libraries, and development artifacts.

Yes, the default behavior is unchanged.

Please let me know if you have any questions.

N-Dekker:
Thanks for explaining, Matt. Another question I posted here already: https://github.com/InsightSoftwareConsortium/ITKElastix/issues/183#issuecomment-1342981340  Reverting the logic would avoid double negatives (like `if(NOT NO_INSTALL)`) , which could improve code readability, in my opinion. What do you think?

N-Dekker:
@thewtex  Thanks very much, Matt!

Issue #672, opened 2022-06-27
Feature request: specify output name for file in transformix/elastix


Svdvoort:
Currently, when running elastix/transformix it is only possible to specify the output directory, and not specifically the output name. 
For example, from elastix the output files will always be named `TransformParameters.0.txt` and from transformix `result.nii.gz`. 

Especially for transformix it would be nice to be able to specify the output file name. 
I often have the situation where I need to apply the same transformation to multiple images (for example different MR modalities). In this case, I either have to create a new directory per image just to save the transformed image or have to write a wrapper around transformix to rename the output file before a new run. 

Could a flag be added to elastix/transformix to specify the output filename directly? 

For example: 

`transformix -tp TransformParameters.0.txt -out /home/user/ -filename T1 original_scan.nii.gz` 

If the `filename` flag is not provided, elastix/transformix can use the current default behavior (in this case, writing the file as `result.nii.gz`), not breaking any existing functionality.

N-Dekker:
Interesting suggestion, thanks!

Note that in Python you can already achieve this, when using ITKElastix https://github.com/InsightSoftwareConsortium/ITKElastix

N-Dekker:
As far as I can see, the default is "output-directory/result.mhd", looking at:

https://github.com/SuperElastix/elastix/blob/0a1e90f6222c3bb7ba4891dc8bfc6c8774a1fc7c/Core/Kernel/elxElastixTemplate.hxx#L385-L389

I guess the proposal is to allow specifying a path relative to the output directory, right? Or would it be relative to the current working directory?

Of course, it should then also allow specifying an absolute path, either starting with a (back) slash, or a drive letter, followed by colon and (back) slash. Right?

But then, would it entirely ignore the "ResultImageFormat" parameter from the parameter file?

If the directory of the specified file path would be non-existing, would it be OK for transformix to just fail? Or should it try to create the directory?

Svdvoort:
In my opinion it can be much simpler than this. Although specifying additional paths would be nice, in principle I think it's fine to store everything just in the output directory. 
However, the file will always be called result. <ResultImageFormat>, and it would be nice to change the "result" part since the image format can already be changed. It could be a flag, but also just a parameter in the parameter file, just like the ResultImageFormat is. 

That way instead of always saving a file as result.<ResultImageFormat> it can be something like <ResultImageName>.<ResultImageFormat>

This allows for a bit more flexibility and clarity when saving transformed images.

[...]

PR #776, opened 2022-12-21
ENH: Support "ResultImageName" in parameter files

N-Dekker:
Allows the user to specify the name of the result file, that would otherwise just be named something like "result.mhd".

This improvement applies mainly to the elastix and transformix *executables*. Moreover, for the elastix library, it also applies to the result image files generated with "WriteResultImageAfterEachResolution" and "WriteResultImageAfterEachIteration".

Addressing issue #672 "Feature request: specify output name for file in transformix/elastix" by Sebastian van der Voort (@Svdvoort).


N-Dekker:
Tested locally, works for me 👍 

I guess it should then also be supported by the elastix executable (when "WriteResultImage" is true), right?

mstaring:
LGTM, and indeed this should then automatically also work for the executable.

N-Dekker:
> LGTM, and indeed this should then automatically also work for the executable.

Thanks @mstaring I'll replace the other occurrences of "result." with ResultImageName as well.

PR #788, opened 2023-01-19
STYLE: Move PrintSettingsVector and Settings types to `OptimizerBase`

N-Dekker:
Removed the duplicated `PrintSettingsVector` implementations from six optimizers:

    AdaGrad
    AdaptiveStochasticGradientDescent
    AdaptiveStochasticLBFGS
    AdaptiveStochasticVarianceReducedGradient
    PreconditionedGradientDescent
    PreconditionedStochasticGradientDescent


N-Dekker:
@mstaring It appears that six optimizers have exactly the same `SettingsType` struct (containing `a, A, alpha, fmax, fmin, omega`), and exactly the same `PrintSettingsVector` member function. I guess we may move the struct and the member function to the OptimizerBase, to reduce duplicate code, OK?

Having only one `PrintSettingsVector` implementation (instead of six) would ease the planned logging task.

PR #799, opened 2023-02-09
Replace our xout library with Gabi Melman's spdlog

N-Dekker:
[spdlog](https://github.com/gabime/spdlog) is especially interesting to elastix because it offers thread support, it allows logging level selection (to be used later), and it is actively maintained.


mstaring:
beautiful work Niels! Perhaps you can show us some logger output, comparing old with new.

N-Dekker:
Thanks @mstaring With this pull request, the logger output should be just the same as with the old logger (but with added thread-safety, and prepared for future extensions, like log level support). But it's a good idea to double-check that all the output is still there!

My idea is to further adjust the logging in follow-up pull requests, because this pull quest is already quite large!

mstaring:
great to hear, if you can show me two log files, then we can move ahead as you suggest with follow-up PRs.

N-Dekker:
The result of the logging is _partially_ online, already  😃 That is: the output to standard I/O is shown by the GitHub CI. I did some comparison between the CI logs at https://github.com/SuperElastix/elastix/actions/workflows/ElastixGitHubActions.yml and they look comparable (comparing the `Add-spdlog-as-submodule` branch to the `main` branch). 

I also compared the "out" directories of a simple registration, after having run both elastix 5.1.0 (release version) and a local build. They look similar (except for changes in the line breaks, from "\r\n" to "\n", the subject of my PR https://github.com/gabime/spdlog/pull/2636), including the "IterationInfo" files. I can show you, of course.

N-Dekker:
[elx-5.1.0-release-out.zip](https://github.com/SuperElastix/elastix/files/10706452/elx-5.1.0-release-out.zip)
[elx-Debug-spdlog-out.zip](https://github.com/SuperElastix/elastix/files/10706453/elx-Debug-spdlog-out.zip)

mstaring:
looks great Niels, go ahead!

N-Dekker:
The pull request now has 14 commits, but I think I'll squash a few of them. I think I would like to just have separate commits for 1. adding the spdlog library, 2. adding the elastix log class, 3. using the new logging, and 4. removing the xout library.

PR #804, opened 2023-02-17
Add "-loglevel" command-line option to elastix and transformix exe

N-Dekker:
@mstaring @stefanklein @ntatsisk Adding log level to the ITKElastix/library interface is still on my TODO list. But the command-line option is here already  😃

### DIFF
--- a/Core/Kernel/elxlog.cxx
+++ b/Core/Kernel/elxlog.cxx
@@ -103,8 +103,41 @@ get_string_from_stream(const std::ostream & stream)
 }
 
 
+spdlog::level::level_enum
+to_spdlog_level_enum(const log::level log_level)
+{
+  switch (log_level)
+  {
+    case log::level::info:
+    {
+      return spdlog::level::info;
+    }
+    case log::level::warn:
+    {
+      return spdlog::level::warn;
+    }
+    case log::level::err:
+    {
+      return spdlog::level::err;
+    }
+    case log::level::off:
+    {
+      return spdlog::level::off;
+    }
+    default:
+    {
+      assert(!"Unsupported log level!");
+      return spdlog::level::level_enum{};
+    }
+  }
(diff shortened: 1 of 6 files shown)


mstaring:
Nice!
To discuss if we want a commandline option or a parameter in the parameter file.

N-Dekker:
> To discuss if we want a commandline option or a parameter in the parameter file.

Thanks for your comment @mstaring Obviously I like a commandline option best for the log level. But more general, can you or @stefanklein explain again the original elastix design philosophy behind commandline options versus parameter file parameters? When _in general_ is a commandline option the most appropriate design choice?

Current elastix commandline options:

    Mandatory:
    -f        fixed image    
    -m        moving image    
    -out      output directory    
    -p        parameter file, elastix handles 1 or more "-p"    

    Optional:
    -fMask    mask for fixed image    
    -mMask    mask for moving image    
    -fp       point set for fixed image    
    -mp       point set for moving image    
    -t0       parameter file for initial transform    
    -priority set the process priority to high, abovenormal, normal, etc.
    -threads  set the maximum number of threads of elastix

stefanklein:
Hi Niels,
Commandline is the most appropriate choice when it's an option that often changes, e.g. inputs that you loop over (images) or execution settings (priority, nr of threads).
Anything else should go into parameter file.
Log level is a bit inbetween. It could be suitable for the command-line, since during debugging you would often change it and it's an option that should be immediately accessible. Another reason for command-line here is that you also want to make sure the logging level is set for messages that relate to errors in reading the parameters file.
Best, Stefan

N-Dekker:
Discussed at today's internal weekly elastix sprint. We agreed to have this as a command-line option, but _in general_ we should be careful not to add too many command-line options. That's exactly why elastix has parameter files in the first place!

PR #807, opened 2023-02-20
STYLE: Move `InitialTransform` from MainBase to ElastixMain

N-Dekker:
Made clearer that TransformixMain does not need to use the `InitialTransform` property of ElastixMain. `MainBase` is the common base class of TransformixMain and ElastixMain (from pull request https://github.com/SuperElastix/elastix/pull/806 commit c2a5de712dd14ade6f36cc45b443e8f7113b501d). So it appears preferable to move InitialTransform from MainBase to the elastix specific class ElastixMain.

Removed the `SetInitialTransform` call from TransformixMain, as it does not seem to make sense. Discussed with Marius (@mstaring) at our weekly internal elastix sprint.

### DIFF
--- a/Core/Kernel/elxElastixMain.h
+++ b/Core/Kernel/elxElastixMain.h
@@ -105,6 +105,13 @@ public:
    */
   itkGetModifiableObjectMacro(FinalTransform, itk::Object);
 
+  /** Set/Get the initial transform
+   * the type is itk::Object, but the pointer should actually point
+   * to an itk::Transform type (or inherited from that one).
+   */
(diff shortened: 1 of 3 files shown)


N-Dekker:
@stefanklein FYI, I have just added the class `elastix::MainBase` as a common (abstract) base class of ElastixMain and TransformixMain, to make it clearer which functionality elastix and transformix have in common, and which functionality is specific to either one of them: pull request #806 Marius and I concluded during the sprint that `InitialTransform` is specific to elastix, not transformix.

PR #809, opened 2023-02-20
ENH: Only write parameter file to log file when level = `info` (default)

N-Dekker:
Replaced `log::to_log_file` with `log::info_to_log_file`, which _only_ writes to file when the log level is `info` (which is the default).

ElastixTemplate::CreateTransformParameterFile and Configuration::PrintParameterFile() no longer write to file when the log level is `warning`, `error`, or `off`.

### DIFF
--- a/Core/Configuration/elxConfiguration.cxx
+++ b/Core/Configuration/elxConfiguration.cxx
@@ -90,12 +90,12 @@ Configuration::PrintParameterFile() const
   std::string params = this->m_ParameterFileParser->ReturnParameterFileAsString();
 
   /** Separate clearly in log-file, before and after writing the parameter file. */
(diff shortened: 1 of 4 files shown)


N-Dekker:
@mstaring @stefanklein Did I understand correctly that when the user requested log level is `error` (errors only), `warning` (errors + warnings), or `off`, writing the content of the (transform) parameter file to the log file should be suppressed?

I mean those sections:

    =============== start of ParameterFile: file.txt  ===============

...

    =============== end of ParameterFile: file.txt  ===============


Those are only "for information", so they should not be written when the user only wants to see warnings or errors, right?

mstaring:
yes for the elastix.log file this is just info indeed. The TransformParameter files should still be written. (there is a separate parameter to disable that)

PR #812, opened 2023-02-21
BUG: Fix dark color scheme appearance of elastix.lumc.nl/doxygen

N-Dekker:
When the user has selected the dark color scheme in their web browser appearance settings, many pages at https://elastix.lumc.nl/doxygen appear unreadable, as both background and foreground appear dark. A simple quick-fix is to just bypass the elastix style when the color scheme is not light.

Addresses the following comment from https://github.com/InsightSoftwareConsortium/ITKElastix/issues/158#issuecomment-1431146504, by Matt McCormick (@thewtex): 

> the Doxygen CSS seems to not render well on my system:

image

This new revision is already uploaded to https://elastix.lumc.nl/doxygen/elastix.css 

With help from Baldur Lew at LKEB.


N-Dekker:
With the new revision (as proposed here) https://elastix.lumc.nl/doxygen/transformparameter.html appears as follows with dark color scheme: 
ElastixNewDarkScheme

And it still appears as follows with light color scheme:

ElastixLightScheme

PR #813, opened 2023-02-23
PERF: Simplify TransformPoint member functions of BSpline Transform class templates

N-Dekker:
Removed unnecessary local variables, `numberOfWeights`, `indicesArray`, `indices`, `transformedPoint`, `outputPoint`, and `basePointer` from both `AdvancedBSplineDeformableTransform` and `CyclicBSplineDeformableTransform`.

A small but significant improvement of the runtime performance (more than 6%) was observed, from ~3.0 sec. (before this commit) down to ~2.8 sec. (with this commit), when calling `AdvancedBSplineDeformableTransform::TransformPoint` 10'000'000 times.

Also did some code clean-up  within `RecursiveBSplineTransform::TransformPoint`

Follow-up to pull request #582 "Code cleanup BSpline Transforms from Common/Transforms" (January 17, 2022)

@ntatsisk Can you see if this obvious code clean-up makes the regular (non-recursive) BSpline Transform any faster?

### DIFF
--- a/Common/Transforms/itkAdvancedBSplineDeformableTransform.hxx
+++ b/Common/Transforms/itkAdvancedBSplineDeformableTransform.hxx
@@ -192,25 +192,11 @@ auto
 AdvancedBSplineDeformableTransform<TScalarType, NDimensions, VSplineOrder>::TransformPoint(
   const InputPointType & point) const -> OutputPointType
 {
-  /** Allocate memory on the stack: */
-  const unsigned long                         numberOfWeights = WeightsFunctionType::NumberOfWeights;
-  typename ParameterIndexArrayType::ValueType indicesArray[numberOfWeights];
-  WeightsType                                 weights;
-  ParameterIndexArrayType                     indices(indicesArray, numberOfWeights, false);
-
-  OutputPointType outputPoint;
-
-  InputPointType transformedPoint = point;
-
   /** Check if the coefficient image has been set. */
   if (!this->m_CoefficientImages[0])
   {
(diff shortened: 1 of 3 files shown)


Common/Transforms/itkAdvancedBSplineDeformableTransform.hxx
```diff
@@ -192,25 +192,11 @@ auto
...
   /** Check if the coefficient image has been set. */
   if (!this->m_CoefficientImages[0])
   {
     itkWarningMacro(<< "B-spline coefficients have not been set");
-    for (unsigned int j = 0; j < SpaceDimension; ++j)
-    {
-      outputPoint[j] = transformedPoint[j];
-    }
-    return outputPoint;
+    return point;
```
mstaring, line 199: should this not be transformedPoint?

N-Dekker: Thanks for reviewing, @mstaring but I think it's OK to just do `return point` here. The original code did: https://github.com/SuperElastix/elastix/blob/59ceb4fc01cdce8d457d105157ef47f72691b1b8/Common/Transforms/itkAdvancedBSplineDeformableTransform.hxx#L203-L214

So it copied `point` to `transformedPoint`, and then it copied the coordinates of `transformedPoint` to `outputPoint`. Then it returned `outputPoint`. In this particular case (if the coefficient image has _not_ been set), I think it's ok to return the input point directly. Right?

PR #815, opened 2023-02-25
ENH: Add `ElastixLogLevel` to the ITK interface

N-Dekker:
Added `GetLogLevel` and `SetLogLevel` member functions to both `itk::ElastixRegistrationMethod` and `itk::TransformixFilter`, allowing the user to reduce the amount of logging from elastix and transformix, both to the log file and the console (standard output).

### DIFF
--- a/Core/Kernel/elxlog.h
+++ b/Core/Kernel/elxlog.h
@@ -30,7 +30,7 @@ namespace elastix
 class log
 {
 public:
-  enum class level
+  enum class level : uint8_t
   {
     info,
     warn,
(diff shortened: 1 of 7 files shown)


mstaring:
lgtm. How would this look in python? setloglevel( integer ) or setloglevel( warn ) ?

N-Dekker:
> How would this look in python? setloglevel( integer ) or setloglevel( warn ) ?

Good question, thanks @mstaring I don't know how SWIG deals with modern C++11 enum types. ITK has a lot of those enums, so I would expect them to be wrapped nicely.

PR #822, opened 2023-03-01
Rename "MyStandardResampler" to "DefaultResampler"

N-Dekker:
Denis (@dpshamonin) suggested me to rename "MyStandardResampler" to something like "DefaultResampler".

"DefaultResampler" sounds more professional than "MyStandardResampler", and it appears more consistent with other elastix component names. (None of them start with "My".)

Note that `elxGetClassName()` already returns "DefaultResampler" (which is the name as it appears in parameter files), and it has been doing so for decades, looking at a commit from 2004:

https://github.com/SuperElastix/elastix/blob/ab62f3deca0c4782de2cb5c8abd0c6d040f893f3/src/Components/Resamplers/MyStandardResampler/elxMyStandardResampler.h#L39


dpshamonin:
Approved

Issue #756, opened 2022-11-16
Executable progress indication appears suppressed on Windows


N-Dekker:
The following logic does not seem to work on Windows: 

https://github.com/SuperElastix/elastix/blob/cf02734f11d0fc3b7e08bb140beca4f019cdaeab/Core/elxProgressCommand.cxx#L39-L46

See https://godbolt.org/z/vTEr9nfs3
```

#include <iostream>

int main()
{
    int currentPos = std::cout.tellp(); 
    std::cout << "currentPos = " << currentPos << '\n';
    return currentPos;
}

```

Output:

> Program returned: 0
> currentPos = 0

Even though Stack Overflow suggests that it should be -1: [why does cout.tellp always return -1?](https://stackoverflow.com/questions/11166398/why-does-cout-tellp-always-return-1), and [Discrimination between file and console streams](https://stackoverflow.com/questions/18081392/discrimination-between-file-and-console-streams)

The consequence is that the progress indicator won't appear on Windows.

N-Dekker:
@mstaring Do you possibly remember, is this a known problem? I see you introduces this `tellp()` logic _(Check if the output of the stream is a console)_ with commit 5a1b41eb9d46f741b986ccbbf683c105fa258b57

N-Dekker:
It was suggested (by Andrew Marlow at the accu-general mailing list, as well as by other replies at https://stackoverflow.com/questions/18081392/discrimination-between-file-and-console-streams) to use `isatty` instead:

- https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/isatty?view=msvc-160
- https://pubs.opengroup.org/onlinepubs/009695299/functions/isatty.html

N-Dekker:
@mstaring @stefanklein Looking at

https://github.com/SuperElastix/elastix/blob/cf02734f11d0fc3b7e08bb140beca4f019cdaeab/Core/elxProgressCommand.cxx#L41-L42

Doesn't `xl::xout["coutonly"].GetCOutputs().find("cout")->second` always just return a pointer to `std::cout`, by definition? If so, why not just use `std::cout` directly?

"coutonly" is initialized in `elastix::xoutSetup`, at:

https://github.com/SuperElastix/elastix/blob/cf02734f11d0fc3b7e08bb140beca4f019cdaeab/Core/Kernel/elxElastixMain.cxx#L97-L113

stefanklein:
[...]

PR #826, opened 2023-03-04
Show progress percentage only when requested by user

N-Dekker:
Until now, during a time consuming process, the progress percentage was shown _only_ by elastix/transformix executables, not by the library. Moreover, it was then _only_ shown if `std::cout.tellp()` would return -1. The return value of `std::cout.tellp()` appears platform dependent, and it does not really tell if showing the progress percentage is desirable. For details, see issue #756 "Executable progress indication appears suppressed on Windows".

This pull request proposes to add an extra `bool` flag to our `elx::log` class which specifies whether the user wants to see a progress percentage.

It extends the executable interface with an option `-progress on`, and it adds `ShowProgressPercentageOn`/`Off` member functions to the ITKElastix/library interface. By default, the progress percentage is _not_ shown (which was the behavior on Windows until now anyway).

### DIFF
--- a/Components/Optimizers/AdaGrad/elxAdaGrad.hxx
+++ b/Components/Optimizers/AdaGrad/elxAdaGrad.hxx
@@ -780,9 +780,9 @@ AdaGrad<TElastix>::SampleGradients(const ParametersType & mu0, double perturbati
   } // end if NewSamplesEveryIteration.
 
   /** Prepare for progress printing. */
-  const auto progressObserver = BaseComponent::IsElastixLibrary()
-                                  ? nullptr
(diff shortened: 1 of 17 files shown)


N-Dekker:
FYI, A more fancy progress monitor was proposed to spdlog, but rejected: issue https://github.com/gabime/spdlog/issues/854 and https://github.com/gabime/spdlog/issues/1030 by [Elmar Pruesse](https://github.com/epruesse)

N-Dekker:
Closed: abandoned, as it is superseded by pull request #832 _Support "ShowProgressPercentage" parameter (false by default)_

PR #832, opened 2023-03-11
Support "ShowProgressPercentage" parameter (false by default)

N-Dekker:
Allows enabling showing progress percentages at the console (standard output), by adding the following elastix parameter:

    (ShowProgressPercentage "true")

As I discussed last week with Marius Staring (@mstaring).

This pull request aims to supersede PR #826 "Show progress percentage only when requested by user" (which would instead add an extra command-line option `-progress on`, and `ShowProgressPercentageOn`/`Off` member functions.) It also addresses issue #756 "Executable progress indication appears suppressed on Windows".


mstaring:
lgtm!

PR #856, opened 2023-03-23
Add `SetInitialTransformParameterObject` to `ElastixRegistrationMethod`

N-Dekker:

PR #857, opened 2023-03-24
BUG: Only use InitialTransformParameterObject during first registration

N-Dekker:
When `ElastixRegistrationMethod::GenerateData()` does a sequence of multiple registrations, only the first one should use the parameter maps from the InitialTransformParameterObject. _Right?_  😃 

Follow-up to pull request https://github.com/SuperElastix/elastix/pull/856 commit 48c64583c8514eac7fa9ad46848be9d100ac80c9 "Add `SetInitialTransformParameterObject` to `ElastixRegistrationMethod`"

### DIFF
--- a/Core/Main/itkElastixRegistrationMethod.hxx
+++ b/Core/Main/itkElastixRegistrationMethod.hxx
@@ -232,7 +232,7 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GenerateData()
     unsigned int isError = 0;
     try
     {
-      isError = m_InitialTransformParameterObject
+      isError = ((i == 0) && m_InitialTransformParameterObject)
                   ? elastixMain->RunWithInitialTransformParameterMaps(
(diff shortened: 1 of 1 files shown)


mstaring:
of course the subsequent registrations should use the combination of the initial transform and the prior registrations as a combined initial transform for itself...

Did you do extensive testing here?

N-Dekker:
> Did you do extensive testing here?

@mstaring  Thanks for your suggestion. I just added a unit test for this case:
- pull request #860 

Then I realized that `elx::TransformBase::BeforeRegistrationBase()` simply ignores any initial parameter map or initial parameter file, when an initial transform object  (`m_Elastix->GetInitialTransform()`) is specified. It only looks for initial parameter maps or initial parameter files in the `else` part of this `if`:

https://github.com/SuperElastix/elastix/blob/48c64583c8514eac7fa9ad46848be9d100ac80c9/Core/ComponentBaseClasses/elxTransformBase.hxx#L173 

So the commit of this PR not really a bug fix, it's just a style improvement  😃

PR #885, opened 2023-05-08
STYLE: Implement logging without third party dependency (spdlog)

N-Dekker:
Gabi Melman's spdlog is a great library, but does not seem necessary here, as elastix was only using a fraction of its functionality.

This commit offers just what elastix needs (thread safety and log levels), while having less lines of code.


N-Dekker:
Discussed by mail, with Stefan and Marius. For now it appears OK to remove the spdlog dependency.

PR #886, opened 2023-05-08
STYLE: Rename TransformixFilter template parameter TMovingImage "TImage"

N-Dekker:
The template parameter of `TransformixFilter` is not _just_ the type of the moving image (the input image). Conceptually, it is also the type of the fixed image. Moreover, it is the type of the output image as well.

----
Note that template parameter of `TransformixFilter` was originally named `TInputImage`. It was renamed to "TMovingImage" by commit d5aa28e408fb26fb7a8c4c95c2f6332373b60e2d "ENH: Rename TransformixFilter's template parameter from TInputImage to TMovingImage to clarify mathematical definition and transformation direction", by Kasper (@kaspermarstal), 24 August 2016.

### DIFF
--- a/Core/Main/itkTransformixFilter.h
+++ b/Core/Main/itkTransformixFilter.h
@@ -55,15 +55,15 @@
 namespace itk
 {
 
-template <typename TMovingImage>
-class ITK_TEMPLATE_EXPORT TransformixFilter : public ImageSource<TMovingImage>
+template <typename TImage>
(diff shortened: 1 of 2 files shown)


N-Dekker:
Ideally I think TransformixFiltershould have two template arguments, `TInputImage` and `TOutputImage`. But that's beyond the scope of this pull request. The output pixel type may be specified by the parameter "ResultImagePixelType", but it should always correspond with the template argument anyway.

PR #895, opened 2023-05-22
ENH: Exclude retrieval of filenames of masks from elastix library

N-Dekker:
Avoided confusing log messages saying "-fMask    unspecified, so no fixed mask used" and "-mMask    unspecified, so no moving mask used" for library users (including ITKElastix).

Relates to issue https://github.com/InsightSoftwareConsortium/ITKElastix/issues/127 "registration with mask does not work", reported by jslalu, June 11, 2021.

### DIFF
--- a/Core/Kernel/elxElastixBase.cxx
+++ b/Core/Kernel/elxElastixBase.cxx
@@ -180,30 +180,34 @@ ElastixBase::BeforeAllBase()
   log::info("ELASTIX version: " ELASTIX_VERSION_STRING "\nCommand line options from ElastixBase:");
   std::string check = "";
 
-  /** Read the fixed and moving image filenames. These are obliged options,
-   * so print an error if they are not present.
-   * Print also some info (second boolean = true).
-   */
   if (!BaseComponent::IsElastixLibrary())
   {
+    // Note: The filenames of the input images and the masks are only relevant to the elastix executable, not the
+    // library.
+
+    /** Read the fixed and moving image filenames. These are obliged options,
+     * so print an error if they are not present.
+     * Print also some info (second boolean = true).
+     */
     m_FixedImageFileNameContainer = GenerateFileNameContainer(*m_Configuration, "-f", returndummy, true, true);
     m_MovingImageFileNameContainer = GenerateFileNameContainer(*m_Configuration, "-m", returndummy, true, true);
-  }
-  /** Read the fixed and moving mask filenames. These are not obliged options,
-   * so do not print any errors if they are not present.
-   * Do print some info (second boolean = true).
-   */
-  int maskreturndummy = 0;
-  m_FixedMaskFileNameContainer = GenerateFileNameContainer(*m_Configuration, "-fMask", maskreturndummy, false, true);
-  if (maskreturndummy != 0)
-  {
-    log::info("-fMask    unspecified, so no fixed mask used");
-  }
-  maskreturndummy = 0;
-  m_MovingMaskFileNameContainer = GenerateFileNameContainer(*m_Configuration, "-mMask", maskreturndummy, false, true);
-  if (maskreturndummy != 0)
-  {
-    log::info("-mMask    unspecified, so no moving mask used");
+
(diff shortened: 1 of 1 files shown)


ntatsisk:
Thanks @N-Dekker, looks good at first glance. However, I have two questions:
* Maybe is it better to remove the prints from ElastixBase altogether? Because I assume everything _below_ the `-mMask` will still be logged. Right?

>  30: ELASTIX version: 5.1.0
> 30: Command line options from ElastixBase:
> 30: -f        D:/a/elastix/elastix/Elastix-source/Testing/Data/3DCT_lung_baseline.mha
> 30: -m        D:/a/elastix/elastix/Elastix-source/Testing/Data/3DCT_lung_followup.mha
> 30: -fMask    unspecified, so no fixed mask used
> 30: -mMask    unspecified, so no moving mask used
> 30: -out      D:\a\elastix\elastix\Elastix-build\Testing\elastix_run_3DCT_lung.NC.euler.ASGD.001\
> 30: -p        D:/a/elastix/elastix/Elastix-source/Testing/Data/parameters.3D.NC.euler.ASGD.001.txt
> 30: -priority unspecified, so NORMAL process priority
> 30: -threads  unspecified, so all available threads are used
> 30: Command line options from TransformBase:
> 30: -t0       unspecified, so no initial transform used

* Does it make sense to enable logging in one of the ElastixRegistrationMethod tests so that we can visually check the log more easily? I see that only the command line tests output the log right now.

N-Dekker:
@ntatsisk Shall I already merge this one, as a tiny step forward? It doesn't address the question how much logging is useful in general, for library users. But it does at least "mitigate" the problem of issue https://github.com/InsightSoftwareConsortium/ITKElastix/issues/127 "registration with mask does not work".

ntatsisk:
Sure @N-Dekker, feel free to merge it from my side. Do you mind sharing your intuition also about the following as well?


> * Does it make sense to enable logging in one of the ElastixRegistrationMethod tests so that we can visually check the log more easily? I see that only the command line tests output the log right now.

ntatsisk (approved):

N-Dekker:
> Does it make sense to enable logging in one of the ElastixRegistrationMethod tests so that we can visually check the log more easily?

Maybe, but in general, I prefer to keep the output of a passing test minimal. In general I think a successfully passing test should just report "passed" to the dashboard. Too much output from a successful test can easily hide relevant information. I know, traditional ITK tests and elastix tests tend to be very verbose, even when they are passing successfully, and I don't really like it. 

For a failing unit test, I would like to see more information on the dashboard, though. "Internal elastix error: See elastix log (use LogToConsoleOn() or LogToFileOn())" isn't always enough for me. 🤔 
 
We can make an exception for your case, of course, but let's not make unnecessary noise at the unit test dashboard!

ntatsisk:
@N-Dekker, That's confusing to me... how can the rest of the tests provide the complete log output then? They are still run via CTest, right?

ntatsisk:
OK then it is cdash vs CI difference because the same test gives the entire output in the CI: https://github.com/SuperElastix/elastix/actions/runs/5070534017/jobs/9105697856?pr=898#step:14:10577

N-Dekker:
Temporarily added a test that produces a lot of output, for you 😺 

`GTEST_TEST(itkTransformixFilter, GetTransformParameterObjectFromRegistration)`:
https://github.com/SuperElastix/elastix/blob/99947da5f08f5ea5f1b448c722e5815e7a854351/Core/Main/GTesting/itkTransformixFilterGTest.cxx#L1465-L1490 

Output at https://github.com/SuperElastix/elastix/actions/runs/5072498636/jobs/9110282589  

From here: https://github.com/SuperElastix/elastix/actions/runs/5072498636/jobs/9110282589#step:14:28981

As well as here: https://dev.azure.com/kaspermarstal/Elastix/_build/results?buildId=4121&view=logs&j=34a65711-f108-577e-6707-abb292f048dd&t=97b2d3e0-53b7-57b8-cbbf-64a4a6e9d9bd&l=28559

Issue #873, opened 2023-04-13
"ERROR: File "output_path_not_set\IterationInfo.0.R0.txt" could not be opened" when the output path is not set


N-Dekker:
When a user does not specify the output path of elastix (command-line argument "-out"), messages appear, even in the library interface, saying things like:

> ERROR: File "output_path_not_set\IterationInfo.0.R0.txt" could not be opened!

Are these messages still useful? If so, should they really be considered _error_ messages, or are they merely informational?

N-Dekker:
Fixed by commit 99902e5da93abd1f5f6befcbd422e8e7c15cd40d "BUG: Library should write to output directory whenever it is specified" (pull request #899)

PR #899, opened 2023-05-25
BUG: Library should also write images to file if WriteResultImage = true

N-Dekker:
Until now, the elastix library would only create the result images in memory, even when the WriteResultImage parameter would be "true". This is an inconsistency between the library and the elastix executable.


mstaring:
only if the output directory is set right?

N-Dekker:
> only if the output directory is set right?

Indeed. Now it causes exceptions, saying:
> File cannot be written: output_path_not_set\result.0.mhd

Maybe we should avoid trying to write to the output directory whenever it is "output_path_not_set". Just like issue #873 ("ERROR: File "output_path_not_set\IterationInfo.0.R0.txt" could not be opened"), right?

mstaring:
Indeed, I am proposing an if statement still, but not if (libarary), but if (outputdir set)

N-Dekker:
Apparently there's a similar issue w.r.t DeformationField images. Historically, `TransformPointsAllPoints` wrote them to file only when using the executable:

https://github.com/SuperElastix/elastix/blob/2ac905afdb1a5a841c05ed6abf92a1bf232d1e89/Core/ComponentBaseClasses/elxTransformBase.hxx#L1432-L1434

Do you agree that this should instead be something like  

    if  output directory is specified
        WriteDeformationFieldImage( deformationfield );

For the library as well?

N-Dekker:
Major update: with this commit, most of the existing `GetCommandLineArgument("-out")` calls are checked now; the (various) output files are now written to this directory _if and only if_ the specified output directory path is a non-empty string. Independent of whether it is the library or the executable version of elastix/transformix. @mstaring Do you agree with this approach?

N-Dekker:
For the record, the last update is discussed with Marius (@mstaring) as well, yesterday, in person. Ready to be merged!

Issue #900, opened 2023-05-31
InitialTransformParametersFileName or InitialTransformParameterFileName? (with or without letter 's')


N-Dekker:
elastix uses the name "InitialTransformParametersFileName" (mind the letter "s") for the parameter that specifies a _parameter file_ that has an initial transform. The library introduced a slightly different name: "InitialTransformParameterFileName" (without an "s"):

https://github.com/SuperElastix/elastix/blob/99902e5da93abd1f5f6befcbd422e8e7c15cd40d/Core/Main/itkElastixRegistrationMethod.h#L195-L202 

The `InitialTransformParameterFileName` property (without an "s") was introduced by Kasper, commit a6e8f1058b9b65af19252d704bba47058d8503ac, 3 Feb 2016.

"InitialTransformParameterFileName" (without an "s") is also written to parameter files, by `ParameterObject::WriteParameterFiles`:

https://github.com/SuperElastix/elastix/blob/99902e5da93abd1f5f6befcbd422e8e7c15cd40d/Core/Main/elxParameterObject.cxx#L396  
It does make sense to drop the "s", as an apple tree isn't called "apples tree" either.

So I would propose to drop the "s" when writing new parameter files, and just keep supporting the "s" when reading old parameter files back in, for backward compatibility. What do you think?

N-Dekker:
Addressed by
- pull request https://github.com/SuperElastix/elastix/pull/903, especially commit https://github.com/SuperElastix/elastix/commit/2170592cbd018cce65698d404cae281f40c28d39
- pull request https://github.com/SuperElastix/manual/pull/3 commit https://github.com/SuperElastix/manual/commit/0259b7a2deac2cd8943b12f180fcaa65aa056721

PR #903, opened 2023-06-02
Remove the letter 's' from "InitialTransformParametersFileName"

N-Dekker:
By default, started using "InitialTransformParameterFileName" (without 's'), instead of "InitialTransformParametersFileName" (with 's'), as parameter name.

This commit still supports reading and retrieving the old parameter name, but only for backward compatibility (producing a warning).

Aims to improve consistency. The elastix manual uses the term "parameter file" (and not "parameters file"), and `itk::ElastixRegistrationMethod` has `GetInitialTransformParameterFileName` and `SetInitialTransformParameterFileName` member functions (without 's' in their name).

Addresses issue https://github.com/SuperElastix/elastix/issues/900, "InitialTransformParametersFileName or InitialTransformParameterFileName? (with or without letter 's')"


N-Dekker:
Discussed at the sprint today, we agreed to have it merged. But carefully!

PR #905, opened 2023-06-02
BUG: ElastixRegistrationMethod shouldn't put "NoInitialTransform" in map

N-Dekker:
ElastixRegistrationMethod should not set "NoInitialTransform" as parameter value of `InitialTransformParametersFileName`. The value "NoInitialTransform" might cause confusion, as the registration might still have an initial transform parameter file, specified by `SetInitialTransformParameterFileName`.

----

For the record, ElastixFilter (the predecessor of ElastixRegistrationMethod) started setting the value "NoInitialTransform" in the registration parameter map with pull request #50 commit 6a7f91641081d9b4a1d7f888695347bfdee93575,  "ENH: Explicitly ignore InitialTransformParametersFileName set via parameter maps in ElastixFilter", merged on March 20, 2018:

https://github.com/SuperElastix/elastix/blob/6a7f91641081d9b4a1d7f888695347bfdee93575/Core/Main/elxElastixFilter.hxx#L218-L222

FYI @kaspermarstal

### DIFF
--- a/Core/Main/itkElastixRegistrationMethod.hxx
+++ b/Core/Main/itkElastixRegistrationMethod.hxx
@@ -222,10 +222,6 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GenerateData()
 
     SetParameterValueAndWarnOnOverride(parameterMap, "ResultImagePixelType", fixedImagePixelTypeString);
 
-    // Initial transform parameter files are handled via arguments and enclosing loop, not
-    // InitialTransformParametersFileName
-    SetParameterValueAndWarnOnOverride(parameterMap, "InitialTransformParametersFileName", "NoInitialTransform");
-
     // Create new instance of ElastixMain
     const auto elastixMain = elx::ElastixMain::New();
     m_ElastixMain = elastixMain;


Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -222,9 +223,15 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GenerateData()
...
-    SetParameterValueAndWarnOnOverride(parameterMap, "InitialTransformParametersFileName", "NoInitialTransform");
+    // An initial transform parameter file may be specified by m_InitialTransformParameterFileName, and passed via the
+    // argument map (specifically the "-t0" argument), but not via the registration parameter map.
+    if (parameterMap.count("InitialTransformParametersFileName") != 0)
+    {
+      elx::log::warn(
+        "The parameter map entry for \"InitialTransformParametersFileName\" is ignored. "
+        "`ElastixRegistrationMethod` only supports specifying an initial transform parameter file by calling "
+        "`SetInitialTransformParameterFileName`");
+    }
```
N-Dekker, line 234: Discussed at the sprint. No need to produce a warning here, as agreed with Marius (@mstaring). The "InitialTransformParametersFileName" is not relevant for elastix registration parameter files, just ignore it. Konstantinos (@ntatsisk) asked to check if the "NoInitialTransform" wasn't there to avoid some warning/error log messages from elastix.

PR #907, opened 2023-06-07
ENH: Allow InitialTransformParameterFileName relative to parameter file

N-Dekker:
Allow the file name specified by the "InitialTransformParameterFileName" parameter in a transform parameter file to be relative to that transform parameter file.


Core/ComponentBaseClasses/elxTransformBase.hxx
```diff
@@ -388,16 +388,36 @@ TransformBase<TElastix>::ReadFromFile()
...
+      if (itksys::SystemTools::CollapseFullPath(fileName) ==
+          itksys::SystemTools::CollapseFullPath(configurationParameterFileName))
       {
         itkExceptionMacro("ERROR: The InitialTransformParameterFileName is identical to the current "
                           "TransformParameters filename! An infinite loop is not allowed.");
       }
 
-      /** We can safely read the initial transform. */
-      this->ReadInitialTransformFromFile(fileName.c_str());
+      const auto lastConfigurationParameterFilePathSeparator = configurationParameterFileName.find_last_of("\\/");
```
mstaring, line 401: does itksys not have a robust function for what you are trying to do here?

N-Dekker: > does itksys not have a robust function for what you are trying to do here?

Good question 😃 I found some itksys functions that do similar things, by splitting a path to its components, and then joining them again: 
```
  static std::vector<std::string> SplitString(const std::string& s,
                                              char separator = '/',
                                              bool isPath = false);
  static std::string Join(const std::vector<std::string>& list,
                          const std::string& separator);
```
But as far as I can see, it involves extra manual programming, it doesn't get much easier than what it does now: Just find the last separator (slash or backslash) in the current transform file path (by `std::string::find_last_of`), and then concatenate the current path _until that last separator_ with the string specified by "InitialTransformParameterFileName".

Core/ComponentBaseClasses/elxTransformBase.hxx
```diff
@@ -388,16 +388,36 @@ TransformBase<TElastix>::ReadFromFile()
...
+      const char firstFileNameLetter = fileName.front();
+      const bool startsWithFilePathSeparator = firstFileNameLetter == '\\' || firstFileNameLetter == '/';
+      const bool hasDriveLetter = !startsWithFilePathSeparator && firstFileNameLetter > 0 &&
+                                  std::isalpha(firstFileNameLetter) && fileName.size() > 1 && fileName[1] == ':';
+
+      if (lastConfigurationParameterFilePathSeparator != std::string::npos && !startsWithFilePathSeparator &&
+          !hasDriveLetter && !itksys::SystemTools::FileExists(fileName))
+      {
```
N-Dekker, line 409: I could reverse `if` and `else`, and then do:

    if (lastConfigurationParameterFilePathSeparator == std::string::npos 
        || startsWithFilePathSeparator
        || hasDriveLetter
        || itksys::SystemTools::FileExists(fileName))

To make the condition _slightly_ simpler, maybe.

Could even just test on `itksys::SystemTools::FileExists(fileName)` and forget about the other three conditions. Might make the code slightly slower sometimes, but slightly easier to read.

Core/ComponentBaseClasses/elxTransformBase.hxx
N-Dekker: ok, thanks! I [just](https://github.com/SuperElastix/elastix/compare/0e96c2f898cc38e3b2e2ed784d6caec446862c6b..088ec1100244df68c313cad23848999764b69ec2) added some more comment, and also made the code clearer by introducing a local variable, `isAbsoluteFilePath`. Hope it's clear enough now!

N-Dekker:
Note that the pull request still supports "InitialTransformParameterFileName" file paths relative to the current working directory as well (rather than relative to the directory of the current transform parameter file).  The current working directory is typically where the executable is started. This is important for backward compatibility, but still for the current executables as well. For example, suppose elastix is started in an input data directory, "C:\InputDir":

    C:\InputDir>elastix -out D:\OutputDir -f fixed.mhd -m moving.mhd -t0 InitialTransformParameters.txt -p RegistrationParameters.txt

Creates a file "D:\OutputDir\TransformParameters.0.txt" having:

    (InitialTransformParametersFileName "InitialTransformParameters.txt")

While the file InitialTransformParameters.txt could be located in "C:\InputDir". This still works fine after this pull request.

PR #914, opened 2023-06-19
COMP: ElastixRegistrationMethod interfaces types for Python wrapping

thewtex:
In the ITK Python wrapping, we require raw pointers as preferred standard on ITK Object interface types to help with Python<->C++ ownership rules and reduce types wrapped types,
i.e. raw pointers as opposed to SmartPointers and raw pointers. Update the interface type in the ITKElastix wrapped
itk::ElastixRegistrationMethod.

### DIFF
--- a/Core/Main/itkElastixRegistrationMethod.hxx
+++ b/Core/Main/itkElastixRegistrationMethod.hxx
@@ -825,7 +825,7 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNumberOfTransforms() co
 template <typename TFixedImage, typename TMovingImage>
 auto
 ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNthTransform(const unsigned int n) const
-  -> SmartPointer<TransformType>
+  -> TransformType *
 {
   const auto * const transformContainer = m_ElastixMain->GetElastixBase().GetTransformContainer();
 
@@ -847,7 +847,7 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNthTransform(const unsi
 
 template <typename TFixedImage, typename TMovingImage>
 auto
-ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform() const -> SmartPointer<TransformType>
+ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform() const -> TransformType *
 {
   const auto * const transformContainer = m_ElastixMain->GetElastixBase().GetTransformContainer();
 
@@ -868,20 +868,20 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform() 
 
 template <typename TFixedImage, typename TMovingImage>
 auto
-ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType & elxTransform)
+ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType * elxTransform)
   -> SmartPointer<TransformType>
 {
   const auto * const combinationTransform =
-    dynamic_cast<const itk::AdvancedCombinationTransform<double, FixedImageDimension> *>(&elxTransform);
+    dynamic_cast<const itk::AdvancedCombinationTransform<double, FixedImageDimension> *>(elxTransform);
 
   const auto itkTransform = combinationTransform
                               ? elx::TransformIO::ConvertToCompositionOfItkTransforms(*combinationTransform)
-                              : elx::TransformIO::ConvertToSingleItkTransform(elxTransform);
+                              : elx::TransformIO::ConvertToSingleItkTransform(*elxTransform);
   if (itkTransform)
   {
     return itkTransform;
   }
-  itkGenericExceptionMacro("Failed to convert transform object " << elxTransform);
+  itkGenericExceptionMacro("Failed to convert transform object " << *elxTransform);
 }
 
 } // namespace itk
--- a/Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx
+++ b/Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx
@@ -1505,7 +1505,7 @@ GTEST_TEST(itkElastixRegistrationMethod, ConvertToItkTransform)
     for (unsigned int n{ 0 }; n < numberOfTransforms; ++n)
     {
       // TODO Check result
-      ElastixRegistrationMethodType<ImageType>::ConvertToItkTransform(*registration.GetNthTransform(n));
+      ElastixRegistrationMethodType<ImageType>::ConvertToItkTransform(registration.GetNthTransform(n));
     }
   }
 }
--- a/Core/Main/itkElastixRegistrationMethod.h
+++ b/Core/Main/itkElastixRegistrationMethod.h
@@ -267,17 +267,17 @@ public:
   GetNumberOfTransforms() const;
 
   /** Returns the nth transformation, produced during the last Update(). */
-  SmartPointer<TransformType>
+  TransformType *
   GetNthTransform(const unsigned int n) const;
 
   /** Returns the combination transformation, produced during the last Update(). */
-  SmartPointer<TransformType>
+  TransformType *
   GetCombinationTransform() const;
 
   /** Converts the specified elastix Transform object to the corresponding ITK Transform object. Returns null if there
    * is no corresponding ITK Transform type. */
   static SmartPointer<TransformType>
-  ConvertToItkTransform(const TransformType &);
+  ConvertToItkTransform(const TransformType *);
 
 protected:
   ElastixRegistrationMethod();


thewtex:
CC @tbirdsong

N-Dekker:
@thewtex Thanks for your pull request. It's really a pity if the ITK Python wrapping cannot return a smart pointer. For ElastixRegistrationMethod::ConvertToItkTransform it's actually essential, to ensure the lifetime of the returned transform object. Do you actually get error messages from those smart pointer return types?

thewtex:
@N-Dekker yes, I think returning the SmartPointer still works (retained in this PR).

N-Dekker:
Thanks Matt. As this PR is marked "COMP", I assume that the current `main` causes compilation errors that are addressed by this PR, right? If so, can you please specify which compiler errors you have encountered? Maybe a link to a dashboard?

thewtex:
@N-Dekker yes, this is compiler / platform related (Python packaging), but there are not dashboard errors associated -- this is the standard to support building for the ITK Python packaging.

I tested locally and did not get any issues. Why are we seeing build failures in CI?

N-Dekker:
> Why are we seeing build failures in CI?

https://my.cdash.org/viewBuildError.php?buildid=2354908 says:

> [Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx:1508](https://github.com/SuperElastix/elastix/blob/master/Core/Main/GTesting/itkElastixRegistrationMethodGTest.cxx#L1508):71: error: no viable conversion from 'itk::ElastixRegistrationMethod<itk::Image<float, 2>, itk::Image<float, 2>>::TransformType' (aka 'Transform<double, FixedImageDimension, MovingImageDimension>') to 'const itk::ElastixRegistrationMethod<itk::Image<float, 2>, itk::Image<float, 2>>::TransformType *' (aka 'const Transform<double, FixedImageDimension, MovingImageDimension> *')

thewtex:
Ah, I needed to enable `ELASTIX_USE_GTEST` -- test compilation fixed.

Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -868,20 +868,20 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform()
 
 template <typename TFixedImage, typename TMovingImage>
 auto
-ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType & elxTransform)
+ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType * elxTransform)
```
N-Dekker, line 871: @thewtex You know, it's a best practice in C++ to prefer pass-by-reference over pass-by-pointer. (Unless "null" should be supported as a valid argument) So if it does not cause any compilation errors, I would rather leave it as it is right now. Sorry.

thewtex: Yes, I understand and agree that C++ prefers pass-by-reference, but this is needed for Python, where we do not have references.

N-Dekker: Sorry, I don't understand what you mean by "this is needed for Python". I just tried the following in Python, on the "old" ITKElastix 0.17.1, and it appears to work just fine, internally passing the transform parameter by-reference:

```
elastix_transform = elastix_object.GetNthTransform(0)
itk_transform = itk.ElastixRegistrationMethod[ImageType, ImageType].ConvertToItkTransform(elastix_transform)

```
It appears that SWIG knows how to deal with C++ references. Did you encounter any specific technical problem with ITKElastix 0.17.1 `ConvertToItkTransform`?

thewtex: Yes, it will run, but that does not mean that the reference counting is correct.

For ITK, SWIG, itk::LightObject, itk::SmartPointer, we have defined ownership rules for raw pointers for itk::LightObject. Not C++ references.

N-Dekker: Thanks Matt, but can you please show a minimal example where the example `ConvertToItkTransform(elastix_transform)` would go wrong, using the 0.17.1 version (pass-by-reference)?

Note that the argument is a `const reference`. It is _not_ modified by `ConvertToItkTransform`, and its lifetime does not need to be extended beyond the `ConvertToItkTransform` function call. The function returns a new transform object, which does not refer to the argument. So as far as I can see, the argument does not need to be reference counted.

thewtex: C++ is different than Python. C++ references, `const` are a compile-time C++ constructs. They do not exist when we get to the CPython runtime. At CPython runtime, there is no `const` and no references, and the referencing counting I am referring to here is not just referring C++ and C++ runtime. At CPython runtime, we only have pointers, `PyObject` reference counting and `itk::LightObject` reference counting and lifetimes and ownership rules need to be correct. Please study the resulting CPython wrapping `*Python.cpp` and ITK's SWIG configuration in `ITK/Wrapping` for more information.

tbirdso: @thewtex Following along here, am I understanding correctly that the issue at hand is the question of `itk::LightObject` reference counting? I.e.

```py
>>> import itk
>>> t = itk.TranslationTransform[itk.D,3].New()
>>> print(t)
TranslationTransform (000001F8A9340530)
  RTTI typeinfo:   class itk::TranslationTransform<double,3>
  Reference Count: 1  #<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< here
  Modified Time: 1
  Debug: Off
  Object Name:
  Observers:
    none
  Offset: [0, 0, 0]
```

Is there somewhere else where an incorrect reference count could be observed in regards to `ConvertToItkTransform`? Or, am I misunderstanding the issue altogether?

Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -825,7 +825,7 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNumberOfTransforms() co
 template <typename TFixedImage, typename TMovingImage>
 auto
 ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNthTransform(const unsigned int n) const
-  -> SmartPointer<TransformType>
+  -> TransformType *
```
N-Dekker, line 828: OK, you convinced me here: it does indeed look fine to me to replace the smart pointer by a raw pointer here, because the "NthTransform" is still owned (and kept alive) by ElastixRegistrationMethod. Thanks Matt.

Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -847,7 +847,7 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNthTransform(const unsi
 
 template <typename TFixedImage, typename TMovingImage>
 auto
-ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform() const -> SmartPointer<TransformType>
+ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform() const -> TransformType *
```
N-Dekker, line 850: OK, LFTM to replace the smart pointer by a raw pointer here, because the "CombinationTransform" is still owned (and kept alive) by ElastixRegistrationMethod. Thanks again Matt.

Core/Main/itkElastixRegistrationMethod.hxx
N-Dekker: Just for the record, I still need more time to see why passing a parameter by const-reference might _in this particular case_ (`ConvertToItkTransform`) cause any possible problem. A code example that would reproduce the problem would be helpful to me. I'm certainly interested to see the resulting CPython wrapping as well.

thewtex: @N-Dekker the issue at hand is `itk::LightObject` reference counting **and** `PyObject` reference counting. They have different reference counts. And they are distinct from what a C++ compiler is doing with `const` and references at compile time. I encourage you to create an example to understand how C++ const references are and **are not** related and what is involved in this particular case and in general. It is non-trivial to create a minimal example because of the code involved and see how it is related because of what can be observed at runtime in the Python interpreter. Again, this is a runtime, CPython+ITK+Swig ownership semantic and reference counting mechanics issue. Please do take a look at the CPython wrappings, which is already possible.

tbirdso: @thewtex After digging into the SWIG C++ generated wrapping layer, I am still a bit confused on the interaction between pass-by-reference and SWIG/ITK/PyObject reference counting that is at the core of this discussion. While the  change here is relatively minor, it would be good to establish the root cause to help better inform ITK guidelines for future development.

Is this change intended to fix an issue with `itk::LightObject` reference counting, `PyObject` reference counting, or both?

For reference, here is the SWIG C++ block generated in `itkElastixRegistrationMethodPython.cpp` under the status quo:
```cpp
SWIGINTERN PyObject *_wrap_itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  itkTransformD22 *arg1 = 0 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  PyObject *swig_obj[1] ;
  itkTransformD22_Pointer result;
  
  if (!args) SWIG_fail;
  swig_obj[0] = args;
  res1 = SWIG_ConvertPtr(swig_obj[0], &argp1, SWIGTYPE_p_itkTransformD22,  0  | 0);
  if (!SWIG_IsOK(res1)) {
    SWIG_exception_fail(SWIG_ArgError(res1), "in method '" "itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform" "', argument " "1"" of type '" "itkTransformD22 const &""'"); 
  }
  if (!argp1) {
    SWIG_exception_fail(SWIG_ValueError, "invalid null reference " "in method '" "itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform" "', argument " "1"" of type '" "itkTransformD22 const &""'"); 
  }
  arg1 = reinterpret_cast< itkTransformD22 * >(argp1);
  {
    try {
      result = itkElastixRegistrationMethodIF2IF2::ConvertToItkTransform((itkTransformD22 const &)*arg1);
    } catch (const std::out_of_range& e) {
      SWIG_exception_fail(SWIG_IndexError, e.what());
      //  } catch (Swig::DirectorException &e) {
      //    SWIG_fail;
    } catch (const std::exception& e) {
      SWIG_exception_fail(SWIG_RuntimeError, e.what());
    }
  }
  {
    // get the raw pointer from the smart pointer
    itkTransformD22 * ptr = result;
    // always tell SWIG_NewPointerObj we're the owner
    resultobj = SWIG_NewPointerObj((void *) ptr, SWIGTYPE_p_itkTransformD22, 1);
    // register the object, it it exists
    if (ptr) {
      ptr->Register();
    }
  }
  return resultobj;
fail:
  return NULL;
}
```

For comparison, here is the same block after rebuilding from the changes proposed in this PR:
```cpp
SWIGINTERN PyObject *_wrap_itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  itkTransformD22 *arg1 = (itkTransformD22 *) 0 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  PyObject *swig_obj[1] ;
  itkTransformD22_Pointer result;
  
  if (!args) SWIG_fail;
  swig_obj[0] = args;
  res1 = SWIG_ConvertPtr(swig_obj[0], &argp1,SWIGTYPE_p_itkTransformD22, 0 |  0 );
  if (!SWIG_IsOK(res1)) {
    SWIG_exception_fail(SWIG_ArgError(res1), "in method '" "itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform" "', argument " "1"" of type '" "itkTransformD22 const *""'"); 
  }
  arg1 = reinterpret_cast< itkTransformD22 * >(argp1);
  {
    try {
      result = itkElastixRegistrationMethodIF2IF2::ConvertToItkTransform((itkTransformD22 const *)arg1);
    } catch (const std::out_of_range& e) {
      SWIG_exception_fail(SWIG_IndexError, e.what());
      //  } catch (Swig::DirectorException &e) {
      //    SWIG_fail;
    } catch (const std::exception& e) {
      SWIG_exception_fail(SWIG_RuntimeError, e.what());
    }
  }
  {
    // get the raw pointer from the smart pointer
    itkTransformD22 * ptr = result;
    // always tell SWIG_NewPointerObj we're the owner
    resultobj = SWIG_NewPointerObj((void *) ptr, SWIGTYPE_p_itkTransformD22, 1);
    // register the object, it it exists
    if (ptr) {
      ptr->Register();
    }
  }
  return resultobj;
fail:
  return NULL;
}
```

Comparing with a `diff` tool shows that the only notable change between these blocks lies in how the SWIG input argument is passed:
```cpp
result = itkElastixRegistrationMethodIF2IF2::ConvertToItkTransform((itkTransformD22 const &)*arg1);
```
```cpp
result = itkElastixRegistrationMethodIF2IF2::ConvertToItkTransform((itkTransformD22 const *)arg1);
```

If we are considering `itk::LightObject` reference counts, where does the difference in behavior occur? Is there implicit incrementing/decrementing happening on assignment, and if can you point to where the issue lies in the wrapping above?

If we are considering `PyObject` reference counts, is there another file where we should look for differences in behavior?

If I'm entirely in the wrong place, could you at least provide a minimum example to demonstrate memory leakage from the current implementation?

Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -868,20 +868,20 @@ ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetCombinationTransform()
 
 template <typename TFixedImage, typename TMovingImage>
 auto
-ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType & elxTransform)
+ElastixRegistrationMethod<TFixedImage, TMovingImage>::ConvertToItkTransform(const TransformType * elxTransform)
   -> SmartPointer<TransformType>
```
tbirdso, line 872: Following from other changes in this PR, shouldn't this also be a raw pointer instead of an `itk::SmartPointer`?

My understanding is that `itk::Transform` objects are `itk::LightObjects` and have their own internal reference counts. As a result it seems possible that wrapping with a `SmartPointer` is introducing a redundant reference count, though possibly with different management rules.

In the SWIG C++ generated wrapping layer only the underlying transform object is explicitly incremented regardless of whether the return type is a raw pointer or `itk::SmartPointer`. For comparison with proposed return type change to `GetCombinationTransform`:

`itk::SmartPointer` return type:
```cpp
SWIGINTERN PyObject *_wrap_itkElastixRegistrationMethodIF2IF2_GetCombinationTransform(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  itkTransformD22_Pointer result;
...
result = ((itkElastixRegistrationMethodIF2IF2 const *)arg1)->GetCombinationTransform();
...
{
    // get the raw pointer from the smart pointer
    itkTransformD22 * ptr = result;
    // always tell SWIG_NewPointerObj we're the owner
    resultobj = SWIG_NewPointerObj((void *) ptr, SWIGTYPE_p_itkTransformD22, 1);
    // register the object, it it exists
    if (ptr) {
      ptr->Register();
    }
  }
  return resultobj;
```

Raw pointer return type:
```cpp
SWIGINTERN PyObject *_wrap_itkElastixRegistrationMethodIF2IF2_GetCombinationTransform(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
itkTransformD22 *result = 0 ;
...
result = (itkTransformD22 *)((itkElastixRegistrationMethodIF2IF2 const *)arg1)->GetCombinationTransform();
...
{
    // always tell SWIG_NewPointerObj we're the owner
    resultobj = SWIG_NewPointerObj((void *) result, SWIGTYPE_p_itkTransformD22, 1);
										
    if (result) {
      result->Register();
    }
  }
  return resultobj;
```

Core/Main/itkElastixRegistrationMethod.hxx
thewtex: There are [ownership rules set up](https://github.com/InsightSoftwareConsortium/ITK/blob/1551ff3f2306346dae7b385680277e75580acf09/Wrapping/Generators/Python/PyBase/pyBase.i#L71-L134) for SWIG/itk::LightObject/PyObject *together*. CPython has its own [ownership rules](  https://docs.python.org/3/c-api/intro.html#objects-types-and-reference-counts).

We have a contract of ownership rules set up for itk::LightObject raw pointers on the interfaces that accounts for CPython reference counting, CPython method argument reference counting, and ITK method argument reference counting. This contract is defined for how *raw pointers* behave at *runtime* in *C*. What a C++ compiler may happen to do with a reference at *compile time* is not part of this contract of ownership rules.

> result = itkElastixRegistrationMethodIF2IF2::ConvertToItkTransform((itkTransformD22 const &)*arg1);

`itkTransformD22` are wrapped raw pointers. We are passing in wrapped raw pointers for the Python wrapping. 

Creating a minimal example for this case is not feasible.

tbirdso: Takeaways from offline discussion with @thewtex:

- ITK Python wrapping is based on predefined rules ("contract") governing ownership behavior. Much of this is out of necessity for interoperability between the Python runtime and C++ ownership rules.
- SWIG Python creates managed Python classes wrapping around underlying ITK C++ classes. The lifetime of a wrapping Python class and underlying C++ class are generally, but not guaranteed to be, tightly linked. For instance, it may be possible to have two different wrapping instances point to one underlying C++ object, which would be a case where reference counts diverge. See [SWIG `.thisown` documentation](https://www.swig.org/Doc1.3/Python.html) (31.4.2 Memory Management) for more information.
- It is difficult to provide a minimum reproducible example because C++ PyObject reference counts are generally visible only in Python debug mode. We could potentially construct an example by building/acquiring a Debug version of the Python runtime, rebuilding ITKElastix and all of its dependencies in Debug mode, running a small `ConvertToITKTransform`, and examining the reference counts of the outputs.
- The "contract" between ITK and Python and managed by SWIG is not necessarily well documented. In general it is predicated on memory management via reference counting in objects described by C-style raw pointers for compatibility with the Python C API. Some starting references are below:
  - https://github.com/InsightSoftwareConsortium/ITK/blob/1551ff3f2306346dae7b385680277e75580acf09/Wrapping/Generators/Python/PyBase/pyBase.i#L71-L134
  - https://docs.python.org/3/c-api/intro.html#objects-types-and-reference-counts

With those details considered, I am still in doubt that the proposed change fixes an actual issue under the existing SWIG implementation. Based on the generated C++ code I shared above there are not discernable concerns regarding reference counts in the SWIG generated interface for either a reference or raw pointer parameter. _However,_ I am in favor of maintaining best practices and compliance with existing API rules to avoid breakage in the future, which in this case appears to mean adherence to the rule of accepting raw pointers over reference counts in the public API of any given ITK module.

@thewtex , please feel free to clarify if I have misunderstood anything. @N-Dekker , please mark as resolved if you feel this answers your concerns sufficiently, or indicate what else you feel needs to happen to move forward.

Core/Main/itkElastixRegistrationMethod.hxx
thewtex: Yes, for Python wrapping, we could just return a raw pointer. And the wrapping object will `Register` and `UnRegister` as partially observed in the generated code.

In pure C++, returning a SmartPointer on an `itk::LightObject` that was created in the function will [Register](https://github.com/InsightSoftwareConsortium/ITK/blob/1551ff3f2306346dae7b385680277e75580acf09/Modules/Core/Common/include/itkSmartPointer.h#L205) and [UnRegister](https://github.com/InsightSoftwareConsortium/ITK/blob/1551ff3f2306346dae7b385680277e75580acf09/Modules/Core/Common/include/itkSmartPointer.h#L103).

Core/Main/itkElastixRegistrationMethod.hxx
N-Dekker: Thanks very much for all your time on this PR, @thewtex and @tbirdso. Tom, your SWIG output (the generated C++ blocks) are very helpful to me. I see that the status quo (pass by reference) adds an extra check:

    if (!argp1) {
      SWIG_exception_fail(SWIG_ValueError, "invalid null reference " "in method '"   "itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform" "', argument " "1"" of type '" "itkTransformD22 const &""'"); 
    }

This is really great 👍 It means that when using pass-by-reference, accidentally passing NULL is prevented. So I still believe pass-by-reference is the right thing to do for this input parameter.

Core/Main/itkElastixRegistrationMethod.hxx
N-Dekker: In this particular case (specifically for `GetCombinationTransform()`), I agree that returning a raw pointer should be fine. Because the "CombinationTransform" is still owned by the ElastixRegistrationMethod object anyway. Shall I make a separate PR to just do this adjustment for `GetCombinationTransform()`?

thewtex:
The null check is required because the CPython runtime does not know what the C++ compiler is doing with the reference and so a check is needed -- an extra check and the associated uncertainty is not preferable.

But it is really not worth stopping the continuation of progress. I will make the argument update as requested sometime when I find the time.

N-Dekker:
Thanks for elaborating on this topic, Matt.

> The null check is required because the CPython runtime does not know what the C++ compiler is doing with the reference and so a check is needed

I just tried the following in Python, using ITKElastix 0.17.1:

    itk.ElastixRegistrationMethod[ImageType, ImageType].ConvertToItkTransform(None)

It produces an error message saying:

> `ValueError: invalid null reference in method 'itkElastixRegistrationMethodIF2IF2_ConvertToItkTransform', argument 1 of type 'itkTransformD22 const &'`

I think such an error message is really appropriate here, because ConvertToItkTransform expects a reference to a valid transform object as argument. Unfortunately it looks like we don't agree in this case. Sorry.

> I will make the argument update as requested sometime when I find the time.

Fortunately we do agree that `GetNthTransform` and `GetCombinationTransform` should return a raw pointer. Do you still have time to just fix these two member functions? Otherwise I wouldn't mind doing it myself.

Issue #916, opened 2023-06-19
GCC 13 cstdint include


thewtex:
`main` with GCC 13:

```
/home/matt/src/elastix/Core/Kernel/elxlog.h:1:1: note: ‘uint8_t’ is def
ined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
  +++ |+#include <cstdint>
    1 | /*=============================================================
============
```

N-Dekker:
Thanks @thewtex So you think it's sufficient to just add an `#include <cstdint>` inside elastix/Core/Kernel/elxlog.h, to fix both this issue and issue #915 (enum usage)?

thewtex:
@N-Dekker that will help to fix this issue, but I think #915 is separate.

PR #917, opened 2023-06-20
COMP: Add `#include <cstdint>` to elxlog.h, for `std::uint8_t`

N-Dekker:
Aims to address a note from GCC 13 saying:

> note: ‘uint8_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?

Reported by Matt McCormick (@thewtex), issue https://github.com/SuperElastix/elastix/issues/916 "GCC 13 cstdint include".

Explicitly specified its `std` namespace, as `<cstdint>` may not add this typedef to the global namespace.

### DIFF
--- a/Core/Kernel/elxlog.h
+++ b/Core/Kernel/elxlog.h
@@ -18,6 +18,7 @@
 #ifndef elxlog_h
 #define elxlog_h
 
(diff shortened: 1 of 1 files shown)


thewtex:
@N-Dekker thank you!

PR #920, opened 2023-06-21
WebAssembly Support

thewtex:
1. Fixes build issues with WebAssembly (Wasm) toolchains (both Emscripten and WASI)
2. Remove file-IO for Wasm builds because Wasm requires special handling of filesystem access and the ITK libraries required for IO will bloat the WASM binary size.

### DIFF
--- a/Common/itkComputeImageExtremaFilter.hxx
+++ b/Common/itkComputeImageExtremaFilter.hxx
@@ -176,7 +176,9 @@ ComputeImageExtremaFilter<TInputImage>::ThreadedGenerateDataImageSpatialMask(con
     } // end for
   }   // end if
 
+#ifndef __wasi__
   std::lock_guard<std::mutex> mutexHolder(m_Mutex);
+#endif
   m_ThreadSum += sum;
   m_SumOfSquares += sumOfSquares;
   m_Count += count;
@@ -227,7 +229,9 @@ ComputeImageExtremaFilter<TInputImage>::ThreadedGenerateDataImageMask(const Regi
     ++it;
   } // end while
 
+#ifndef __wasi__
   std::lock_guard<std::mutex> mutexHolder(m_Mutex);
+#endif
   m_ThreadSum += sum;
   m_SumOfSquares += sumOfSquares;
   m_Count += count;
--- a/Components/Metrics/PolydataDummyPenalty/elxPolydataDummyPenalty.hxx
+++ b/Components/Metrics/PolydataDummyPenalty/elxPolydataDummyPenalty.hxx
@@ -20,6 +20,11 @@
 
 #include <typeinfo>
 
+#ifndef __wasm32__
+#include "itkMeshFileReader.h"
+#include "itkMeshFileWriter.h"
+#endif
+
 namespace elastix
 {
 
@@ -270,6 +275,12 @@ template <class TElastix>
 unsigned int
 PolydataDummyPenalty<TElastix>::ReadMesh(const std::string & meshFileName, typename FixedMeshType::Pointer & mesh)
 {
+#ifdef __wasm32__
+  const std::string message = "File IO not supported in WebAssembly builds.";
+  log::error(message);
+  itkExceptionMacro(<< message);
+  return 0;
+#else
   /** Read the input mesh. */
   auto meshReader = itk::MeshFileReader<MeshType>::New();
   meshReader->SetFileName(meshFileName);
@@ -291,6 +302,7 @@ PolydataDummyPenalty<TElastix>::ReadMesh(const std::string & meshFileName, typen
   log::info(std::ostringstream{} << "  Number of specified input points: " << nrofpoints);
 
   return nrofpoints;
+#endif
 } // end ReadMesh()
 
 
@@ -302,6 +314,11 @@ template <class TElastix>
 void
 PolydataDummyPenalty<TElastix>::WriteResultMesh(const std::string & filename, MeshIdType meshId)
 {
+#ifdef __wasm32__
+  const std::string message = "File IO not supported in WebAssembly builds.";
+  log::error(message);
+  itkExceptionMacro(<< message);
+#else
   /** Set the points of the latest transformation. */
   const MappedMeshContainerPointer mappedMeshContainer = this->GetModifiableMappedMeshContainer();
   FixedMeshPointer                 mappedMesh = mappedMeshContainer->ElementAt(meshId);
@@ -360,7 +377,7 @@ PolydataDummyPenalty<TElastix>::WriteResultMesh(const std::string & filename, Me
   {
     mappedMesh->SetCellData(nullptr);
   }
-
+#endif
 } // end WriteResultMesh()
 
 
--- a/Core/ComponentBaseClasses/elxResamplerBase.hxx
+++ b/Core/ComponentBaseClasses/elxResamplerBase.hxx
@@ -22,7 +22,6 @@
 #include "elxConversion.h"
 #include "elxDeref.h"
 
-#include "itkImageFileCastWriter.h"
 #include "itkChangeInformationImageFilter.h"
 #include "itkAdvancedRayCastInterpolateImageFunction.h"
 #include "itkTimeProbe.h"
--- a/Core/ComponentBaseClasses/elxResamplerBase.hxx
+++ b/Core/ComponentBaseClasses/elxResamplerBase.hxx
@@ -395,12 +395,17 @@ ResamplerBase<TElastix>::WriteResultImage(OutputImageType *   image,
   {
     log::to_stdout("\n  Writing image ...");
   }
+#ifndef __wasm32__
   try
   {
-    itk::WriteCastedImage(*(infoChanger->GetOutput()), filename, resultImagePixelType, doCompression);
+   itk::WriteCastedImage(*(infoChanger->GetOutput()), filename, resultImagePixelType, doCompression);
   }
   catch (itk::ExceptionObject & excp)
   {
+#else
+// Always throw -- do not include support code or access filesystem with wasm
+    itk::ExceptionObject excp;
+#endif
     /** Add information to the exception. */
     excp.SetLocation("ResamplerBase - AfterRegistrationBase()");
     std::string err_str = excp.GetDescription();
@@ -408,8 +413,12 @@ ResamplerBase<TElastix>::WriteResultImage(OutputImageType *   image,
     excp.SetDescription(err_str);
 
     /** Pass the exception to an higher level. */
+#ifndef __wasm32__
     throw;
   }
+#else
+    throw excp;
+#endif
 } // end WriteResultImage()
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,23 @@ if(ELASTIX_USE_OPENCL)
   list(APPEND _GPU_depends ITKGPUCommon)
 endif()
 
+set(_ITK_io_depends
+    ITKImageIO
+    ITKTransformIO
+    ITKMeshIO
+    ITKIOImageBase
+    ITKIOMeshBase
+    ITKIOMeshOBJ
+    ITKIOMeta
+    ITKIOTransformBase
+    ITKIOXML
+)
+
+if(WASI OR EMSCRIPTEN)
+  # Keep Wasm binaries small
+  set(_ITK_io_depends)
+endif()
+
 # Find ITK.
 find_package(ITK 5.3 REQUIRED COMPONENTS
     ITKCommon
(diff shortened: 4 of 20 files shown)


Core/Main/elxMainExeUtilities.cxx
```diff
@@ -108,7 +112,7 @@ elastix::GetExtendedVersionInformation(const char * const executableName, const
 #  endif
 #endif
                      << newline << "Memory address size: " << std::numeric_limits<std::size_t>::digits << "-bit"
-                     << newline << "CMake version: " ELX_CMAKE_VERSION << newline
+                     << newline << "CMake version: " << ELX_CMAKE_VERSION << newline
```
N-Dekker, line 115: Please revert this change, if it is unnecessary. The preprocessor should still be able to do string concatenation, right?

N-Dekker: Do I understand correctly that you do not use CMake, when building elastix for Wasm?

Components/Metrics/PolydataDummyPenalty/elxPolydataDummyPenalty.hxx
```diff
@@ -270,6 +275,10 @@ template <class TElastix>
 unsigned int
 PolydataDummyPenalty<TElastix>::ReadMesh(const std::string & meshFileName, typename FixedMeshType::Pointer & mesh)
 {
+#ifdef __wasm32__
+  log::error(std::ostringstream{} << "File IO not supported in WebAssembly builds.");
```
N-Dekker, line 279: Please pass the string directly to log::error, without `std::ostringstream{} <<`.  (ostringstream is only necessary when a string needs to be built by `<<` insertions). So please just:

    log::error("File IO not supported in WebAssembly builds.");

N-Dekker: BTW, maybe an `itkExceptionMacro("File IO not supported in WebAssembly builds.");` would be more appropriate, right?

Common/OpenCL/ITKimprovements/itkGPUDataManager.cxx
```diff
@@ -155,7 +155,9 @@ GPUDataManager::UpdateCPUBuffer()
     return;
   }
 
+#ifndef __wasi__
   MutexHolderType holder(m_Mutex);
+#endif
```
N-Dekker, line 160: Does WASI support the OpenCL elastix components in general?

Common/itkComputeImageExtremaFilter.hxx
```diff
@@ -176,7 +176,9 @@ ComputeImageExtremaFilter<TInputImage>::ThreadedGenerateDataImageSpatialMask(con
     } // end for
   }   // end if
 
+#ifndef __wasi__
   std::lock_guard<std::mutex> mutexHolder(m_Mutex);
+#endif
```
N-Dekker, line 181: Does WASI support multi-threading? Specifically, will this ComputeImageExtremaFilter run using multiple threads in parallel?

Components/Metrics/PolydataDummyPenalty/elxPolydataDummyPenalty.hxx
thewtex: Yes, we could do both -- WebAssembly has limited C++ exception support at the moment, so it will help to also log an error message.

Core/CMakeLists.txt
```diff
@@ -192,7 +192,7 @@ add_executable(elastix_exe
 )
 set_target_properties(elastix_exe PROPERTIES OUTPUT_NAME elastix)
 target_compile_definitions(elastix_exe PRIVATE ELX_CMAKE_VERSION="${CMAKE_VERSION}")
-target_link_libraries(elastix_exe ${ELASTIX_TARGET_LINK_LIBRARIES})
+target_link_libraries(elastix_exe PUBLIC ${ELASTIX_TARGET_LINK_LIBRARIES})
```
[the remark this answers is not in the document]
thewtex: Yes, this was in a separate commit, but it was dropped from the branch.

Core/Main/elxMainExeUtilities.cxx
thewtex: It does use CMake, but the preprocessor variable is not set, so a fallback is provided.

Common/OpenCL/ITKimprovements/itkGPUDataManager.cxx
thewtex: OpenCL features are not supported.

Common/itkComputeImageExtremaFilter.hxx
thewtex: Currently, the default WASI toolchain does not support multi-threading. This is being improved, but we need these for now.

Common/OpenCL/ITKimprovements/itkGPUDataManager.cxx
N-Dekker: OK, thanks, so then source files from Common/OpenCL do not need to be adjusted, right? Because they are excluded from compilation anyway, when doing a WASI build.

Common/CostFunctions/itkAdvancedImageToImageMetric.hxx
```diff
@@ -781,7 +781,7 @@ AdvancedImageToImageMetric<TFixedImage, TMovingImage>::GetValueThreaderCallback(
 
   temp->st_Metric->ThreadedGetValue(threadID);
 
-  return itk::ITK_THREAD_RETURN_DEFAULT_VALUE;
+  return ITK_THREAD_RETURN_DEFAULT_VALUE;
```
N-Dekker, line 784: This commit is easy for me to grasp. I don't like the unnecessary `itk::` in front of each ITK_THREAD_RETURN_DEFAULT_VALUE anyway. Shall I already cherry-pick and merge this one? (The other commits may take me some more time.)

Components/Metrics/PolydataDummyPenalty/elxPolydataDummyPenalty.hxx
N-Dekker: Thanks, can you please explain what you mean by _limited C++ exception support_?

thewtex: The current Emscripten toolchain adds a JavaScript shim for C++ exceptions, and I added a shim for C++ exceptions in WASI that calls `abort();`. There is newer, experimental WASM-exceptions in some newer toolchains, but I have not enabled them by default.

Common/itkComputeImageExtremaFilter.hxx
N-Dekker: Sorry, still not very happy about having `#ifndef __wasi__` all over the place. Would it be an idea to have a "dummy" mutex class (that does not do anything) for single-threaded builds.

    #ifdef ELX_SINGLE_THREADED_BUILD
      using ElastixMutexType = DummyMutex;  // Mutex-like dummy type that does nothing
    #else
      using ElastixMutexType = std::mutex;
    #endif

Core/Main/elxMainExeUtilities.cxx
N-Dekker: Does it process elastix\Core\CMakeLists.txt ?

Common/OpenCL/ITKimprovements/itkGPUDataManager.cxx
thewtex: OK, I will remove this from the patch.

Common/CostFunctions/itkAdvancedImageToImageMetric.hxx
thewtex: Feel free to cherry-pick / merge.

I need all the commits in this branch for a working build used by 
https://github.com/InsightSoftwareConsortium/ITKElastix/pull/227

Core/Main/elxMainExeUtilities.cxx
thewtex: Yes.

In the most recent commit, I removed the addition of `<<` here.

Common/itkComputeImageExtremaFilter.hxx
thewtex: Sorry, I do not see how a dummy mutex type addresses the `std::lock_guard` and the mutex class members.

Core/CMakeLists.txt
```diff
@@ -254,13 +258,13 @@ set(ELASTIX_LIBRARIES elastix_lib transformix_lib PARENT_SCOPE)
...
 
-if(NOT WIN32)
+if(NOT WIN32 AND NOT WASI AND NOT EMSCRIPTEN)
   # Tell the executables where to find the required .so files.
   set_target_properties(elastix_exe transformix_exe
     PROPERTIES INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib:${ITK_DIR}")
 endif()
 
-if(NOT ELASTIX_NO_INSTALL_EXECUTABLES)
+if(NOT ELASTIX_NO_INSTALL_EXECUTABLES AND NOT WASI AND NOT EMSCRIPTEN)
```
N-Dekker, line 267: I guess this could instead be achieved at the WASI side, by setting `ELASTIX_NO_INSTALL_EXECUTABLES`. Right?

Core/CMakeLists.txt
```diff
@@ -216,20 +218,22 @@ target_link_libraries(elastix_lib ${ELASTIX_TARGET_LINK_LIBRARIES})
...
-  Kernel/elxMainBase.cxx
-  Kernel/elxMainBase.h
-  Kernel/elxTransformixMain.cxx
-  Kernel/elxTransformixMain.h
-  ${InstallFilesForExecutables}
-)
-set_target_properties(transformix_exe PROPERTIES OUTPUT_NAME transformix)
-target_compile_definitions(transformix_exe PRIVATE ELX_CMAKE_VERSION="${CMAKE_VERSION}")
-target_link_libraries(transformix_exe ${ELASTIX_TARGET_LINK_LIBRARIES})
+if(NOT WASI AND NOT EMSCRIPTEN)
```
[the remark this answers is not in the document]
thewtex: OK, but should this be `if(ELASTIX_EXECUTABLE_BUILDS)`?

Core/CMakeLists.txt
thewtex: Yes, I can do that.

Common/itkComputeImageExtremaFilter.hxx
N-Dekker: `std::lock_guard` should be able to support any "mutex-like" type, that has a `lock()` and an `unlock()` member function. So for example, would the following code compile on WASI? https://godbolt.org/z/oP5T6Pn35  Can you please try that out?

```
#define ELX_SINGLE_THREADED // For WebAssembly Support

#include <mutex>

// Mutex for single-threading, does nothing.
class SingleThreadedMutex
{
  public:
    void lock() {}
    void unlock() {}
};

// The MutexType type alias:
#ifdef ELX_SINGLE_THREADED
using MutexType = SingleThreadedMutex;
#else
using MutexType = std::mutex;
#endif

// Use case:
MutexType mutexVariable;
std::lock_guard<MutexType> lockGuard(mutexVariable);

```

Core/CMakeLists.txt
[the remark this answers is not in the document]
N-Dekker: Update: The CMake option `ELASTIX_BUILD_EXECUTABLE` has just been added to the main branch! With pull request #922 commit 9d76b86f955043e2ae17bbbd621aaa5f0f60e001  @thewtex  Please rebase your wasm branch, and remove these two commits:

  - COMP: Do not build elastix and transformix exe with Wasm
  - COMP: Provide default ELX_CMAKE_VERSION

"COMP: Provide default ELX_CMAKE_VERSION" should no longer be necessary, because `ELX_CMAKE_VERSION` was only used when building the executables.

Common/itkComputeImageExtremaFilter.hxx
N-Dekker: @thewtex Would it be helpful to you if I would implement such a `SingleThreadedMutex` for elastix, that would replace `std::mutex` when a CMake option like `ELX_SINGLE_THREADED` would be ON?

thewtex:
Rebasing on `main`

Core/ComponentBaseClasses/elxResamplerBase.hxx
```diff
@@ -22,7 +22,6 @@
 #include "elxConversion.h"
 #include "elxDeref.h"
 
-#include "itkImageFileCastWriter.h"
```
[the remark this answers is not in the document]
thewtex: The way the include was added causes build errors when the include needs to avoided and causes build errors. This is a bug. `BUG` here does not refer to the use case referenced.

Core/CMakeLists.txt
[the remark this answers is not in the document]
thewtex: Done

thewtex:
Rebase

thewtex:
Superceded by #1331

PR #909, opened 2023-06-13
ENH: ElastixRegistrationMethod writes initial transform parameter files

N-Dekker:
Let ElastixRegistrationMethod write the transform maps from its InitialTransformParameterObject to the output directory.

PR #924, opened 2023-06-23
Add `TransformParameterFileName` property to itk::TransformixFilter

N-Dekker:
This pull request is basically complementary to improvements in `itk::ElastixRegistrationMethod`:
 - pull request #907 (Allow InitialTransformParameterFileName relative to parameter file)
 - pull request #909 (ElastixRegistrationMethod writes initial transform parameter files)

It allows `itk::TransformixFilter` to read back the transform parameter files written by `itk::ElastixRegistrationMethod` (as well as those written by the elastix executable)


N-Dekker:
@N-Dekker TODO Document the effect of calling _both_ SetTransformParameterFileName and SetTransformParameterObject on the same TransformixFilter instance. Discussed with Marius (@mstaring): the current effect is correct: one call "overrules" the other.

PR #933, opened 2023-07-11
BUG: Fix reference count in ConvertItkTransformBaseToSingleItkTransform

N-Dekker:
itk::ObjectFactoryBase::CreateInstance appears to return an object whose reference count is 2, while the object returned by ConvertItkTransformBaseToSingleItkTransform should have reference count 1.

Related to commit https://github.com/InsightSoftwareConsortium/ITK/commit/8e3c544ae5b331577f3b7c923886d1de77d5fd8e "BUG: The ability to override classes with factories was not working", Bill Lorensen, November 8, 2007.

----

@thewtex @tbirdso Do I understand correctly that after a successful `instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str())` call, `instance` has a reference count of `2` (as returned by `instance->GetReferenceCount()`), while it should have reference count = `1`?

Do you agree that the following `while` loop is recommended after doing such a `itk::ObjectFactoryBase::CreateInstance` call?

    while (instance->GetReferenceCount() > 1)
    {
      instance->UnRegister();
    }

### DIFF
--- a/Common/Transforms/elxTransformIO.cxx
+++ b/Common/Transforms/elxTransformIO.cxx
@@ -171,15 +171,20 @@ elastix::TransformIO::ConvertItkTransformBaseToSingleItkTransform(const itk::Tra
 
   const auto instanceName = className + "_double_" + std::to_string(elxTransform.GetInputSpaceDimension()) + '_' +
                             std::to_string(elxTransform.GetOutputSpaceDimension()) + optionalSplineOrderPostfix;
-  const auto instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str());
-  const auto itkTransform = dynamic_cast<itk::TransformBase *>(instance.GetPointer());
-  if (itkTransform == nullptr)
+  if (const auto instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str()))
   {
-    return nullptr;
+    // itk::ObjectFactoryBase::CreateInstance appears to return an object that has ReferenceCount == 2.
+    assert(instance->GetReferenceCount() == 2);
+    instance->UnRegister();
+
+    if (const auto itkTransform = dynamic_cast<itk::TransformBase *>(instance.GetPointer()))
+    {
+      itkTransform->SetFixedParameters(elxTransform.GetFixedParameters());
+      itkTransform->SetParameters(elxTransform.GetParameters());
+      return itkTransform;
+    }
   }
-  itkTransform->SetFixedParameters(elxTransform.GetFixedParameters());
-  itkTransform->SetParameters(elxTransform.GetParameters());
-  return itkTransform;
+  return nullptr;
 }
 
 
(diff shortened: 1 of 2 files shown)


tbirdso:
> @thewtex @tbirdso Do I understand correctly that after a successful instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str()) call, instance has a reference count of 2 (as returned by instance->GetReferenceCount()), while it should have reference count = 1?

Yes, I am able to reproduce:

```py
>>> fft_filter = itk.ObjectFactoryBase.CreateInstance('class itk::ForwardFFTImageFilter<class itk::Image<double,2>,class itk::Image<class std::complex<double>,2> >')
>>> print(fft_filter)
VkForwardFFTImageFilter (000001F6BB8F94A0)
  RTTI typeinfo:   class itk::VkForwardFFTImageFilter<class itk::Image<double,2>,class itk::Image<class std::complex<double>,2> >
  Reference Count: 2
  Modified Time: 429
...
```

ITK in practice may use the object factory implicitly through `itkNewMacro` which internally handles decrementing the reference count (xref https://github.com/InsightSoftwareConsortium/ITK/blob/1551ff3f2306346dae7b385680277e75580acf09/Modules/Core/Common/include/itkMacro.h#L315):
```py
>>> fft_filter = itk.ForwardFFTImageFilter[itk.Image[itk.F,2],itk.Image[itk.complex[itk.F],2]].New()
>>> print(fft_filter)
VkForwardFFTImageFilter (000001F6BB8FA850)
  RTTI typeinfo:   class itk::VkForwardFFTImageFilter<class itk::Image<float,2>,class itk::Image<class std::complex<float>,2> >
  Reference Count: 1
  Modified Time: 416
...
```

Some discussion around smart pointers and the object factory previously occurred in this closed PR.  https://github.com/InsightSoftwareConsortium/ITK/pull/2744

N-Dekker:
Thanks Tom! Very helpful insight information 😃 I just submitted a pull request to document that the object returned by CreateInstance has a reference count of 2, instead of 1. Please check: 
- https://github.com/InsightSoftwareConsortium/ITK/pull/4104

N-Dekker:
@tbirdso I guess the `while` loop is a bit overdone in my elastix PR, right? As I did:

    while (instance->GetReferenceCount() > 1)
    {
      instance->UnRegister();
    }

Instead it should be sufficient to just call `instance->UnRegister()` once, unconditionally, directly after calling `CreateInstance` right?

Common/Transforms/elxTransformIO.cxx
```diff
@@ -171,15 +171,20 @@ elastix::TransformIO::ConvertItkTransformBaseToSingleItkTransform(const itk::Tra
...
                             std::to_string(elxTransform.GetOutputSpaceDimension()) + optionalSplineOrderPostfix;
-  const auto instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str());
-  const auto itkTransform = dynamic_cast<itk::TransformBase *>(instance.GetPointer());
-  if (itkTransform == nullptr)
+  if (const auto instance = itk::ObjectFactoryBase::CreateInstance(instanceName.c_str()))
   {
-    return nullptr;
+    // itk::ObjectFactoryBase::CreateInstance appears to return an object that has ReferenceCount == 2.
+    assert(instance->GetReferenceCount() == 2);
+    instance->UnRegister();
```
N-Dekker, line 178: Update: just made the assumption that the reference count of an object returned by CreateInstance is always `2` more explicit by an `assert`. Hope the assumption is indeed always correct!

Issue #665, opened 2022-06-16
What to do about the outputpoints.txt file format?


N-Dekker:
The outputpoints.txt file format (produced by `transformix -def inputPoints.txt...`) appears a bit too verbose, and old-fashioned. For example:

```
Point	0	; InputIndex = [ 0 0 ]	; InputPoint = [ 0.000000 0.000000 ]	; OutputIndexFixed = [ 1 -2 ]	; OutputPoint = [ 1.000000 -2.000000 ]	; Deformation = [ 1.000000 -2.000000 ]
Point	1	; InputIndex = [ 0 1 ]	; InputPoint = [ 0.000000 1.000000 ]	; OutputIndexFixed = [ 1 -1 ]	; OutputPoint = [ 1.000000 -1.000000 ]	; Deformation = [ 1.000000 -2.000000 ]
Point	2	; InputIndex = [ 1 0 ]	; InputPoint = [ 1.000000 0.000000 ]	; OutputIndexFixed = [ 2 -2 ]	; OutputPoint = [ 2.000000 -2.000000 ]	; Deformation = [ 1.000000 -2.000000 ]
Point	3	; InputIndex = [ 1 1 ]	; InputPoint = [ 1.000000 1.000000 ]	; OutputIndexFixed = [ 2 -1 ]	; OutputPoint = [ 2.000000 -1.000000 ]	; Deformation = [ 1.000000 -2.000000 ]

```

Is it still relevant nowadays? If so could it be made slightly more compact? For example more like, starting with a header, having the names of the columns tab separated:

```
Point	InputIndex	InputPoint	OutputIndexFixed	OutputIndexPoint	Deformation
0	[ 0 0 ]	[ 0.0 0.0 ]	[ 1 -2 ]	[ 1.0 -2.0 ]	[ 1.0 -2.0 ]
1	[ 0 1 ]	[ 0.0 1.0 ]	[ 1 -1 ]	[ 1.0 -1.0 ]	[ 1.0 -2.0 ]
2	[ 1 0 ]	[ 1.0 0.0 ]	[ 2 -2 ]	[ 2.0 -2.0 ]	[ 1.0 -2.0 ]
3	[ 1 1 ]	[ 1.0 1.0 ]	[ 2 -1 ]	[ 2.0 -1.0 ]	[ 1.0 -2.0 ]
```

[...]

PR #943, opened 2023-08-02
ENH: Support for binary point file (.bin) reading and writing

ChristophKirst:
Here we provide support for a binary point file reader and writer to elastix to enable transformations of a large number of points (millions or more) for which the text format is not feasible.

This is based on the discussion in issue #665

The pull request provides the following:

  * Binary file reading and writing is detected  if the -def file ends with extension 'bin'.

  * The binary serial format is assumed to be:
     isindex(uint64), numberofpoints(uint64), points(nx3, double)

   * Only the transformed points are written to get maximal performance.

We would be very happy to discuss any changes to this strategy that might fit your overall strategy better. 
A binary read/writing option is essential for us. If it was integrated to your repository it would allow us elastix in our project rather then shipping our own patched version with our software.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ CMakeFiles/
 CTestTestfile.cmake
 .idea/
 .githooks/clangFormatMac
+/cmake-build-debug/
+/build/
--- a/Common/itkTransformixInputPointFileReader.hxx
+++ b/Common/itkTransformixInputPointFileReader.hxx
@@ -40,30 +40,60 @@ TransformixInputPointFileReader<TOutputMesh>::GenerateOutputInformation()
   {
     this->m_Reader.close();
   }
-  this->m_Reader.open(this->m_FileName);
 
-  /** Read the first entry */
-  std::string indexOrPoint;
-  this->m_Reader >> indexOrPoint;
+  if( this->m_Binary ) {
+    this->m_Reader.open(this->m_FileName, std::ios::binary);
 
-  /** Set the IsIndex bool and the number of points.*/
-  if (indexOrPoint == "point")
-  {
-    /** Input points are specified in world coordinates. */
-    this->m_PointsAreIndices = false;
-    this->m_Reader >> this->m_NumberOfPoints;
+    /** Read index flag */
+    unsigned long isindex;
+    this->m_Reader.read((char*) &isindex, sizeof(isindex));
+
+    if( isindex == 0 )
+    {
+      /** Input points are specified in world coordinates. */
+      this->m_PointsAreIndices = false;
+    }
+    else //if( indexOrPoint == 1 )
+    {
+      /** Input points are specified as image indices. */
+      this->m_PointsAreIndices = true;
+    }
+
+    /** Read number of points */
+    unsigned long nrofpoints;
+    this->m_Reader.read((char*) &nrofpoints, sizeof(nrofpoints));
+    this->m_NumberOfPoints = nrofpoints;
+
+    /** Leave the file open for the generate data method */
+
   }
-  else if (indexOrPoint == "index")
+  else //if( this->m_Binary )
   {
-    /** Input points are specified as image indices. */
-    this->m_PointsAreIndices = true;
-    this->m_Reader >> this->m_NumberOfPoints;
-  }
-  else
-  {
-    /** Input points are assumed to be specified as image indices. */
-    this->m_PointsAreIndices = true;
-    this->m_NumberOfPoints = atoi(indexOrPoint.c_str());
+    this->m_Reader.open(this->m_FileName);
+
+    /** Read the first entry */
+    std::string indexOrPoint;
+    this->m_Reader >> indexOrPoint;
+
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -38,3 +38,5 @@ CMakeFiles/
 CTestTestfile.cmake
 .idea/
 .githooks/clangFormatMac
+/cmake-build-debug/
+/build/
```
N-Dekker, line 42: I would like to see this as a separate commit. (Following the principle of "One Commit. One Change.") Why do those directories appear in your source tree? Are they specific to a particular tool or platform? I'm fine with the change itself, but I would like some clarification!

Common/itkTransformixInputPointFileReader.hxx
```diff
@@ -40,30 +40,60 @@ TransformixInputPointFileReader<TOutputMesh>::GenerateOutputInformation()
   {
     this->m_Reader.close();
   }
-  this->m_Reader.open(this->m_FileName);
 
-  /** Read the first entry */
-  std::string indexOrPoint;
-  this->m_Reader >> indexOrPoint;
+  if( this->m_Binary ) {
```
N-Dekker, line 44: We use automatic code formatting by clang-format. It looks like this code still needs to be formatted that way, as I would expect it to place the `{` on the next line. (We use a specific (old) version of clang-format: [8.0.1](https://data.kitware.com/#collection/57b5c9e58d777f126827f5a1/folder/5d635cbcd35580e6dcbd80b6)) No problem, of course, we can fix that later.

Common/itkTransformixInputPointFileReader.hxx
```diff
@@ -40,30 +40,60 @@ TransformixInputPointFileReader<TOutputMesh>::GenerateOutputInformation()
...
+    this->m_Reader.open(this->m_FileName, std::ios::binary);
+
+    /** Read index flag */
+    unsigned long isindex;
```
N-Dekker, line 48: I'm sorry to say the use of (signed or unsigned) `long` is problematic, because its size differs between currently supported platforms and compilers. On Linux/GCC it's typically 64 bits, whereas on Windows/MSVC, it's 32 bits.  If you want to write unsigned integers on one platform and read them back on another, better use either `uint32_t` or `uint64_t`. (In general use cases, `unsigned int` is usually fine, as it is the "default" unsigned type, and it's 32-bits on all supported platforms.)

Common/itkTransformixInputPointFileReader.hxx
```diff
@@ -40,30 +40,60 @@ TransformixInputPointFileReader<TOutputMesh>::GenerateOutputInformation()
...
+    this->m_Reader.read((char*) &isindex, sizeof(isindex));
```
N-Dekker, line 49: Instead of C-style `(char*)` cast, please use `reinterpret_cast<const char *>`. C-style casts are generally considered unsafe.

N-Dekker:
Some tests should be still added. I'm not sure if they can be written as small GoogleTest unit tests (testing the extra TransformixInputPointFileReader functionality at C++ class level), or only as elastix executable tests. But I would like to see tests for various number of points (N = 0, N = 1, N = 2), as well as various coordinate values. Still to be discussed, maybe.

N-Dekker:
@ChristophKirst Is the binary file format already in use outside of your (patched) elastix version? Otherwise we would prefer to support a more commonly used file format. Would the ".vtk" file format (from vtk.org) also be of interest to you?

.gitignore
ChristophKirst: Hi,  sorry for the late reply due to other obligations.
This can be safely ignored, i did not realize this change in .gitignore

ChristophKirst:
@N-Dekker sorry for the late reply due to many other obligations. 
We currently use the patched version with our software ClearMap to align cleared tissues, and it would be much easier to link to your repository rather than ship ClearMap with our patched elastix version.
That said, I have not looked into vtk files, is there support for fast writing and reading of millions of points via vtk files in elastix / transformix ?

ChristophKirst:
@N-Dekker sorry for the late reply.

We do use the patched version with our software ClearMap. It owuld be much easier to link your repository directly than using a patched version.

That said, I have not looked into vtk files. Does elastix support fast reading of millions of points from that format ?

ChristophKirst:
@N-Dekker I have checked, the transformix seems to read a vtk binary files, but will generates a text based vtk output file. 
I could try to adapt elastix so one can choose the output file as text or binary ? 
On the sort run, is it possible to adapt the solution above with the fixes you suggested about avoiding long data type?

N-Dekker:
@ChristophKirst Thanks for your continuing interest  😃 

> the transformix seems to read vtk binary files, but will generate a text based vtk output file, which is not practical in our case.

I think I can fix that, by a few code adjustments 😃 What if transformix would generate a text based vtk output file, _if_ its input vtk file would be binary? Would that solve your problem?

And by the way, do you prefer 32-bit float or 64-bit double precision for the coordinates of the points?

Issue #874, opened 2023-04-14
Allow ITK `DisplacementFieldTransform` as input to library interface


N-Dekker:
ITK's [DisplacementFieldTransform](https://itk.org/Doxygen53/html/classitk_1_1DisplacementFieldTransform.html) should probably be converted internally to an elastix [DeformationFieldTransform](https://github.com/SuperElastix/elastix/blob/c3d478e8e3b6b1806862abbd0d777beff9b66346/Components/Transforms/DeformationFieldTransform/elxDeformationFieldTransform.h), which internally has a [DeformationFieldInterpolatingTransform](https://github.com/SuperElastix/elastix/blob/main/Components/Transforms/DeformationFieldTransform/itkDeformationFieldInterpolatingTransform.h) 

Requested by Konstantinos (@ntatsisk)

N-Dekker:
Should be addressed by supporting external transforms, by the following pull requests:
- #944 
- #946 
- #948 
- #949

PR #944, opened 2023-08-10
Add "ExternalTransform" component and `ElastixRegistrationMethod::SetExternalInitialTransform`

N-Dekker:
Adds a component that allows specifying an ITK transform by its address in memory. The transform is then _not_ converted to the corresponding elastix AdvancedTransform. Instead, it is only "adapted", meaning that its `TransformPoint` is still called directly, when using the transform. This component is only meant to be used with the elastix/transformix library, not with the elastix/transformix executable. It only serves as input transform, so it is meant not to be adjusted during registration.

Adds a `SetExternalInitialTransform` member function to `ElastixRegistrationMethod`, allowing to specify any ITK transform (for example an `itk::DisplacementFieldTransform`) as initial transform. Aims to address the following issue:
- #874

### DIFF
--- a/Common/GTesting/elxConversionGTest.cxx
+++ b/Common/GTesting/elxConversionGTest.cxx
@@ -444,6 +444,23 @@ GTEST_TEST(Conversion, ToString)
 }
 
 
+// Tests that conversion of itk::Object pointers to string and back is lossless.
+GTEST_TEST(Conversion, LosslessRoundTripOfObjectPointers)
+{
+  const auto expectLosslessRoundTrip = [](const itk::Object * const ptr) {
+    const itk::Object * actualPtr{};
+    EXPECT_TRUE(Conversion::StringToValue(Conversion::ObjectPtrToString(ptr), actualPtr));
+    EXPECT_EQ(actualPtr, ptr);
+  };
+
+  expectLosslessRoundTrip(nullptr);
+  expectLosslessRoundTrip(itk::Object::New());
+
+  const elx::DefaultConstruct<itk::Object> localVariable{};
+  expectLosslessRoundTrip(&localVariable);
+}
+
+
 GTEST_TEST(Conversion, ToVectorOfStrings)
 {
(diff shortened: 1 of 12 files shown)


N-Dekker:
The GitHub Actions CI failures appear unrelated to this PR:
```

ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    SimpleITK from https://files.pythonhosted.org/packages/d4/e8/f37a48af6c4df67a8bbbbab6f21cb0a30ae024d383a3c2a4d76f90e29e46/SimpleITK-2.2.1-cp37-cp37m-win_amd64.whl (from -r D:\a\elastix\elastix/Elastix-source/Testing/PythonTests/requirements.txt (line 2)):
        Expected sha256 4878f5c032d638d7833901d8d3f79d4c6ed90d041d1fcc1b074dd6e4b7c9abc2
             Got        93b4fffc1582d3d799d0f808a5469576c963eeb8533bc35b0d4d3385182d0f27
```

N-Dekker:
Discussed with @mstaring and @stefanklein at the internal weekly elastix sprint of last Monday, 14 August.

N-Dekker:
@mstaring @stefanklein Now that we have `ElastixRegistrationMethod::SetExternalInitialTransform`, would it be useful to add a "SetExternalTransform" member function to `TransformixFilter` as well? TransformixFilter already has a SetTransform that accepts an ITK transform as input. But _only_ if it can be converted to an existing elastix component (Translation, Affine, Euler, Similarity, Bspline). `TransformixFilter::SetExternalTransform` would support _any arbitrary_ ITK transform.

stefanklein:
Yes, that would be useful I think!

mstaring:
good idea!
________________________________

N-Dekker:
Here it is, the `TransformixFilter::SetExternalTransform` pull request:
 - #948

PR #945, opened 2023-08-10
STYLE: Let GetNthTransform, GetCombinationTransform return a raw pointer

N-Dekker:
In general, the use of a smart pointer as return type allows extending the lifetime of the returned object beyond the function call. But as these two `ElastixRegistrationMethod` member functions both just return a pointer to a transform object that is "owned" by the `ElastixRegistrationMethod` object, the caller of `GetNthTransform` or `GetCombinationTransform` does not need to extend the lifetime of the returned transform object. This returned transform object will stay alive while the `ElastixRegistrationMethod` object is there, even when the return type is a raw pointer.

This adjustment was already proposed by Matt McCormick (@thewtex).

----

Aims to (at least partially) supersede the following pull request: 
- #914 

This little pull request is there just to ensure that these changes to  `GetNthTransform` or `GetCombinationTransform` can already be included with the next ITKElastix release.

### DIFF
--- a/Core/Main/itkElastixRegistrationMethod.h
+++ b/Core/Main/itkElastixRegistrationMethod.h
@@ -273,11 +273,11 @@ public:
   GetNumberOfTransforms() const;
 
   /** Returns the nth transformation, produced during the last Update(). */
-  SmartPointer<TransformType>
+  TransformType *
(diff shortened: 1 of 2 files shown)


thewtex (approved):
Thanks @N-Dekker

PR #946, opened 2023-08-17
ENH: Write external initial transforms to ITK transform file

N-Dekker:
When `ElastixRegistrationMethod` writes an external initial transform (as specified by `SetExternalInitialTransform`) to its output directory, it will now write the transform to a separate ITK transform file. By default, it will use ITK's TFM file format, but it can be set by the "ITKTransformOutputFileNameExtension" parameter.


N-Dekker:
Discussed with Marius (@mstaring), ready to be merged.

PR #948, opened 2023-08-18
ENH: Add `SetExternalTransform` to TransformixFilter

N-Dekker:
Including GoogleTest unit test.

### DIFF
--- a/Core/Main/itkTransformixFilter.hxx
+++ b/Core/Main/itkTransformixFilter.hxx
@@ -237,6 +237,18 @@ TransformixFilter<TImage>::GenerateData()
     }
   }
 
+  if (m_ExternalTransform)
+  {
+    // External transforms should use "ResampleInterpolator" and the output image domain specification
+    // (Size/Spacing/Origin/Index/Direction) of the last transform parameter map.
+    auto transformParameterMap = std::move(transformParameterMapVector.back());
+    SetParameterValueAndWarnOnOverride(transformParameterMap, "Transform", "ExternalTransform");
+    SetParameterValueAndWarnOnOverride(
+      transformParameterMap, "TransformAddress", elx::Conversion::ObjectPtrToString(m_ExternalTransform));
+    transformParameterMapVector.clear();
+    transformParameterMapVector.push_back(std::move(transformParameterMap));
+  }
+
   const auto movingImageDimensionString = std::to_string(MovingImageDimension);
   const auto movingImagePixelTypeString = elx::PixelTypeToString<typename TImage::PixelType>();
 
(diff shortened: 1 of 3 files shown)


Core/Main/itkTransformixFilter.hxx
```diff
@@ -237,6 +237,14 @@ TransformixFilter<TImage>::GenerateData()
     }
   }
 
+  if (m_ExternalTransform)
+  {
+    auto & transformParameterMap = transformParameterMapVector.front();
+    SetParameterValueAndWarnOnOverride(transformParameterMap, "Transform", "ExternalTransform");
+    SetParameterValueAndWarnOnOverride(
+      transformParameterMap, "TransformAddress", elx::Conversion::ObjectPtrToString(m_ExternalTransform));
```
N-Dekker, line 245: Note that for TransformixFilter, a specified external ITK transform does not _replace_ the parameter maps. Instead, its "TransformAddress" is just added to the first transform parameter map.

Transform parameter maps may have extra information that are not represented by an external ITK transform, specifically the "ResampleInterpolator" and the output image domain specification (Size/Spacing/Origin/Index/Direction)

mstaring: ok, but does it replace the previous transform information? Here it should not act as an initial transform, just as the main transform

N-Dekker: > does it replace the previous transform information? 

@mstaring When the user has provided a _single_ transform parameter map, it does. When the user has provided a vector of _multiple_ transform parameter maps, it only replaces the transform information of the first map. Is that OK?

In either case, it will always replace the transform specified by a previous call to the old `SetTransform` member function.

mstaring: Then it would act as the initial transform of the other transformations. The expected behaviour for transformix is that the transform that you set, is the one that is evaluated. So, that would mean deleting the other transforms in the vector of parameter maps, and only keep the first (or last?) one and replace it with the ExternalTransform

N-Dekker: > that would mean deleting the other transforms in the vector of parameter maps, and only keep the first (or last?) one and replace it with the ExternalTransform

Maybe then it would be best to only keep the last map from the transform parameter maps. Because that's the one that should contain the "ResampleInterpolator" and the output image domain specification (Size/Spacing/Origin/Index/Direction), right?

Another option could be to require exactly 1 map. And throw an exception when the user-specified vector of transform parameter maps has more than 1 item. What do you think?

Core/Main/itkTransformixFilter.hxx
```diff
@@ -237,6 +237,18 @@ TransformixFilter<TImage>::GenerateData()
...
+  if (m_ExternalTransform)
+  {
+    // External transforms should use "ResampleInterpolator" and the output image domain specification
+    // (Size/Spacing/Origin/Index/Direction) of the last transform parameter map.
+    auto transformParameterMap = std::move(transformParameterMapVector.back());
+    SetParameterValueAndWarnOnOverride(transformParameterMap, "Transform", "ExternalTransform");
+    SetParameterValueAndWarnOnOverride(
+      transformParameterMap, "TransformAddress", elx::Conversion::ObjectPtrToString(m_ExternalTransform));
+    transformParameterMapVector.clear();
+    transformParameterMapVector.push_back(std::move(transformParameterMap));
```
N-Dekker, line 249: @mstaring  Update: if an external transform is specified, only the _last_ map of the vector of transform parameter maps will now be taken into account. If the specified transform parameter object has `N` maps, the first `N - 1` maps will be ignored. OK?

N-Dekker:
@mstaring I think it's OK now. It will now only use the last map of the specified vector of transform parameter maps, and combine that with the specified external transform. Will merge before the lunch  😃

PR #949, opened 2023-08-22
`ElastixRegistrationMethod.ConvertToItkTransform` support external initial transforms 

N-Dekker:
When `ConvertToItkTransform` has converted the CombinationTransform produced by a registration that used an external initial transform, the result should be an `itk::CompositeTransform` that has a pointer to the initial transform as its "back transform".

Follow-up on a comment by Dženan Zukić (@dzenanz) at https://github.com/InsightSoftwareConsortium/ITKElastix/pull/241#issuecomment-1682550470


dzenanz (commented):
Looks good on a glance.

PR #963, opened 2023-09-18
COMP: Replace include <itkTransformFileWriter.h> with "elxTransformIO.h"

N-Dekker:
Aims to avoid an ITKElastix compile error reported by Dženan Zukić (@dzenanz), saying:

> _deps/elx-src/Core/Main/itkElastixRegistrationMethod.hxx:43:10: fatal error: 'itkTransformFileWriter.h' file not found


dzenanz (approved):
LGTM

Issue #965, opened 2023-09-25
ElastixRegistrationMethod GetNumberOfTransforms, GetCombinationTransform, etc. does not work with non-float ImageType


thewtex:
When the ElastixRegistrationMethod ImageType template parameters do not have a float pixel type, the dynamic_cast in these methods fails.

This can be reproduced by, for example, removing the cast of the input images to float in this example (Cell 2): https://github.com/InsightSoftwareConsortium/ITKElastix/blob/main/examples/ITK_Example22_ConvertToITKTransform.ipynb

PR #966, opened 2023-09-27
BUG: Fix ElastixRegistrationMethod GetNumberOfTransforms GetNthTransform

N-Dekker:
ElastixRegistrationMethod member functions `GetNumberOfTransforms`, `GetNthTransform`, and `GetCombinationTransform` did always just return zero or null, erroneously, when the pixel type of the input images (specified by `TFixedImage` and `TMovingImage`) would be different from the internal pixel type (specified by "FixedInternalImagePixelType" and "MovingInternalImagePixelType", `float` by default). The code incorrectly assumed that the objects stored in the ElastixBase `TransformContainer` are always of type `elx::TransformBase<ElastixTemplate<TFixedImage, TMovingImage>>`.

This commit just assumes that the objects stored in `TransformContainer` are of type `AdvancedCombinationTransform<double, FixedImageDimension>`.

The commit aims to fix issue https://github.com/SuperElastix/elastix/issues/965 "ElastixRegistrationMethod GetNumberOfTransforms, GetCombinationTransform, etc. does not work with non-float ImageType", reported by Matt McCormick (@thewtex).

Extended the GoogleTest `itkElastixRegistrationMethod.CheckMinimumMovingImageUsingAnyInternalPixelType` unit test to check this issue.

----

@mstaring @stefanklein This fix depends on the assumption that a transform from the ElastixBase `TransformContainer` is always a "combination transform". That's a correct assumption, right? `ElastixBase::m_TransformContainer` is declared as a generic `ObjectContainerPointer` (but it seems to me that it only contains combination transforms):

https://github.com/SuperElastix/elastix/blob/03bf07851c95f9e8a771ff9ba344feae8f1f5d6f/Core/Kernel/elxElastixBase.h#L505

### DIFF
--- a/Core/Main/itkElastixRegistrationMethod.hxx
+++ b/Core/Main/itkElastixRegistrationMethod.hxx
@@ -979,17 +979,8 @@ template <typename TFixedImage, typename TMovingImage>
 unsigned int
 ElastixRegistrationMethod<TFixedImage, TMovingImage>::GetNumberOfTransforms() const
 {
-  const auto * const transformContainer = m_ElastixMain->GetElastixBase().GetTransformContainer();
-
-  if ((transformContainer == nullptr) || transformContainer->empty())
-  {
-    return 0;
-  }
-
-  const auto * const elxTransformBase =
-    dynamic_cast<ElastixTransformBaseType *>(transformContainer->front().GetPointer());
-
-  return (elxTransformBase == nullptr) ? 0 : elxTransformBase->GetAsITKBaseType()->GetNumberOfTransforms();
+  const auto advancedCombinationTransform = GetAdvancedCombinationTransform();
+  return advancedCombinationTransform ? advancedCombinationTransform->GetNumberOfTransforms() : 0;
 }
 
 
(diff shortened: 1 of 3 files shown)


Core/Main/itkElastixRegistrationMethod.hxx
```diff
@@ -979,60 +979,47 @@ template <typename TFixedImage, typename TMovingImage>
...
+{
   {
+    if (m_ElastixMain)
+    {
+      if (const auto transformContainer = m_ElastixMain->GetElastixBase().GetTransformContainer();
+          (transformContainer != nullptr) && !transformContainer->empty())
+      {
+        if (const auto firstTransform = transformContainer->front().GetPointer())
+        {
+          return static_cast<AdvancedCombinationTransformType *>(firstTransform);
```
N-Dekker, line 1009: I assume the `static_cast<AdvancedCombinationTransformType *>(firstTransform)` is OK here, casting the first element from `ElastixBase::m_TransformContainer`.

thewtex:
@N-Dekker thank you!!

Issue #904, opened 2023-06-02
ParameterObject::WriteParameterFiles does not write the first map!


N-Dekker:
It appears that `ParameterObject::WriteParameterFiles` skips the first map from the specified `parameterMapVector`: https://github.com/SuperElastix/elastix/blob/81324cf0158256ae48a14762514ce60e786ac485/Core/Main/elxParameterObject.cxx#L379-L401

The bug appears introduced with pull request #50 commit 9ba16836e1a0642ecd2006a70ce3886b1fa4720e "ENH: Add convenience methods for writing parameter files", merged on March 20, 2018, when the member function became as follows: https://github.com/SuperElastix/elastix/blob/9ba16836e1a0642ecd2006a70ce3886b1fa4720e/Core/Main/elxParameterObject.cxx#L256-L280

FYI @kaspermarstal

thewtex:
I also encountered this issue:

https://github.com/InsightSoftwareConsortium/ITKElastix/blob/29ce6866414070136bb5bc2f70129eafcdde7040/wasm/write-parameter-files.cxx#L74-L76

N-Dekker:
@dzenanz @thewtex Please check:

- https://github.com/SuperElastix/elastix/pull/967

N-Dekker:
> I also encountered this issue:
> https://github.com/InsightSoftwareConsortium/ITKElastix/blob/29ce6866414070136bb5bc2f70129eafcdde7040/wasm/write-parameter-files.cxx#L74-L76

@thewtex Thanks Matt! Please remember that these two lines should be removed, once ITKElastix has been updated to include elastix PR #967 (commit a57470b0ccfcab925e101f6e6577abaf2074574a):

    // The initial is not written by WriteParameterFiles
    ITK_WASM_CATCH_EXCEPTION(pipeline, elastix::ParameterObject::WriteParameterFile(parameterMaps[0], parameterFiles[0]));

PR #967, opened 2023-09-27
BUG: `ParameterObject::WriteParameterFiles` should write all maps

N-Dekker:
With this commit `ParameterObject::WriteParameterFiles` writes _each_ maps to a file, including the first one. Aims to address issue https://github.com/SuperElastix/elastix/issues/904 "ParameterObject::WriteParameterFiles does not write the first map!"

A GoogleTest unit test is added, testing that each of the files specified by the `parameterFileNameVector` parameter is indeed written to disk.

### DIFF
--- a/Core/Main/GTesting/ParameterObjectGTest.cxx
+++ b/Core/Main/GTesting/ParameterObjectGTest.cxx
@@ -1,0 +1,75 @@
+/*=========================================================================
+ *
+ *  Copyright UMC Utrecht and contributors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


Core/Main/GTesting/ParameterObjectGTest.cxx
```diff
@@ -0,0 +1,75 @@
...
+    {
+      fileNames.push_back(outputDirectoryPath + '/' + "ParameterFile." + std::to_string(i) + ".txt");
+    }
+
+    parameterObject.WriteParameterFiles(fileNames);
+
+    // Check that each of the specified files is written to disk.
+    for (const auto & fileName : fileNames)
+    {
+      EXPECT_TRUE(itksys::SystemTools::FileExists(fileName.c_str(), true));
```
dzenanz, line 72: Did elastix switch to C++17 yet? If so, `std::filesystem::exists` could be used here.

dzenanz (approved):

Core/Main/GTesting/ParameterObjectGTest.cxx
N-Dekker: Thanks for asking! The elastix main branch has switched to C++17, but I'm still reluctant to use std::filesystem, because ITK does not yet use std::filesystem either. As we discussed at https://discourse.itk.org/t/shall-we-start-using-c-17-at-the-master-branch/5828, March 2023. (std::filesystem would require at least GCC version 8). Of course I'm glad to hear if it's OK by now for ITK to require at least GCC 8. 😃

PR #973, opened 2023-10-13
ENH: ImageRandomSampler `UseMultiThread` now using ITK's ThreadPool

N-Dekker:
Internally calling `multiThreader->SingleMethodExecute()`, instead of overriding `ThreadedGenerateData`.

When 10'000 random samples are retrieved, a significant performance improvement is observed by enabling `UseMultiThread`: from ~1.2 millisecond (single-threaded) to ~0.9 millisecond (multi-threaded). When 100'000 random samples are retrieved, even from ~7 to ~3 millisecond: Tested on a Windows 10 PC (6 cores, 12 logical processors), VS2019 Release build.

### DIFF
--- a/Common/GTesting/itkImageRandomSamplerGTest.cxx
+++ b/Common/GTesting/itkImageRandomSamplerGTest.cxx
@@ -87,3 +87,25 @@ GTEST_TEST(ImageRandomSampler, SetSeedMakesRandomizationDeterministic)
     EXPECT_EQ(generateSamples(), generateSamples());
   }
 }
+
+
+GTEST_TEST(ImageRandomSampler, HasSameOutputWhenUsingMultiThread)
+{
+  using PixelType = int;
+  using ImageType = itk::Image<PixelType>;
+  using SamplerType = itk::ImageRandomSampler<ImageType>;
+
+  const auto image =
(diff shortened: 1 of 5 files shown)


N-Dekker:
For the record, I tested the performance (single-threaded vs multi-threaded) as follows:

    using PixelType = int;
    using ImageType = itk::Image<PixelType, 3>;
    using SamplerType = itk::ImageRandomSampler<ImageType>;

    const auto image = CreateImageFilledWithSequenceOfNaturalNumbers<PixelType>(ImageType::SizeType::Filled(100));

    const auto generateSamples = [image](const bool useMultiThread, const SamplerType::SeedIntegerType seed) {
      elx::DefaultConstruct<SamplerType> sampler{};
      sampler.SetUseMultiThread(useMultiThread);
      sampler.SetSeed(1);
      sampler.SetNumberOfSamples(sampler.GetNumberOfSamples() * 10);
      sampler.SetInput(image);
      using namespace std::chrono;
      const auto timePoint = high_resolution_clock::now();
      sampler.Update();
      std::cout << sampler.GetNumberOfSamples() << " samples "
                << " Duration: " << duration_cast<duration<double>>(high_resolution_clock::now() - timePoint).count()
                << " seconds " << (useMultiThread ? "MultiThread" : "single threaded") << std::endl;
      return std::move(DerefRawPointer(sampler.GetOutput()).CastToSTLContainer());

PR #992, opened 2023-11-16
Replace "-mts" command-line option with `UseMultiThreadingForSamplers` parameter (default "true")

N-Dekker:
Replaced the "-mts" command-line option with a new elastix parameter, `UseMultiThreadingForSamplers` ("true" by default).


N-Dekker:
@mstaring @stefanklein This is just what we discussed in the sprint meetings this month, so I guess it's OK to you, right? Using an elastix parameter, "UseMultiThreadingForSamplers", instead of a command-line option.

PR #994, opened 2023-11-17
PERF: `ImageGridSampler` twice as fast, doing `sampleVector.reserve(numberOfSamplesOnGrid)`

N-Dekker:
Reserving the right number of samples beforehand appears to possibly make ImageGridSampler twice as fast. When having a 4096x4096 grid (no mask), it was observed that the duration of an ImageGridSampler::Update() went from 0.45 second (before this commit) to almost 0.20 second  (after this commit), using Visual C++ 2019, Release configuration.

This pull request also proposes to use std::accumulate to calculate `numberOfSamplesOnGrid`.

----

Note that originally `numberOfSamplesOnGrid` was just calculated without being used. From here, commit 6e0e6f88163dfce901d72bbcb57fb1a388186ac3, July 17, 2006, "implemented grid and full sampler":

https://github.com/SuperElastix/elastix/blob/6e0e6f88163dfce901d72bbcb57fb1a388186ac3/src/Common/ImageSamplers/itkImageGridSampler.txx#L37

### DIFF
--- a/Common/ImageSamplers/itkImageGridSampler.hxx
+++ b/Common/ImageSamplers/itkImageGridSampler.hxx
@@ -21,6 +21,7 @@
 #include "itkImageGridSampler.h"
 
 #include "itkImageRegionConstIteratorWithIndex.h"
+#include <cassert>
 
 namespace itk
 {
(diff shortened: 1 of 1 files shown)


N-Dekker:
This pull request yields a performance improvement of the "Grid" ImageSampler component, as well as other parts of the software, including `ComputeJacobianTerms::SampleFixedImageForJacobianTerms`, which uses ImageGridSampler:

https://github.com/SuperElastix/elastix/blob/c3fb0eba5644a2122e11bd25f5042e47d3b4ec71/Common/itkComputeJacobianTerms.hxx#L536

PR #1018, opened 2024-01-17
PERF: ImageGridSampler replace IsInsideInWorldSpace calls

N-Dekker:
Made `GenerateData()` around 2x as fast, when having a mask.

### DIFF
--- a/Common/ImageSamplers/itkImageGridSampler.hxx
+++ b/Common/ImageSamplers/itkImageGridSampler.hxx
@@ -241,6 +241,8 @@ ImageGridSampler<TInputImage>::SingleThreadedGenerateData(const TInputImage &   
 
   if (mask)
   {
+    const auto & worldToObjectTransform = elastix::Deref(mask->GetObjectToWorldTransformInverse());
+
     /* Ugly loop over the grid; checks also if a sample falls within the mask. */
     for (unsigned int t = 0; t < GetGridSizeValue<3>(gridSize); ++t)
     {
@@ -255,7 +257,9 @@ ImageGridSampler<TInputImage>::SingleThreadedGenerateData(const TInputImage &   
             // Translate index to point.
             inputImage.TransformIndexToPhysicalPoint(index, tempSample.m_ImageCoordinates);
 
-            if (mask->IsInsideInWorldSpace(tempSample.m_ImageCoordinates))
+            // Equivalent to `mask->IsInsideInWorldSpace(tempSample.m_ImageCoordinates)`, but much faster.
+            if (mask->IsInsideInObjectSpace(
+                  worldToObjectTransform.WorldToObjectTransformType::TransformPoint(tempSample.m_ImageCoordinates)))
             {
               // Get sampled fixed image value.
               tempSample.m_ImageValue = inputImage.GetPixel(index);
@@ -343,6 +347,7 @@ ImageGridSampler<TInputImage>::MultiThreadedGenerateData(MultiThreaderBase &    
 
   UserData userData{ inputImage,
                      mask,
+                     mask ? mask->GetObjectToWorldTransformInverse() : nullptr,
                      gridSpacing,
                      GenerateWorkUnits(numberOfWorkUnits, croppedInputImageRegion, gridIndex, gridSpacing, samples) };
 
@@ -440,6 +445,9 @@ ImageGridSampler<TInputImage>::ThreaderCallback(void * const arg)
   const auto * const mask = userData.Mask;
   assert((mask == nullptr) == (!VUseMask));
 
+  const auto * const worldToObjectTransform = userData.WorldToObjectTransform;
+  assert((worldToObjectTransform == nullptr) == (!VUseMask));
+
   const auto                gridSpacing = userData.GridSpacing;
   const SampleGridSizeType  gridSizeForThread = workUnit.GridSize;
   const SampleGridIndexType gridIndexForThread = workUnit.GridIndex;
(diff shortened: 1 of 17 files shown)


Common/ImageSamplers/itkImageGridSampler.hxx
```diff
@@ -255,7 +257,9 @@ ImageGridSampler<TInputImage>::SingleThreadedGenerateData(const TInputImage &
             // Translate index to point.
             inputImage.TransformIndexToPhysicalPoint(index, tempSample.m_ImageCoordinates);
 
-            if (mask->IsInsideInWorldSpace(tempSample.m_ImageCoordinates))
+            // Equivalent to `mask->IsInsideInWorldSpace(tempSample.m_ImageCoordinates)`, but much faster.
+            if (mask->IsInsideInObjectSpace(
+                  worldToObjectTransform.WorldToObjectTransformType::TransformPoint(tempSample.m_ImageCoordinates)))
```
N-Dekker, line 262: @mstaring Thanks for looking at the issue! The original elastix `itk::ImageGridSampler` (before this commit) called `SpatialObject::IsInsideInWorldSpace`, which is implemented by ITK, [as follows](https://github.com/InsightSoftwareConsortium/ITK/blob/0b47b7bf95665eb31e00442074a3a77e11e1f790/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx#L187-L194):

```
SpatialObject<TDimension>::IsInsideInWorldSpace(const PointType &   point,
                                                unsigned int        depth,
                                                const std::string & name) const
{
  PointType pnt;
  pnt = this->GetObjectToWorldTransformInverse()->TransformPoint(point);
  return IsInsideInObjectSpace(pnt, depth, name);
}

```

`this->GetObjectToWorldTransformInverse()` is a rather expensive call, so this PR takes that call out of the nested for loops. Also it saves time by skipping the `depth` and `name` arguments. And then it bypasses the virtual table, by doing a non-virtual call to `TransformPoint` (because the runtime-time of worldToObjectTransform is already known at compile-time).

But maybe I misunderstand your question...?  🤷

N-Dekker:
@mstaring @stefanklein Related to this PR: The mask used for ImageSamplers is specified by a spatial object:

https://github.com/SuperElastix/elastix/blob/1d30c13a3fbb1bd75e23038a38f8d7d537e30494/Common/ImageSamplers/itkImageSamplerBase.h#L84

In practice (in the context of elastix and ITKElastix), is this always an `ImageMaskSpatialObject`? Or can this still be "any other type of spatial object"? Specifically, could any of the following spatial object types possibly be useful when specifying a mask?

```
ArrowSpatialObject
BoxSpatialObject
EllipseSpatialObject
GaussianSpatialObject
GroupSpatialObject
ImageSpatialObject
MeshSpatialObject
PointBasedSpatialObject
FEMObjectSpatialObject
```

The performance may be further improved if it is safe to assume that the spatial object used as a mask is _always_ an `ImageMaskSpatialObject`.

mstaring:
yes in elastix it is always ImageMaskSpatialObject. (but briefly check by searching the code for spatialobject)

N-Dekker:
Benchmark:

```cpp
  using PixelType = int;
  constexpr auto Dimension = 2U;
  constexpr auto ImageSizeValue = 1U << 12U;
  using SamplerType = itk::ImageGridSampler<itk::Image<PixelType, Dimension>>;
  const ImageDomain<Dimension> imageDomain(itk::Size<Dimension>::Filled(ImageSizeValue));
  const auto                   image = CreateImage<PixelType>(imageDomain);

  using MaskSpatialObjectType = itk::ImageMaskSpatialObject<Dimension>;
  const auto maskImage = CreateImage<MaskSpatialObjectType::PixelType>(ImageDomain(*image));

  maskImage->SetPixel({}, 1);
  maskImage->SetPixel(itk::Index<Dimension>::Filled(ImageSizeValue - 1), 1);

  const auto maskSpatialObject = MaskSpatialObjectType::New();
  maskSpatialObject->SetImage(maskImage);
  maskSpatialObject->Update();

  const auto generateSamples = [image, maskSpatialObject]() {
    elx::DefaultConstruct<SamplerType> sampler{};

    sampler.SetInput(image);
    sampler.SetMask(maskSpatialObject);
    sampler.SetUseMultiThread(false);
    using namespace std::chrono;
    const auto timePoint = high_resolution_clock::now();
    sampler.Update();
    std::cout << " Duration: " << duration_cast<duration<double>>(high_resolution_clock::now() - timePoint).count()
              << " seconds" << std::endl;
    return std::move(DerefRawPointer(sampler.GetOutput()).CastToSTLContainer());
  };

  for (int i{}; i < 5; ++i)
  {
    EXPECT_EQ(generateSamples().size(), 2);
  }

```
Results (VS2022 Release):
 - Before this pull request: 0.510 sec
 - After replacing IsInsideInWorldSpace: 0.242 sec
 - After making IsInsideInObjectSpace calls non-virtual: 0.236 sec

Common/ImageSamplers/itkImageGridSampler.hxx
```diff
@@ -258,7 +258,7 @@ ImageGridSampler<TInputImage>::SingleThreadedGenerateData(const TInputImage &
             inputImage.TransformIndexToPhysicalPoint(index, tempSample.m_ImageCoordinates);
 
             // Equivalent to `mask->IsInsideInWorldSpace(tempSample.m_ImageCoordinates)`, but much faster.
-            if (mask->IsInsideInObjectSpace(
+            if (mask->MaskType::IsInsideInObjectSpace(
                   worldToObjectTransform.WorldToObjectTransformType::TransformPoint(tempSample.m_ImageCoordinates)))
```
N-Dekker: @mstaring @stefanklein  Can we possibly still optimize for if the mask and the input image have the same geometry?
https://github.com/SuperElastix/elastix/blob/217a07cc455b2afad55a260584c688e50672466d/Common/itkComputeImageExtremaFilter.hxx#L138

N-Dekker: Reported as issue at:
- https://github.com/SuperElastix/elastix/issues/1023

Issue #1023, opened 2024-01-22
ComputeImageExtremaFilter SameGeometry should include Index, Spacing, and Direction


N-Dekker:
The `ComputeImageExtremaFilter::SameGeometry()` (which is used to optimize `ThreadedGenerateDataImageSpatialMask`) now only takes Size and Origin into account, when comparing the input image and the mask:

https://github.com/SuperElastix/elastix/blob/217a07cc455b2afad55a260584c688e50672466d/Common/itkComputeImageExtremaFilter.hxx#L57-L65

It should also include Index, Spacing, and Direction. As discussed at our internal weekly sprint meeting, with @stefanklein and @mstaring

PR #1024, opened 2024-01-23
BUG: ComputeImageExtremaFilter SameGeometry include Spacing, Direction

N-Dekker:
Fixes issue https://github.com/SuperElastix/elastix/issues/1023 "ComputeImageExtremaFilter SameGeometry should include Index, Spacing, and Direction"

Tested that changes of Spacing or Direction of the mask may indeed affect the results of ComputeImageExtremaFilter.

### DIFF
--- a/Common/itkComputeImageExtremaFilter.hxx
+++ b/Common/itkComputeImageExtremaFilter.hxx
@@ -20,6 +20,7 @@
 #include "itkComputeImageExtremaFilter.h"
 
 #include <itkImageRegionConstIterator.h>
+#include "elxImageDomain.h"
 
 namespace itk
 {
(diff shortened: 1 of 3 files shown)


Common/itkComputeImageExtremaFilter.hxx
```diff
@@ -56,9 +56,12 @@ template <typename TInputImage>
...
 {
-  if (this->GetInput()->GetLargestPossibleRegion().GetSize() ==
-        this->m_ImageSpatialMask->GetImage()->GetLargestPossibleRegion().GetSize() &&
-      this->GetInput()->GetOrigin() == this->m_ImageSpatialMask->GetImage()->GetOrigin())
+  const ImageBase<ImageDimension> & inputImage = *(this->GetInput());
+  const ImageBase<ImageDimension> & maskImage = *(this->m_ImageSpatialMask->GetImage());
+
+  if (inputImage.GetLargestPossibleRegion() == maskImage.GetLargestPossibleRegion() &&
+      inputImage.GetOrigin() == maskImage.GetOrigin() && inputImage.GetSpacing() == maskImage.GetSpacing() &&
+      inputImage.GetDirection() == maskImage.GetDirection())
```
N-Dekker, line 64: Using `==` to compare floating point numbers may be a bit too strict, but it's hard to choose an acceptable error margin for "almost equal". In this case I guess we could compare the coordinates of the corresponding corner points of the image and the mask. Too much work, I guess 😸

PR #1039, opened 2024-02-07
Remove unreachable error from `ImageRandomSamplerSparseMask`, extend error message on missing mask 

### LINKED ISSUE #1035: Unreachable error handling, "Consider using the ImageRandomSampler", in ImageRandomSamplerSparseMask
`ImageRandomSamplerSparseMask::GenerateData()` is supposed to produce an error message, in some particular case, saying:

> You are using the ImageRandomSamplerSparseMask sampler, but you did not set a mask. The internal ImageFullSampler therefore requires a lot of memory. Consider using the ImageRandomSampler instead.

At: https://github.com/SuperElastix/elastix/blob/37288103745b0127dc6263dfc8dfa7025e006669/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx#L74-L81

However, at that point, the mask is never null (anymore), so the error handling is unreachable. Because at the begin of `GenerateData()`, it is already checked that the mask isn't null:

https://github.com/SuperElastix/elastix/blob/37288103745b0127dc6263dfc8dfa7025e006669/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx#L38-L42

This check at the begin of `GenerateData()` was added with commit 2fc2c495837f64416263fa7c5d1e2bc4e3e061fa (November 14, 2013): https://github.com/SuperElastix/elastix/blob/2fc2c495837f64416263fa7c5d1e2bc4e3e061fa/src/Common/ImageSamplers/itkImageRandomSamplerSparseMask.txx#L52-L55 Before that commit of November 14, 2013, the error handling ("Consider using the ImageRandomSampler instead") was still reachable.

How are we supposed to deal with this? Shall I just remove the unreachable error handling code, as it has been unreachable for more than 10 years now?

N-Dekker:
Addressed the following issue:
- #1035 

Fixes #1035

### DIFF
--- a/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx
+++ b/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx
@@ -38,7 +38,9 @@ ImageRandomSamplerSparseMask<TInputImage>::GenerateData()
   /** Sanity check. */
   if (mask == nullptr)
   {
-    itkExceptionMacro("ERROR: do not call this function when no mask is supplied.");
+    itkExceptionMacro("ERROR: do not call this function when no mask is supplied. When using the "
+                      "ImageRandomSamplerSparseMask sampler, a mask is required. Otherwise you may consider using a "
+                      "sampler that does not require a mask, for example, ImageRandomSampler.");
   }
 
   /** Get handles to the input image and output sample container. */
--- a/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx
+++ b/Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx
@@ -61,24 +61,11 @@ ImageRandomSamplerSparseMask<TInputImage>::GenerateData()
   {
     this->m_InternalFullSampler->Update();
   }
-  catch (ExceptionObject & err)
+  catch (const ExceptionObject & err)
   {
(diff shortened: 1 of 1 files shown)


Common/ImageSamplers/itkImageRandomSamplerSparseMask.hxx
```diff
@@ -38,7 +38,9 @@ ImageRandomSamplerSparseMask<TInputImage>::GenerateData()
   /** Sanity check. */
   if (mask == nullptr)
   {
-    itkExceptionMacro("ERROR: do not call this function when no mask is supplied.");
+    itkExceptionMacro("ERROR: do not call this function when no mask is supplied. When using the "
+                      "ImageRandomSamplerSparseMask sampler, a mask is required. Otherwise you may consider using a "
+                      "sampler that does not require a mask, for example, ImageRandomSampler.");
```
N-Dekker, line 43: @mstaring Can you please check if you like the proposed extended error message for `mask == nullptr`? Proposed text:

> ERROR: do not call this function when no mask is supplied. When using the ImageRandomSamplerSparseMask sampler, a mask is required. Otherwise you may consider using a sampler that does not require a mask, for example, ImageRandomSampler.

mstaring: prima

PR #1064, opened 2024-03-01
Pull request 52 (ReducedFullSampler) updated

N-Dekker:
- Brought pull request #52 from @MathiasPolfliet up-to-date with the current revision of the `main` branch.

### DIFF
--- a/Common/ImageSamplers/itkImageReducedFullSampler.hxx
+++ b/Common/ImageSamplers/itkImageReducedFullSampler.hxx
@@ -1,0 +1,150 @@
+/*=========================================================================
+ *
+ *  Copyright UMC Utrecht and contributors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef __ImageReducedFullSampler_txx
+#define __ImageReducedFullSampler_txx
+
+#include "itkImageReducedFullSampler.h"
+
+#include "itkImageRegionConstIteratorWithIndex.h"
+
+namespace itk
+{
+
+/**
+ * ******************* GenerateData *******************
+ */
+
+template< class TInputImage >
+void
+ImageReducedFullSampler< TInputImage >
+::GenerateData( void )
+{
+
+  /** Get handles to the input image, output sample container, and the mask. */
+  InputImageConstPointer inputImage = this->GetInput();
+  typename ImageSampleContainerType::Pointer sampleContainer = this->GetOutput();
+  typename MaskType::ConstPointer mask                       = this->GetMask();
+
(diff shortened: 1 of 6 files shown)


Common/ImageSamplers/itkImageReducedFullSampler.hxx
```diff
@@ -0,0 +1,145 @@
...
+  /** Set up a region iterator within the user specified image region. */
+  typedef ImageRegionConstIteratorWithIndex<InputImageType> InputImageIterator;
+  InputImageIndexType                                       index = this->GetCroppedInputImageRegion().GetIndex();
+  index[ReducedInputImageDimension] = 0;
+  InputImageSizeType size = this->GetCroppedInputImageRegion().GetSize();
+  size[ReducedInputImageDimension] = 1;
+  InputImageRegionType region;
+  region.SetIndex(index);
+  region.SetSize(size);
+  InputImageIterator iter(inputImage, region);
```
N-Dekker, line 55: This appears to be the essential part, compared to the "non-reduced" FullSampler.

N-Dekker:
Comparing ReducedFullSampler and FullSampler, based on the original code at commit 2ec22d0faa5189f8c27ae489fb97a2ed0f626318 "VER: increased to 4.900 in preparation of release" (2018):

elxReducedFullSampler.h vs elxFullSampler.h
- 124 vs 123 lines
- Added `ReducedInputImageDimension`

elxReducedFullSampler.h vs elxFullSampler.h
- 27 vs 32 lines
- no relevant changes (just "nothing")

elxReducedFullSampler.cxx vs elxFullSampler.cxx
- 18 vs 22 lines
- no relevant changes (just `elxInstallMacro`)

itkReducedFullSampler.h vs itkFullSampler.h
- 124 vs 126 lines
- Added `ReducedInputImageDimension` and `InputImageSizeType`
- No `ThreadedGenerateData`

itkReducedFullSampler.hxx vs itkFullSampler.hxx
- 150 vs 246 lines
- No ThreadedGenerateData
- Essential added code at the begin of `GenerateData`:
```
  InputImageIndexType index = this->GetCroppedInputImageRegion().GetIndex();
  index[ ReducedInputImageDimension ] = 0;
  InputImageSizeType size = this->GetCroppedInputImageRegion().GetSize();
  size[ ReducedInputImageDimension ] = 1;
  InputImageRegionType region;
  region.SetIndex( index );
  region.SetSize( size );
  InputImageIterator iter( inputImage, region );
```

N-Dekker:
Some comments:
- There appears a lot of code duplication between this component and FullSampler. Can't ReducedFullSampler be written in terms of FullSampler? 
- I would rename it "ReducedDimensionFullSampler", on order to be consistent with the existing "ReducedDimensionBSplineInterpolator". OK?
- Does this suggest a reduced-dimension version for Grid, Random, RandomCoordinate, and RandomSparseMask, as well?

Common/ImageSamplers/itkImageReducedFullSampler.hxx
N-Dekker: @MathiasPolfliet @mstaring @stefanklein Is it really necessary for the "reduced" sampler to do `index[ReducedInputImageDimension] = 0`? Or would it be sufficient to only adjust the size, and leave the index alone?

I'm asking, because ITK allows specifying start index of the buffered region of an image to be non-zero. So an index value of zero may be outside the buffered region. Trying to retrieve samples outside the buffered region might cause crashes. So if it is sufficient for the "reduced" sampler to only adjust the size, I would prefer to leave the index alone. 

Another option might be something like:

    // Instead of index[ReducedInputImageDimension] = 0
    index[ReducedInputImageDimension] = bufferedRegion.GetIndex(ReducedInputImageDimension);

But I wonder if that's useful/necessary.

PR #1069, opened 2024-03-07
STYLE: Don't reset st_NumberOfPixelsCounted in Metric "AfterThreaded"

N-Dekker:
st_NumberOfPixelsCounted is properly reset by the InitializeThreadingParameters() member functions of AdvancedImageToImageMetric, AdvancedKappaStatisticImageToImageMetric, and AdvancedNormalizedCorrelationImageToImageMetric. So it isn't necessary to _also_ reset this "thread variable" in the "AfterThreaded" member functions of the metrics.

### DIFF
--- a/Components/Metrics/AdvancedMeanSquares/itkAdvancedMeanSquaresImageToImageMetric.hxx
+++ b/Components/Metrics/AdvancedMeanSquares/itkAdvancedMeanSquaresImageToImageMetric.hxx
@@ -336,9 +336,6 @@ AdvancedMeanSquaresImageToImageMetric<TFixedImage, TMovingImage>::AfterThreadedG
   for (ThreadIdType i = 1; i < numberOfThreads; ++i)
   {
     this->m_NumberOfPixelsCounted += this->m_GetValueAndDerivativePerThreadVariables[i].st_NumberOfPixelsCounted;
-
-    /** Reset this variable for the next iteration. */
-    this->m_GetValueAndDerivativePerThreadVariables[i].st_NumberOfPixelsCounted = 0;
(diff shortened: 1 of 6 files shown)


Components/Metrics/AdvancedMeanSquares/itkAdvancedMeanSquaresImageToImageMetric.hxx
```diff
@@ -336,9 +336,6 @@ AdvancedMeanSquaresImageToImageMetric<TFixedImage, TMovingImage>::AfterThreadedG
   for (ThreadIdType i = 1; i < numberOfThreads; ++i)
   {
     this->m_NumberOfPixelsCounted += this->m_GetValueAndDerivativePerThreadVariables[i].st_NumberOfPixelsCounted;
-
-    /** Reset this variable for the next iteration. */
-    this->m_GetValueAndDerivativePerThreadVariables[i].st_NumberOfPixelsCounted = 0;
```
N-Dekker, line 341: @mstaring @stefanklein It appears that those "AfterThreaded" member functions typically reset `PerThreadVariables[i].st_NumberOfPixelsCounted` for i = 1 to numberOfThreads, but not for i = 0. Apparently that was never a real problem for i = 0, right? May I then conclude that resetting this variable isn't really necessary for i = 1 to numberOfThreads either?

N-Dekker: I'll merge this pull request now, but I still hope you can confirm that it's OK  😃

PR #1079, opened 2024-03-21
STYLE: Adjust `if` condition on when to use ITK's GradientImageFilter

N-Dekker:
The metrics (currently) only use ITK's `GradientImageFilter` when the interpolator is a NearestNeighbor interpolator.

### DIFF
--- a/Common/CostFunctions/itkAdvancedImageToImageMetric.hxx
+++ b/Common/CostFunctions/itkAdvancedImageToImageMetric.hxx
@@ -29,6 +29,7 @@
 #endif
 
 #include <algorithm> // For min.
+#include <cassert>
 
 namespace itk
 {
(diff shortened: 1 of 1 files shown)


N-Dekker:
The image created by ITK's `GradientImageFilter` (m_GradientImage) is used in `EvaluateMovingImageValueAndDerivativeWithOptionalThreadId`: https://github.com/SuperElastix/elastix/blob/98aa401157d5608b47a3f89892e119c74b0c88fe/Common/CostFunctions/itkAdvancedImageToImageMetric.hxx#L461-L498

Note that AdvancedKappaStatisticImageToImageMetric does not use ITK's `GradientImageFilter`. It assigns its own manually created image to m_GradientImage.

Common/CostFunctions/itkAdvancedImageToImageMetric.hxx
```diff
@@ -360,6 +359,12 @@ AdvancedImageToImageMetric<TFixedImage, TMovingImage>::CheckForBSplineInterpolat
     }
     else
     {
+      using RayCastInterpolatorType =
+        itk::AdvancedRayCastInterpolateImageFunction<MovingImageType, CoordinateRepresentationType>;
+
+      assert(m_BSplineInterpolator || m_BSplineInterpolatorFloat || m_ReducedBSplineInterpolator ||
+             m_LinearInterpolator || dynamic_cast<RayCastInterpolatorType *>(interpolator));
+
       this->m_GradientImage = nullptr;
```
N-Dekker, line 368: @mstaring @stefanklein Do you know, what is the status of RayCastInterpolator? It isn't handled explicitly in EvaluateMovingImageValueAndDerivativeWithOptionalThreadId, meaning that it will use m_GradientImage, by default, at:

https://github.com/SuperElastix/elastix/blob/98aa401157d5608b47a3f89892e119c74b0c88fe/Common/CostFunctions/itkAdvancedImageToImageMetric.hxx#L497

However, m_GradientImage appears to be set to null, for RayCastInterpolator!

PR #60, opened 2018-03-30
ADD: Added the StackTransformBendingEnergy

MathiasPolfliet:
This pull request adds the StackTransformBendingEnergy to the metric components

PR #1083, opened 2024-03-22
ENH: Add "ReduceDimension" option to ImageSampler components

N-Dekker:
Allows reducing the dimensionality of any image sampler by a new elastix parameter:

    (ReduceDimension "true")

Aims to implement a generalization to the ReducedFullSampler, pull request #52 proposed by Mathias Polfliet (@MathiasPolfliet) in 2018.

### DIFF
--- a/Common/ImageSamplers/itkImageSamplerBase.hxx
+++ b/Common/ImageSamplers/itkImageSamplerBase.hxx
@@ -301,6 +301,12 @@ ImageSamplerBase<TInputImage>::CropInputImageRegion()
    * InputImageRegion and the BoundingBoxRegion.
    */
   m_CroppedInputImageRegion = m_InputImageRegion;
+
+  if (m_ReduceDimension)
+  {
+    m_CroppedInputImageRegion.GetModifiableSize().back() = 1;
+  }
+
   if (!m_Mask.IsNull())
   {
     /** Get a handle to the input image. */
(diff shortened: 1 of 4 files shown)


Common/ImageSamplers/itkImageSamplerBase.hxx
```diff
@@ -301,6 +301,12 @@ ImageSamplerBase<TInputImage>::CropInputImageRegion()
    * InputImageRegion and the BoundingBoxRegion.
    */
   m_CroppedInputImageRegion = m_InputImageRegion;
+
+  if (m_ReduceDimension)
+  {
+    m_CroppedInputImageRegion.GetModifiableSize().back() = 1;
```
N-Dekker, line 307: @mstaring @stefanklein It does not seem to be necessary to also adjust the index (by doing `index[N-1] = 0`, maybe), but that depends on the use case, I guess... See https://github.com/SuperElastix/elastix/pull/1064#discussion_r1534075016

N-Dekker:
At today's internal elastix sprint, we (@mstaring, @stefanklein) agreed to use a different name for the new parameter: `"UseSamplingForGroupwiseRegistration"`

N-Dekker:
The parameter is now renamed to "UseSamplingForGroupwiseRegistration", but Stefan (@stefanklein) suggested to double-check if it can really be applied to any groupwise registration. For BendingEnergy, it may not be appropriate.
- Check also the `StackTransformBendingEnergy` component proposed by @MathiasPolfliet: https://github.com/SuperElastix/elastix/pull/60 and https://github.com/SuperElastix/elastix/tree/MathiasPolfliet/StackTransformBendingEnergy

PR #1099, opened 2024-04-17
Replace `GetNumberOfWeights()`, `GetSupportSize()`, and `m_SupportSize` with compile-time equivalents (STYLE/PERF)

N-Dekker:
Replaced run-time evaluation of number of weights and support size with compile-time evaluation of those values. Might improve the run-time performance of bspline transformations and interpolations. Just marked as style improvement, because the changes are not benchmarked (yet).

### DIFF
--- a/Common/Transforms/itkBSplineInterpolationDerivativeWeightFunction.hxx
+++ b/Common/Transforms/itkBSplineInterpolationDerivativeWeightFunction.hxx
@@ -91,7 +91,7 @@ BSplineInterpolationDerivativeWeightFunction<TCoordRep, VSpaceDimension, VSpline
 
     if (i != this->m_DerivativeDirection)
     {
-      for (unsigned int k = 0; k < this->m_SupportSize[i]; ++k)
+      for (unsigned int k = 0; k < VSplineOrder + 1; ++k)
       {
         weights1D[i][k] = KernelType::FastEvaluate(x);
         x -= 1.0;
@@ -99,7 +99,7 @@ BSplineInterpolationDerivativeWeightFunction<TCoordRep, VSpaceDimension, VSpline
     }
     else
     {
-      for (unsigned int k = 0; k < this->m_SupportSize[i]; ++k)
+      for (unsigned int k = 0; k < VSplineOrder + 1; ++k)
       {
         weights1D[i][k] = DerivativeKernelType::FastEvaluate(x);
         x -= 1.0;
--- a/Common/Transforms/itkCyclicBSplineDeformableTransform.hxx
+++ b/Common/Transforms/itkCyclicBSplineDeformableTransform.hxx
@@ -42,7 +42,10 @@ CyclicBSplineDeformableTransform<TScalarType, NDimensions, VSplineOrder>::SetGri
   /** Check if last dimension of supportregion < last dimension of grid. */
   const int lastDim = this->m_GridRegion.GetImageDimension() - 1;
   const int lastDimSize = this->m_GridRegion.GetSize(lastDim);
-  const int supportLastDimSize = this->m_SupportSize.GetElement(lastDim);
+
+  // The support size is the same for all dimensions.
+  const int supportLastDimSize = VSplineOrder + 1;
+
   if (supportLastDimSize > lastDimSize)
   {
     itkExceptionMacro("Last dimension (" << lastDim << ") of support size (" << supportLastDimSize
@@ -163,7 +166,7 @@ CyclicBSplineDeformableTransform<TScalarType, NDimensions, VSplineOrder>::Transf
   this->m_WeightsFunction->Evaluate(cindex, supportIndex, weights);
 
   /** For each dimension, correlate coefficient with weights. */
-  const RegionType supportRegion(supportIndex, Superclass::m_SupportSize);
+  const RegionType supportRegion(supportIndex, WeightsFunctionType::SupportSize);
 
   /** Split support region into two parts. */
   RegionType supportRegions[2];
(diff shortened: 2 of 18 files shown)


Common/Transforms/itkCyclicBSplineDeformableTransform.hxx
```diff
@@ -42,7 +42,7 @@ CyclicBSplineDeformableTransform<TScalarType, NDimensions, VSplineOrder>::SetGri
   /** Check if last dimension of supportregion < last dimension of grid. */
   const int lastDim = this->m_GridRegion.GetImageDimension() - 1;
   const int lastDimSize = this->m_GridRegion.GetSize(lastDim);
-  const int supportLastDimSize = this->m_SupportSize.GetElement(lastDim);
+  const int supportLastDimSize = VSplineOrder + 1;
```
N-Dekker, line 45: @mstaring @stefanklein The original code specifically picked the last dimension from the support size, but as far as I can see, all of its dimensions are the same: `SplineOrder + 1`.

stefanklein:
Hey Niels, thanks. Maybe they are now the same, but in theory/future it could be different, so I think the existing implementation was more robust and generic. Best, Stefan

N-Dekker:
Side note, there appear two different definitions of NumberOfWeights:

- NumberOfWeights = (SplineOrder + 1) * SpaceDimension
- NumberOfWeights = pow(SplineOrder + 1, SpaceDimension)

I believe that only the last one is the correct one: `pow(SplineOrder + 1, SpaceDimension)`. And the first one is wrong. OK?

The one I find suspicious, `NumberOfWeights = (SplineOrder + 1) * SpaceDimension` was introduced 8 years ago, commit 5a80b720269e7c6b660283d75bb9a087f6dcb678 by `GetConstNumberOfWeightsHackRecursiveBSpline`: 

https://github.com/SuperElastix/elastix/blob/5a80b720269e7c6b660283d75bb9a087f6dcb678/src/Common/itkRecursiveBSplineInterpolationWeightFunction.h#L47-L54

It looks like the recursive step of the "hack" was missing. 🤷

stefanklein:
Aren’t those in different files/functions? And both are used somewhere, right?
I don’t remember exactly, but I think in the recursive bspline transform this is actually correct, because it exploits the separability of the kernel.
Best, Stefan

stefanklein:
No concrete plans.
I agree that it is not likely that it will change dynamically.
I'm a bit in a doubt. The previous code was quite logical as well , right?
Is the main aim to make it a static variable, instead of dynamic, or to make the code more clear?
Stefan

N-Dekker:
> Aren’t those in different files/functions? And both are used somewhere, right? I don’t remember exactly, but I think in the recursive bspline transform this is actually correct, because it exploits the separability of the kernel.

Thanks Stefan, but if `NumberOfWeights = (SplineOrder + 1) * SpaceDimension` would have been correct, then why would Wyke Huizinga (@whuizinga) have made a "hack" to estimate its value? Referring again to https://github.com/SuperElastix/elastix/blob/5a80b720269e7c6b660283d75bb9a087f6dcb678/src/Common/itkRecursiveBSplineInterpolationWeightFunction.h#L47-L54 which did:

```cpp
/** Recursive template to retrieve the number of B-spline weights at compile time. */
template< unsigned int SplineOrder, unsigned int Dimension >
class GetConstNumberOfWeightsHackRecursiveBSpline
{
public:

  itkStaticConstMacro( Value, unsigned int, ( SplineOrder + 1 ) * Dimension  );
};

```

and then Wyke (I assume) did:
```cpp
  /** Get number of hacks. */
  typedef GetConstNumberOfWeightsHackRecursiveBSpline<
    itkGetStaticConstMacro( SplineOrder ),
    itkGetStaticConstMacro( SpaceDimension ) > GetConstNumberOfWeightsHackRecursiveBSplineType;
  itkStaticConstMacro( NumberOfWeights, unsigned int, GetConstNumberOfWeightsHackRecursiveBSplineType::Value );
```

It looks to me like the intention was to cleverly calculate `pow(SplineOrder + 1, SpaceDimension)` at compile-time, but it produces just `( SplineOrder + 1 ) * Dimension`. Otherwise, why was there a hack anyway? Why not directly do:

     itkStaticConstMacro( NumberOfWeights, unsigned int, ( SplineOrder + 1 ) * Dimension  );

Anyway, maybe only Wyke knows 🤷 . In order to avoid any confusion, I would prefer to just remove `RecursiveBSplineInterpolationWeightFunction::NumberOfWeights`, as it is unused anyway:
- pull request #1100

whuizinga:
Hi,

I am not sure anymore. It is a long time ago. But isn't space dimension a
template argument that increases in the recursion? Then it would make sense
to put it in the recursion. Otherwise I dont really see it either....

Cheers,

Wyke

Op wo 17 apr 2024 18:16 schreef Niels Dekker ***@***.***>:

whuizinga:
Ok yes, the other function you mention makes more sense.

But if it is not being used, then indeed, it can be removed. Maybe it was
there to validate some computations..

Best,

Wyke

Op wo 17 apr 2024 22:37 schreef Niels Dekker ***@***.***>:

Common/Transforms/itkBSplineInterpolationDerivativeWeightFunction.hxx
```diff
@@ -91,15 +91,15 @@ BSplineInterpolationDerivativeWeightFunction<TCoordRep, VSpaceDimension, VSpline
 
     if (i != this->m_DerivativeDirection)
     {
-      for (unsigned int k = 0; k < this->m_SupportSize[i]; ++k)
+      for (unsigned int k = 0; k <= VSplineOrder; ++k)
```
N-Dekker, line 94: Suggestion by Marius (@mstaring): do `k < (VSplineOrder + 1)`, instead of `k <= VSplineOrder`, for the sake of clarity. (In each of those `for` loops over `k`.)

N-Dekker: Addressed by [this force-push](https://github.com/SuperElastix/elastix/compare/019070bbf1f85ea8dbe322cf7bb873b502fa672e..ec639f4505e4b43ca97aeda9f50710d0d131ee79)

N-Dekker:
- For the record, this pull request would pave the way to further improvements, like ITK pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2721 commit https://github.com/InsightSoftwareConsortium/ITK/commit/974540982b2926ee6dc4f3933c07962ff74ac9a6

PR #1100, opened 2024-04-17
STYLE: Remove `NumberOfWeights` from RecursiveBSpline WeightFunction

N-Dekker:
`RecursiveBSplineInterpolationWeightFunction::NumberOfWeights` was never used. Its value `(SplineOrder + 1) * SpaceDimension` was different from both `BSplineInterpolationWeightFunctionBase::NumberOfWeights` _and_ `RecursiveBSplineInterpolationWeightFunction::m_NumberOfWeights` (both equivalent to `pow(SplineOrder + 1, SpaceDimension)`), which appeared somewhat confusing.

### DIFF
--- a/Common/itkRecursiveBSplineInterpolationWeightFunction.h
+++ b/Common/itkRecursiveBSplineInterpolationWeightFunction.h
@@ -75,9 +75,6 @@ public:
   using typename Superclass::SizeType;
   using typename Superclass::ContinuousIndexType;
 
-  /** The number of weights. */
-  static constexpr unsigned NumberOfWeights = (VSplineOrder + 1) * VSpaceDimension;
-
   /** The number of indices. */
   static constexpr unsigned NumberOfIndices = Math::UnsignedPower(VSplineOrder + 1, VSpaceDimension);
 


Common/itkRecursiveBSplineInterpolationWeightFunction.h
```diff
@@ -75,9 +75,6 @@ class ITK_TEMPLATE_EXPORT RecursiveBSplineInterpolationWeightFunction
   using typename Superclass::SizeType;
   using typename Superclass::ContinuousIndexType;
 
-  /** The number of weights. */
-  static constexpr unsigned NumberOfWeights = (VSplineOrder + 1) * VSpaceDimension;
```
N-Dekker, line 79: For the record, originally it as defined by `GetConstNumberOfWeightsHackRecursiveBSplineType::Value`, here: https://github.com/SuperElastix/elastix/blob/5a80b720269e7c6b660283d75bb9a087f6dcb678/src/Common/itkRecursiveBSplineInterpolationWeightFunction.h#L112

PR #1104, opened 2024-04-21
ENH: Set SubtractMean parameter "true" (instead of "false") by default

N-Dekker:
Let the metrics other than AdvancedNormalizedCorrelation also have default value "true" for their SubtractMean parameter.

Specifically, this affects the following metrics:
- PCAMetric
- PCAMetric2
- SumOfPairwiseCorrelationCoefficients
- VarianceOverLastDimensionImageMetric.

Discussed with Marius Staring (@mstaring) and Stefan Klein (@stefanklein).

- This pull request is _somewhat_ related to pull request #1103 (but this one is _not_ about AdvancedNormalizedCorrelation!)

### DIFF
--- a/Components/Metrics/PCAMetric/elxPCAMetric.hxx
+++ b/Components/Metrics/PCAMetric/elxPCAMetric.hxx
@@ -58,7 +58,7 @@ PCAMetric<TElastix>::BeforeEachResolution()
   this->SetNumEigenValues(NumEigenValues);
 
   /** Get and set if we want to subtract the mean from the derivative. */
-  bool subtractMean = false;
+  bool subtractMean = true;
   this->GetConfiguration()->ReadParameter(subtractMean, "SubtractMean", this->GetComponentLabel(), 0, 0);
   this->SetSubtractMean(subtractMean);
 
--- a/Components/Metrics/PCAMetric/itkPCAMetric.h
+++ b/Components/Metrics/PCAMetric/itkPCAMetric.h
@@ -188,7 +188,7 @@ private:
   double m_VarNoise{};
 
   /** Bool to determine if we want to subtract the mean derivate from the derivative elements. */
-  bool m_SubtractMean{};
+  bool m_SubtractMean{ true };
 
   /** GridSize of B-spline transform. */
   FixedImageSizeType m_GridSize{};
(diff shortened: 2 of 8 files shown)


N-Dekker:
Might conflict with https://github.com/SuperElastix/elastix/tree/Remove-old-itkPCAMetric-files commit 8916b5f6bfc4ff083665f3526502b9f839f01091 _(STYLE: Remove "itkPCAMetric.h" and "itkPCAMetric.hxx")_

N-Dekker:
Note: Specifically for these metrics, this parameter is intended to be renamed to something like "GroupwiseImplicitReferenceFrame", as was discussed with @mstaring and @stefanklein

N-Dekker:
I just noticed that all of the metrics affected by this pull request also have a `m_TransformIsStackTransform` data member. "TransformIsStackTransform" is true _if-and-only-if_ we are doing group-wise registration, right? @stefanklein ?

stefanklein:
Hi Niels,
Good question. It’s not entirely true. When using the (cyclic) 4D bspline deformable transform it is not a stack transform, but still a groupwise registration.
And on the other hand, I can imagine cases where we are registering a stack of slices to another stack of slices using a stack transform, which do not follow the scheme of a typical groupwise registration (although you could call it a groupwise registration, depending on definition).
Does that help? 😊
Best, Stefan

N-Dekker:
Thanks @stefanklein I'm asking, because it appears that "SubtractMean" is relevant for both stack transforms and "non-stack" transforms:

https://github.com/SuperElastix/elastix/blob/de07d4ce0610a48b62379a8a084e90073cc306a3/Components/Metrics/PCAMetric/itkPCAMetric_F_multithreaded.hxx#L527-L531 basically says:

     if (this->m_SubtractMean)
       if (!this->m_TransformIsStackTransform)
         ...do the subtract-mean one way... 
       else
         ...do the subtract-mean another way...
 
So is this SubtractMean option only for groupwise registrations, or is SubtractMean also for a regular registrations?

stefanklein:
Hi Niels,
SubtractMean indeed only applies to groupwise registration. But Groupwise registration can be done with a stacktransform, or with a nD+time B-spline transform (see paper by Coert Metz for the method description),  and for the latter case there is the scenario '...do the subtract-mean one way...' .
Best, Stefan

N-Dekker:
- This pull request (1104) is superseded by pull request #1177

PR #1107, opened 2024-04-26
COMP: Unused parameter `u` BSplineSecondOrderDerivativeKernelFunction2

N-Dekker:
Removed the name of the unused second parameter of `BSplineSecondOrderDerivativeKernelFunction2::Evaluate(const Dispatch<2> &, const double u, double * weights)`

Found by warnings from Clang and GCC, "unused parameter [-Wunused-parameter]".

### DIFF
--- a/Common/Transforms/itkBSplineSecondOrderDerivativeKernelFunction2.h
+++ b/Common/Transforms/itkBSplineSecondOrderDerivativeKernelFunction2.h
@@ -150,7 +150,7 @@ private:
 
 
   static void
-  Evaluate(const Dispatch<2> &, const double u, double * weights)
+  Evaluate(const Dispatch<2> &, const double, double * weights)
   {
     weights[0] = 1.0;
     weights[1] = -2.0;


Common/Transforms/itkBSplineSecondOrderDerivativeKernelFunction2.h
```diff
@@ -150,7 +150,7 @@ class ITK_TEMPLATE_EXPORT BSplineSecondOrderDerivativeKernelFunction2 : public K
 
 
   static void
-  Evaluate(const Dispatch<2> &, const double u, double * weights)
+  Evaluate(const Dispatch<2> &, const double, double * weights)
   {
     weights[0] = 1.0;
     weights[1] = -2.0;
```
N-Dekker, line 156: Note that the other three `Evaluate(const Dispatch<N> &,`...`)` member functions do use the `u` parameter!

N-Dekker: Original code from commit 767a05b178a98249b338e23bec9da45d61dbe105 [ENH: manual syncing the trunk and whuizinga branch](https://github.com/SuperElastix/elastix/commit/767a05b178a98249b338e23bec9da45d61dbe105) (2016). @mstaring @whuizinga As you see, the second parameter (`u`) wasn't used here. No problem to me, I'm just wondering if that was indeed the intention 🤷 

https://github.com/SuperElastix/elastix/blob/767a05b178a98249b338e23bec9da45d61dbe105/src/Common/Transforms/itkBSplineSecondOrderDerivativeKernelFunction2.h#L131-L136

N-Dekker: Just as an experiment, I tried the following instead:
```cpp
    weights[0] = std::numeric_limits<double>::quiet_NaN();
    weights[1] = std::numeric_limits<double>::quiet_NaN();
    weights[2] = std::numeric_limits<double>::quiet_NaN();
```
https://github.com/SuperElastix/elastix/compare/main...WIP-BSplineSecondOrderDerivativeKernelFunction2-NaN 

It appeared to cause no test failures at all. 🤷

PR #1112, opened 2024-05-07
STYLE: Remove `samplenr` which was used in commented-out code in Compute

N-Dekker:
This particular `samplenr` variable was only used as argument for commented-out `progressObserver->UpdateAndPrintProgress` calls.

Those `progressObserver->UpdateAndPrintProgress` calls were already commented-out with commit e9de4b1281a07d2fdbaa96db865f3c7aaaf4e5d2 (May 25, 2016): https://github.com/SuperElastix/elastix/blob/e9de4b1281a07d2fdbaa96db865f3c7aaaf4e5d2/src/Common/itkComputeJacobianTerms.hxx#L235-L236

Follow-up to pull request https://github.com/SuperElastix/elastix/pull/1111 "COMP: Fix `-Wshadow` warnings on local `samplenr` variable"

### DIFF
--- a/Common/itkComputeJacobianTerms.hxx
+++ b/Common/itkComputeJacobianTerms.hxx
@@ -223,8 +223,6 @@ ComputeJacobianTerms<TFixedImage, TTransform>::Compute(double & TrC, double & Tr
   /** Initialize band matrix. */
   CovarianceMatrixType bandcov(numberOfParameters, bandcovsize, 0.0);
 
-  unsigned int samplenr = 0;
-
   /**
    *    TERM 1
    *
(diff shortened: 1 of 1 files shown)


Common/itkComputeJacobianTerms.hxx
```diff
@@ -239,10 +237,6 @@ ComputeJacobianTerms<TFixedImage, TTransform>::Compute(double & TrC, double & Tr
   }
   for (const auto & sample : *sampleContainer)
   {
-    /** Print progress 0-50%.
-     *progressObserver->UpdateAndPrintProgress( samplenr );*/
-    ++samplenr;
```
N-Dekker, line 244: @mstaring The progressObserver calls in itkComputeJacobianTerms have been commented out from the very first itkComputeJacobianTerms, as far as I can see, commit e9de4b1281a07d2fdbaa96db865f3c7aaaf4e5d2 almost 8 years ago. Is it OK if I just remove them now? The helper variable `samplenr` is now unused, and caused compiler warnings.

N-Dekker:
@mstaring FYI, I'll just merge this PR now, but if we ever want to have this unused variable (`samplenr`) back here (and actually use it to implement progress updating), it's easy to put it back any time later.

Common/itkComputeJacobianTerms.hxx
N-Dekker: For the record, Marius just double-checked, and thought this commit should be OK 👍

PR #1115, opened 2024-05-09
Enable and address `-wshadow` warnings from both GCC and clang

N-Dekker:
Enabled and addressed `-wshadow` warnings, triggered by comments by Bradley Lowekamp (@blowekamp) from https://github.com/SimpleITK/SimpleITK/pull/2104#issuecomment-2075030813

The warnings are addressed by choosing more specific (more distinguishing) identifiers, removing unnecessary types and variables, reducing the scope of variables, and using modern C++ enum classes.

### DIFF
--- a/Common/GTesting/itkImageFullSamplerGTest.cxx
+++ b/Common/GTesting/itkImageFullSamplerGTest.cxx
@@ -129,9 +129,11 @@ GTEST_TEST(ImageFullSampler, ExactlyEqualVersusSlightlyDifferentMaskImageDomain)
   static constexpr auto Dimension = 2U;
   using SamplerType = itk::ImageFullSampler<itk::Image<PixelType, Dimension>>;
 
-  std::mt19937 randomNumberEngine{};
-  const auto   image =
-    CreateImageFilledWithSequenceOfNaturalNumbers<PixelType>(CreateRandomImageDomain<Dimension>(randomNumberEngine));
+  const auto image = [] {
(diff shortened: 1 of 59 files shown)


blowekamp:
Nice work!

Common/GTesting/itkImageFullSamplerGTest.cxx
```diff
@@ -129,9 +129,11 @@ GTEST_TEST(ImageFullSampler, ExactlyEqualVersusSlightlyDifferentMaskImageDomain)
...
   using SamplerType = itk::ImageFullSampler<itk::Image<PixelType, Dimension>>;
 
-  std::mt19937 randomNumberEngine{};
-  const auto   image =
-    CreateImageFilledWithSequenceOfNaturalNumbers<PixelType>(CreateRandomImageDomain<Dimension>(randomNumberEngine));
+  const auto image = [] {
+    std::mt19937 randomNumberEngine{};
+    return CreateImageFilledWithSequenceOfNaturalNumbers<PixelType>(
+      CreateRandomImageDomain<Dimension>(randomNumberEngine));
+  }();
```
N-Dekker, line 136: GCC 14.1 still warns about shadowing by a local variable inside a lambda:  

```cpp
    int x{};
    auto lamb = []
    {
        int x{}; // <== warning: declaration of 'x' shadows a previous local [-Wshadow]
    };
```
Looking at https://godbolt.org/z/4jE9PbK9W

PR #1118, opened 2024-05-14
BUG: Let `elx::ConjugateGradientFRPR` override ITK's FRPROptimizer

N-Dekker:
`elx::ConjugateGradientFRPR` member functions GetValueAndDerivative and LineOptimize did originally override the corresponding virtual member functions of ITK's FRPROptimizer. This property was broken by a change in the signature of those ITK member functions: ITK commit https://github.com/InsightSoftwareConsortium/ITK/commit/a088a95fca4e0f94ef661cdc43a6bec26ad687a9, "PERF: Changed to pass by reference for faster performance", 2008-04-23

This commit restores those "overrides", by adjusting the signature of the `elx::ConjugateGradientFRPR` member functions, in accordance with those ITK member functions.

Bug found by macos-12/clang compiler warnings, like:

> warning: 'elastix::ConjugateGradientFRPR::GetValueAndDerivative' hides overloaded virtual function [-Woverloaded-virtual]

Note: These two `elx::ConjugateGradientFRPR` member functions don't seem to be tested at the CI, currently.

### DIFF
--- a/Components/Optimizers/ConjugateGradientFRPR/elxConjugateGradientFRPR.h
+++ b/Components/Optimizers/ConjugateGradientFRPR/elxConjugateGradientFRPR.h
@@ -167,8 +167,8 @@ protected:
    * This implementation calls the Superclass' implementation and caches
    * the computed derivative's magnitude. Besides, it invokes the
    * SelectNewSamples method. */
-  virtual void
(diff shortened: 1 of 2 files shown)


N-Dekker:
While it may or may not be an _important_ bug fix, I think it must be merged anyway!

stefanklein:
Agree!

PR #1120, opened 2024-05-16
Let elastix `ParabolicErodeDilateImageFilter::SplitRequestedRegion` override ITK's ImageSource 

N-Dekker:
Adjusted `ParabolicErodeDilateImageFilter::SplitRequestedRegion` to ensure that its signature matches `ImageSource::SplitRequestedRegion` (again). Following ITK commit https://github.com/InsightSoftwareConsortium/ITK/commit/c70141073652a8c4fef98495405aea0ac03abd3a, June 14, 2011, which modified the signature of `ImageSource::SplitRequestedRegion`, by replacing signed with unsigned integers.

Similar to pull request https://github.com/SuperElastix/elastix/pull/1118 commit 34bdc3ba92ff5a9854b7b6e9b4b4459766b192d5 "BUG: Let `elx::ConjugateGradientFRPR` override ITK's FRPROptimizer"

Note: This commit changes the behavior (restoring the originally intended behavior), because `ParabolicErodeDilateImageFilter::SplitRequestedRegion` does avoid the current dimension (`m_CurrentDimension`), when setting the `splitAxis`.

Bug found by macos-12/clang `-Woverloaded-virtual` warnings.

Removed an assert, and adjusted a GoogleTest unit test, to work around test failures caused by this change.

### DIFF
--- a/Common/itkParabolicErodeDilateImageFilter.h
+++ b/Common/itkParabolicErodeDilateImageFilter.h
@@ -142,8 +142,8 @@ protected:
   void
   GenerateData() override;
 
-  int
-  SplitRequestedRegion(int i, int num, OutputImageRegionType & splitRegion);
+  unsigned int
+  SplitRequestedRegion(unsigned int i, unsigned int num, OutputImageRegionType & splitRegion) override;
 
   void
   ThreadedGenerateData(const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId) override;
--- a/Common/ImageSamplers/itkImageSamplerBase.hxx
+++ b/Common/ImageSamplers/itkImageSamplerBase.hxx
@@ -510,7 +510,6 @@ ImageSamplerBase<TInputImage>::SplitRegion(const InputImageRegionType & inputReg
     if (splitAxis == 0)
     {
       // cannot split
-      assert(!"The region size must be greater than 1!");
       return { inputRegion };
     }
     --splitAxis;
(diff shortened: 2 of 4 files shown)


Common/itkParabolicErodeDilateImageFilter.h
```diff
@@ -142,8 +142,8 @@ class ITK_TEMPLATE_EXPORT ParabolicErodeDilateImageFilter : public ImageToImageF
   void
   GenerateData() override;
 
-  int
-  SplitRequestedRegion(int i, int num, OutputImageRegionType & splitRegion);
+  unsigned int
+  SplitRequestedRegion(unsigned int i, unsigned int num, OutputImageRegionType & splitRegion) override;
```
N-Dekker, line 146: @mstaring @stefanklein I think this change is really necessary, otherwise the special `SplitRequestedRegion` version of ParabolicErodeDilateImageFilter will not be called!  `ParabolicErodeDilateImageFilter::SplitRequestedRegion` is special, because it avoids the "current dimension", when it estimates the `splitAxis`.

This bug must have been introduced with an upgrade of ITK after June 14, 2011. I wonder how it could remain unnoticed for so long! Because ParabolicErodeDilateImageFilter is actually being used by elastix!

N-Dekker: It appears that our ParabolicErodeDilateImageFilter is not tested at the CI, except for being used indirectly by the `ElastixLib.SingleSliceMaskedTranslation3D` GoogleTest unit test (until now). But it still appears obvious that its SplitRequestedRegion member function was intended to override the corresponding member function of ITK's ImageSource. So I'll just merge this now.

PR #1142, opened 2024-06-06
STYLE: Remove unused jacjjacj variable from ComputeUsingSearchDirection

N-Dekker:
The unused local `jacjjacj` variable appears to be introduced with commit e9de4b1281a07d2fdbaa96db865f3c7aaaf4e5d2 "ENH: manual syncing with the performance branch", 25 May 2016.

In other cases, such a variable is (post-)initialized by `vnl_fastops::ABt(jacjjacj, jacj, jacj)`, but in this case, it was left uninitialized.

### DIFF
--- a/Common/itkComputeDisplacementDistribution.hxx
+++ b/Common/itkComputeDisplacementDistribution.hxx
@@ -519,7 +519,6 @@ ComputeDisplacementDistribution<TFixedImage, TTransform>::ComputeUsingSearchDire
   Jgg.Fill(0.0);
   std::vector<double> JGG_k;
   double              globalDeformation = 0.0;
-  JacobianType        jacjjacj(outdim, outdim);
 
   samplenr = 0;
   for (const auto & sample : *sampleContainer)


Common/itkComputeDisplacementDistribution.hxx
```diff
@@ -519,7 +519,6 @@ ComputeDisplacementDistribution<TFixedImage, TTransform>::ComputeUsingSearchDire
   Jgg.Fill(0.0);
   std::vector<double> JGG_k;
   double              globalDeformation = 0.0;
-  JacobianType        jacjjacj(outdim, outdim);
```
N-Dekker, line 522: @mstaring Do you agree that this unused parameter (`jacjjacj`) may simply be removed? It allocates dynamic memory, so removing the variable _might_ improve the performance  😺 More importantly though, was it _meant_ to be used at all?

N-Dekker:
@mstaring FYI, I'll just merge this PR now, but if we ever want to have this unused variable (`jacjjacj`) back here (and actually _use_ it), it's easy to put it back any time later.

Common/itkComputeDisplacementDistribution.hxx
N-Dekker: For the record, Marius just double-checked, and thought this commit should be OK 👍

PR #1151, opened 2024-06-12
Restore windows-2022/Visual Studio 2022 at the CI

N-Dekker:
Looks like the windows-2022 issue is fixed...


N-Dekker:
Giving windows-2022 another try, for both GitHub Actions and Azure Pipelines, based on hopeful comments at other places, for example at https://github.com/actions/runner-images/issues/10004#issuecomment-2162866653

PR #1153, opened 2024-06-12
Remove old "itkPCAMetric" files, rename "itkPCAMetric_F_multithreaded" files to "itkPCAMetric"

N-Dekker:
The old "itkPCAMetric.h" and "itkPCAMetric.hxx" files were not being used anymore. They were not even `#include`'d anywhere. Moreover, they _cannot_ even be used anymore, because they would then conflict with "itkPCAMetric_F_multithreaded.h" and "itkPCAMetric_F_multithreaded.hxx", which define a class template with exactly the same name! Using both "itkPCAMetric_F_multithreaded" _and_ "itkPCAMetric" would probably cause link errors, or otherwise maybe undefined behavior.

- Removing these source files eases further maintenance of elastix, including uncoming changes regarding the `SubtractMean` parameter (pull request #1104)

@mstaring @stefanklein Do you agree that the old "itkPCAMetric.h" and "itkPCAMetric.hxx" files may be removed? And that we can then rename those "itkPCAMetric_F_multithreaded" files to "itkPCAMetric"?


N-Dekker:
For the record, "itkPCAMetric.h" and "itkPCAMetric.hxx" were unused already with their first commit, 764490d5e37de792fea4340fc11850bfd067318d, "Add: Added PCAMetric, PCAMetric2 and SumOfPairwiseCorrelationCoefficientsMetric...", by Wyke (@whuizinga), 24 Nov 2015. Nevertheless, these source files were maintained with every code modernization and style improvement. (20+ "itkPCAMetric.h" commits, 40 "itkPCAMetric.hxx" commits, and counting...)

N-Dekker:
Discussed at the weekly internal elastix sprint. Marius and Stefan agreed that it is OK to remove the old "itkPCAMetric" files, and rename the "itkPCAMetric_F_multithreaded" files to "itkPCAMetric".

PR #1155, opened 2024-06-14
ENH: Add support ComputeZYX parameter to EulerStackTransform (for 3D)

N-Dekker:
Supports specifying parameter `(ComputeZYX "true")`, when the transform is "EulerStackTransform" and the image stack is 4D (meaning that the images are 3D).

Triggered by a question from Laurens Beljaards.


N-Dekker:
It appears that this ITK `Euler3DTransform` bug fix should be included with elastix `AdvancedEuler3DTransform` as well: "BUG: Interpretation of the Euler angles ZYX or ZXY was not exported.", Ziv Yaniv (@zivy), May 9, 2016, https://github.com/InsightSoftwareConsortium/ITK/commit/47c24916609afbd078f2fc51d81962a62be08f7a

(elastix `AdvancedEuler3DTransform` is still from 2009, so it did not yet have this ITK `Euler3DTransform` fix from 2016.)

Issue #1159, opened 2024-06-24
SumOfPairwiseCorrelationCoefficientsMetric m_TransformIsStackTransform always true?!?


N-Dekker:
Looking at commit 764490d5e37de792fea4340fc11850bfd067318d "Add: Added PCAMetric, PCAMetric2 and SumOfPairwiseCorrelationCoefficientsMetric to trunkpublic", Wyke Huizinga (@whuizinga), Nov 24, 2015, it appear that the `TransformIsStackTransform` property of `SumOfPairwiseCorrelationCoefficientsMetric` is always true. It is initialized by `true` in the default-constructor of the metric:

https://github.com/SuperElastix/elastix/blob/764490d5e37de792fea4340fc11850bfd067318d/src/Components/Metrics/SumOfPairwiseCorrelationsMetric/itkSumOfPairwiseCorrelationCoefficientsMetric.hxx#L33

And then it may be set `true`, by calling ` SetTransformIsStackTransform(true)`, in `BeforeEachResolution()`: 

https://github.com/SuperElastix/elastix/blob/764490d5e37de792fea4340fc11850bfd067318d/src/Components/Metrics/SumOfPairwiseCorrelationsMetric/elxSumOfPairwiseCorrelationCoefficientsMetric.hxx#L116

However, it is never `false`. Is it means to be like that?

----

Note that other metrics (PCAMetric, PCAMetric2, VarianceOverLastDimension) have a similar `TransformIsStackTransform` property, but those are initialized to `false`, during the construction of those metrics.

N-Dekker:
@stefanklein A related topic: `BeforeEachResolution()` does:
https://github.com/SuperElastix/elastix/blob/764490d5e37de792fea4340fc11850bfd067318d/src/Components/Metrics/SumOfPairwiseCorrelationsMetric/elxSumOfPairwiseCorrelationCoefficientsMetric.hxx#L111-L116

So: if the current transform is a stack transform, call `SetTransformIsStackTransform(true)`.  But what _else_? There is no `else` part. Is that OK? What if a previous transform was a stack transform, but the current transform is _not_?  Should it not call `SetTransformIsStackTransform(false)`?

PR #1163, opened 2024-06-24
BUG: SumOfPairwiseCorrelationCoefficientsMetric TransformIsStack `false`

N-Dekker:
Initialized `SumOfPairwiseCorrelationCoefficientsMetric::TransformIsStackTransform` to false. Otherwise it would always be true.

Other metrics (PCAMetric, PCAMetric2, VarianceOverLastDimension) have a similar TransformIsStackTransform property, but those are initialized to false, during the construction of those metrics.

 - Addresses issue https://github.com/SuperElastix/elastix/issues/1159 "SumOfPairwiseCorrelationCoefficientsMetric m_TransformIsStackTransform always true?!?"

### DIFF
--- a/Components/Metrics/SumOfPairwiseCorrelationsMetric/itkSumOfPairwiseCorrelationCoefficientsMetric.h
+++ b/Components/Metrics/SumOfPairwiseCorrelationsMetric/itkSumOfPairwiseCorrelationCoefficientsMetric.h
@@ -167,7 +167,7 @@ private:
   FixedImageSizeType m_GridSize{};
 
   /** Bool to indicate if the transform used is a stacktransform. Set by elx files. */
-  bool m_TransformIsStackTransform{ true };
+  bool m_TransformIsStackTransform{ false };
 };
 
 } // end namespace itk


Components/Metrics/SumOfPairwiseCorrelationsMetric/itkSumOfPairwiseCorrelationCoefficientsMetric.h
```diff
@@ -167,7 +167,7 @@ class ITK_TEMPLATE_EXPORT SumOfPairwiseCorrelationCoefficientsMetric
   FixedImageSizeType m_GridSize{};
 
   /** Bool to indicate if the transform used is a stacktransform. Set by elx files. */
-  bool m_TransformIsStackTransform{ true };
+  bool m_TransformIsStackTransform{ false };
```
N-Dekker, line 170: This commit just changes the initial value of m_TransformIsStackTransform, for SumOfPairwiseCorrelationCoefficientsMetric. Without this fix, I think the `if (!this->m_TransformIsStackTransform)` branch cannot be reached:

https://github.com/SuperElastix/elastix/blob/4f35e6fb2b21bde7903ade3266a5658cd8471268/Components/Metrics/SumOfPairwiseCorrelationsMetric/itkSumOfPairwiseCorrelationCoefficientsMetric.hxx#L486-L491

@stefanklein Do you think this fix is OK? Or would it be preferable to just assume that m_TransformIsStackTransform is always true, for this particular metric?

stefanklein:
Hi Niels, I think you are right. This metric should in principle also support the option to run with non-StackTransform. I guess that was never tested with this metric.
That code to subtract the mean (for stack and non-stack transform) is actually duplicated in several metrics probably? Maybe a good candidate for refactoring? But then we need some good tests for groupwise registration with different settings.
Best, Stefan

PR #1167, opened 2024-06-26
BUG: Do `SetTransformIsStackTransform(false)` in BeforeEachResolution()

N-Dekker:
Whenever BeforeEachResolution() detects that there is _no_ stack transform as current transform, let it set TransformIsStackTransform = false, just like it sets TransformIsStackTransform = true when the current transform _is_ a stack transform.

----

- @stefanklein Do you agree that this would be a useful commit? Or would it be unnecessary? (TransformIsStackTransform is now already initialized to false during the construction of all four metrics, with pull request #1163. So I wonder if it's useful to then reset this property with each BeforeEachResolution() call.)

### DIFF
--- a/Components/Metrics/PCAMetric/elxPCAMetric.hxx
+++ b/Components/Metrics/PCAMetric/elxPCAMetric.hxx
@@ -121,9 +121,14 @@ PCAMetric<TElastix>::BeforeEachResolution()
             this->SetGridSize(FixedImageSizeType::Filled(numberOfSubTransforms));
           }
         }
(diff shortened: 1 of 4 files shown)


N-Dekker:
@stefanklein Do you happen to know, why is this Boolean property (`metric->m_TransformIsStackTransform`) set before _each_ resolution? Do you agree that it would make (more) sense to only just set it in `BeforeRegistration()`?

stefanklein:
Hi Niels,
I think you are right. I don't see a reason why it couldn't be done in the BeforeRegistration function
Cheers! Stefan

PR #1171, opened 2024-06-28
STYLE: Do not set GridSize when TransformIsStackTransform, in Metrics

N-Dekker:
The GridSize is only relevant when the transform is _not_ a stack transform!

----

@stefanklein This looks strange to me: each of those three metrics _fill_ their GridSize with numberOfSubTransforms. So for 3D, the grid size would then become numberOfSubTransforms x numberOfSubTransforms x numberOfSubTransforms. But they do so, specifically when the transform is a stack transform. And they do so from the very first commit, by Wyke (@whuizinga) 764490d5e37de792fea4340fc11850bfd067318d, Nov 24, 2015. Looking at elxPCAMetric.hxx: 

https://github.com/SuperElastix/elastix/blob/764490d5e37de792fea4340fc11850bfd067318d/src/Components/Metrics/PCAMetric/elxPCAMetric.hxx#L115-L132

But they only seem to _use_ their GridSize (or at least `m_GridSize[lastDim]`) when the transform is _not_ a stack transform, Looking at itkPCAMetric.hxx: 

https://github.com/SuperElastix/elastix/blob/764490d5e37de792fea4340fc11850bfd067318d/src/Components/Metrics/PCAMetric/itkPCAMetric.hxx#L795-L803

So do you agree that when the transform is a stack transform, the GridSize does _not_ need to be filled? (As proposed by this pull request.)

### DIFF
--- a/Components/Metrics/PCAMetric/elxPCAMetric.hxx
+++ b/Components/Metrics/PCAMetric/elxPCAMetric.hxx
@@ -112,15 +112,6 @@ PCAMetric<TElastix>::BeforeEachResolution()
         /** Set itk member variable. */
         this->SetTransformIsStackTransform(true);
 
-        if (const unsigned int numberOfSubTransforms{ stackTransform->GetNumberOfSubTransforms() };
-            numberOfSubTransforms > 0)
-        {
-          /** Check if subtransform is a B-spline transform. */
-          if (dynamic_cast<ReducedDimensionBSplineTransformBaseType *>(stackTransform->GetSubTransform(0).GetPointer()))
-          {
-            this->SetGridSize(FixedImageSizeType::Filled(numberOfSubTransforms));
-          }
-        }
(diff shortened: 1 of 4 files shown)


N-Dekker:
Oh, I see now, the VarianceOverLastDimension metric already filled its GridSize with NumberOfSubTransforms when its registration had a StackTransform, with commit 5fc4bcc2750edab13cdeee55e93ba7552bd8ed32 Coert Metz (@coertmetz), 27 Oct 2010. Even while it would only actually _use_ its grid size when its registration would _not_ have a StackTransform. (Again, it would only use its `m_GridSize[lastDim]`.) Can anyone please explain?

N-Dekker:
Discussed at the weekly internal elastix sprint. Marius and Stefan agreed that it is not necessary to set the GridSize in these particular cases, so the pull request should be OK.

PR #1174, opened 2024-06-28
STYLE: Replace `this->` with `Superclass::` on m_NumberOfPixelsCounted

N-Dekker:
`m_NumberOfPixelsCounted` is a protected member of ITK's `itk::ImageToImageMetric`, which is a common base class ("superclass") of elastix metric classes.


N-Dekker (commented):
Still "Hang in there while we check the branch’s status"!

PR #1177, opened 2024-06-29
ENH: Replace "SubtractMean" with "UseZeroAverageDisplacementConstraint"

N-Dekker:
Sets its default "true". (For the affected elastix metrics (`PCAMetric`, `PCAMetric2`, `SumOfPairwiseCorrelationCoefficients`, `VarianceOverLastDimension`), the SubtractMean parameter originally had "false" as default.)

The term "UseZeroAverageDisplacementConstraint" is more specific than "SubtractMean", so it appears preferably. Discussed with Marius  (@mstaring) and Stefan (@stefanklein).

Added a warning about this change to the `Initialize()` member functions of the affected elastix metrics.

If "SubtractMean" is specified, now using its value as default for "UseZeroAverageDisplacementConstraint", for backward compatibility.

----

- This pull request supersedes pull request #1104

### DIFF
--- a/Components/Metrics/PCAMetric/elxPCAMetric.h
+++ b/Components/Metrics/PCAMetric/elxPCAMetric.h
@@ -49,9 +49,11 @@ namespace elastix
  *    be set with parameter NumSamplesLastDimension. \n
  * \parameter NumSamplesLastDimension: the number of random samples to take in the time
  *    time direction of the data when SampleLastDimensionRandomly is set to true.
- * \parameter SubtractMean: subtract the over time computed mean parameter value from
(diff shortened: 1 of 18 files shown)


N-Dekker:
@mstaring @stefanklein What do you think? We could are add "Use" prefix: "UseGroupwiseImplicitReferenceFrame". 

The groupwise paper has a chapter "Zero average displacement constraint". Doesn't that suggest "UseZeroAverageDisplacementConstraint" as parameter name?

N-Dekker:
TODO + backwards compat + Rel notes

PR #1180, opened 2024-07-02
BUG: Fix ParabolicErodeDilateImageFilter progress memory leak, use stack

N-Dekker:
Declaring `ProgressReporter` on the stack (instead of on the heap) may benefit run-time performance. But moreover, it fixes a memory leak.

- Corresponds with pull request https://github.com/richardbeare/parabolicMorphology/pull/11 "BUG: Fix memory leaks by moving `progress` from the heap to the stack"


N-Dekker:
Looks like there is more room for improvement:
- Calculate NumberOfRows only for i = m_CurrentDimension, instead of for all dimensions
- Replace m_MagnitudeSign and m_MagnitudeSign with local `constexpr static` variables
- ...

PR #1208, opened 2024-07-24
Add notes on the order of matrix elements of `Direction` and `GridDirection` parameters

N-Dekker:
Triggered by the discussion "Elastix idiosyncracy?", started by Soren Christensen, July 11, 2024, at https://discourse.itk.org/t/elastix-idiosyncracy/7049

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.h
+++ b/Core/ComponentBaseClasses/elxTransformBase.h
@@ -96,8 +96,12 @@ namespace elastix
  * \transformparameter Direction: The direction cosines matrix of the fixed image
  * that was used during registration, and which is used for resampling the deformed moving image
  * if the UseDirectionCosines parameter is set to "true".\n
+ * The matrix elements must be sorted as follows: [ d11 d21 d31 d12 d22 d32 d13 d23 d33] (in 3D).
+ * So the elements must be specified in column-major order (Fortran-style).
+ * Please note: This is not the same as ITK's convention of using row-major order (C-style) for flattening
+ * the direction matrix of an image into a list.\n
  * example: <tt>(Direction -1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 0.1)</tt>\n
- * Default: identity matrix. Elements are sorted as follows: [ d11 d21 d31 d12 d22 d32 d13 d23 d33] (in 3D).
+ * Default: identity matrix.
  * \transformparameter TransformParameters: the transform parameter vector that defines the transformation.\n
  * example <tt>(TransformParameters 0.03 1.0 0.2 ...)</tt>\n
  * The number of entries is stored the NumberOfParameters entry.
--- a/Components/Transforms/AdvancedBSplineTransform/elxAdvancedBSplineTransform.h
+++ b/Components/Transforms/AdvancedBSplineTransform/elxAdvancedBSplineTransform.h
@@ -88,6 +88,7 @@ namespace elastix
  * \transformparameter GridOrigin: stores the origin of the B-spline grid. \n
  *    example: <tt>(GridOrigin 0.0 0.0 0.0)</tt>
  * \transformparameter GridDirection: stores the direction cosines of the B-spline grid. \n
+ *    The matrix elements must be specified in column-major order. \n
  *    example: <tt>(GridDirection 1.0 0.0 0.0  0.0 1.0 0.0  0.0 0.0 0.1)</tt>
  * \transformparameter BSplineTransformSplineOrder: stores the B-spline order 1,2, or 3. \n
  *    example: <tt>(BSplineTransformSplineOrder 3)</tt>
--- a/Components/Transforms/AdvancedBSplineTransform/elxAdvancedBSplineTransform.hxx
+++ b/Components/Transforms/AdvancedBSplineTransform/elxAdvancedBSplineTransform.hxx
@@ -420,6 +420,8 @@ AdvancedBSplineTransform<TElastix>::ReadFromFile()
     this->m_Configuration->ReadParameter(gridindex[i], "GridIndex", i);
     this->m_Configuration->ReadParameter(gridspacing[i], "GridSpacing", i);
     this->m_Configuration->ReadParameter(gridorigin[i], "GridOrigin", i);
+
+    // Retrieve grid direction cosines. (The matrix elements must be specified in column-major order.)
     for (unsigned int j = 0; j < SpaceDimension; ++j)
     {
       this->m_Configuration->ReadParameter(griddirection(j, i), "GridDirection", i * SpaceDimension + j);
--- a/Components/Transforms/BSplineStackTransform/elxBSplineStackTransform.h
+++ b/Components/Transforms/BSplineStackTransform/elxBSplineStackTransform.h
@@ -92,6 +92,7 @@ namespace elastix
  * \transformparameter GridOrigin: stores the origin of the B-spline grid. \n
  *    example: <tt>(GridOrigin 0.0 0.0 0.0)</tt>
  * \transformparameter GridDirection: stores the direction cosines of the B-spline grid. \n
+ *    The matrix elements must be specified in column-major order. \n
  *    example: <tt>(GridDirection 1.0 0.0 0.0  0.0 1.0 0.0  0.0 0.0 0.1)</tt>
  * \transformparameter BSplineTransformSplineOrder: stores the B-spline order 1,2, or 3. \n
  *    example: <tt>(BSplineTransformSplineOrder 3)</tt>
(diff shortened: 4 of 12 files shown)


Core/ComponentBaseClasses/elxTransformBase.h
```diff
@@ -96,6 +96,7 @@ namespace elastix
  * \transformparameter Direction: The direction cosines matrix of the fixed image
  * that was used during registration, and which is used for resampling the deformed moving image
  * if the UseDirectionCosines parameter is set to "true".\n
+ * Please note: The matrix elements should be specified in column-major order!\n
```
N-Dekker, line 99: @stefanklein Please check!

N-Dekker:
For the record, `elastix::ResamplerBase::ReadFromFile()` did already read the Direction matrix in column-major order with commit 695f922245586d1074ca81c99940c09fb48a4224, "ENH: write result images with correct direction cosines (as in fixed image). Also store direction cosines in TransformParameters file", 19 June 2008:

https://github.com/SuperElastix/elastix/blob/695f922245586d1074ca81c99940c09fb48a4224/src/Core/ComponentBaseClasses/elxResamplerBase.hxx#L329-L332

N-Dekker:
Just double-checked: a direction matrix for 3D images specified by elastix parameter `(Direction v0 v1 v2 v3 v4 v5 v6 v7 v8)` will internally be stored in the data buffer of a 3x3 `itk::Matrix` as `{ v0, v3, v6, v1, v4, v7, v2, v5, v8 }`.

Core/ComponentBaseClasses/elxTransformBase.h
lassoan, line 99: I would describe more explicitly that this differs from ITK conventions:
```suggestion
 * Please note: The matrix elements are specified in column-major order (Fortran-style).
 * This is not the same as ITK's convention of using row-major order (C-style) for flattening
 * direction matrix of an image into a list.\n
```

sorenchr2011: I think @lassoan's comments reduces the room for error even further!

N-Dekker: @lassoan Thank you for your suggestion 👍 I'm fine with your suggestion, at least if it's the only place where it should the documented. (If this specification should be documented on multiple places, I'd rather elaborate on the meaning of "column-major order" just once, rather than on each of those places. But this appears the one and only place where it needs to be documented, right?)

One nit-pick question, why do you suggest replacing
> The matrix elements _should be_ specified in column-major order!

with

> The matrix elements _are_ specified in column-major order!
?

I wanted to express that the user _should_ specify the matrix elements in that order.

lassoan: I agree, this detailed note should only appear once in the Elastix documentation. So, if this needs to be mentioned elsewhere in the future than that place should refer to this note or this note could be moved out elsewhere.

To me, "should" has some degree of uncertainty. If "should" is needed to make it clear that it is up to the user has to specify the values then we could use "must" or "shell" to make it clear that column-major order is not just a recommendation but a requirement.

N-Dekker: OK, thanks, then shall we make it "must"?

N-Dekker:
@lassoan I added you as co-author, please check!

N-Dekker:
For the record, my little test to observe the ordering of matrix elements:

```cpp
static constexpr unsigned int                       ImageDimension = 3;
itk::Matrix<double, ImageDimension, ImageDimension> direction;

// Create a "flat" array of matrix elements, {1, 2, ..., N}
std::array<double, ImageDimension * ImageDimension> flatArray;
std::iota(flatArray.begin(), flatArray.end(), 1.0);

// Retrieve the elements by a neste for-loop, just like elastix does at:
// https://github.com/SuperElastix/elastix/blob/995f471aeb17929f8c0557135997e78f54326390/Core/Kernel/elxElastixTemplate.hxx#L1081-L1090
for (unsigned int i = 0; i < ImageDimension; ++i)
{
  for (unsigned int j = 0; j < ImageDimension; ++j)
  {
    std::cout << " v" << (i * ImageDimension + j);
    direction(j, i) = flatArray[i * ImageDimension + j];
  }
}

std::cout << std::endl;

for (const double x : direction.GetVnlMatrix())
{
  std::cout << " v" << x;
}

std::cout << std::endl;

```

Based on https://github.com/SuperElastix/elastix/blob/995f471aeb17929f8c0557135997e78f54326390/Core/Kernel/elxElastixTemplate.hxx#L1081-L1090

Output:
> v0 v1 v2 v3 v4 v5 v6 v7 v8
> v1 v4 v7 v2 v5 v8 v3 v6 v9

N-Dekker:
I still wonder, shouldn't "Direction" be documented in "elxResamplerBase.h", if it's actually being retrieved by `ResamplerBase::ReadFromFile()`...? (It's being used as OutputDirection of its ResampleImageFilter.) But then again, it's also being used for the dummyImage in `SimilarityTransformElastix::ReadCenterOfRotationIndex`, and for `ElastixTemplate::GetOriginalFixedImageDirection`.

N-Dekker:
Related ITK commit: https://github.com/InsightSoftwareConsortium/ITK/commit/a2eccee89a115c5c198a41cb0648f3368fc21eae `STYLE: Fixed documentation about the parameters of AffineTransform and Rigid3DTransform. The first parameters are the matrix elements in row-major order an not column-major.`, Tom Vercauteren, 17 Jun 2009.

Core/ComponentBaseClasses/elxTransformBase.h
N-Dekker: @lassoan @sorenchr2011 Stefan (@stefanklein) just pointer out to me that MetaIO also uses column-major order for its `TransformMatrix` representation. See https://discourse.itk.org/t/metaio-image-rotation-conventions/4525 and https://itk.org/Wiki/ITK/MetaIO/Documentation#Associated_transformations

So then, is ITK's convention really _using row-major order_?  Shouldn't the comment say something like:

```
Please note: The matrix elements are specified in column-major order (Fortran-style, like MetaIO's matrix representation).
ITK often uses row-major order (C-style) for flattening direction matrix of an image into a list.
```

lassoan: > @lassoan @sorenchr2011 Stefan (@stefanklein) just pointer out to me that MetaIO also uses column-major order for its `TransformMatrix` representation. See https://discourse.itk.org/t/metaio-image-rotation-conventions/4525 and https://itk.org/Wiki/ITK/MetaIO/Documentation#Associated_transformations
> 
> So then, is ITK's convention really _using row-major order_?
> 

ITK internally uses row-major order. Various file formats don't have to use the same conventions as ITK. ITK-based libraries (such as Elastix) don't have to use the same conventions as ITK either, but when a user gets a vector from an ITK method then he expects to be able to use it as input in another method in an ITK-based library without any conversion, so it is worth trying to harmonize conventions. Bringing in MetaIO into this discussion is unnecessary and may be distracting.

> The matrix elements are specified in column-major order (Fortran-style, like MetaIO's matrix representation).
ITK often uses row-major order (C-style) for flattening direction matrix of an image into a list.

I don't think this is accurate, because ITK is consistent in how it stores matrices in memory and how they may be exposed as flat arrays via Python wrapping.

Anyway, any documentation would improve things and I don't really mind the details.

lassoan:
> I still wonder, shouldn't "Direction" be documented in "elxResamplerBase.h"

Maybe the best could be describe the convention in every method where the user can get or set a direction array, but keep it very short and simple, with a link to more details. Something like "direction: ..., matrix elements in column-major order (conversion may be needed, see details at URL)" .

Core/ComponentBaseClasses/elxTransformBase.h
```diff
@@ -96,6 +96,9 @@ namespace elastix
  * \transformparameter Direction: The direction cosines matrix of the fixed image
  * that was used during registration, and which is used for resampling the deformed moving image
  * if the UseDirectionCosines parameter is set to "true".\n
+ * Please note: The matrix elements must be specified in column-major order (Fortran-style).
+ * This is not the same as ITK's convention of using row-major order (C-style) for flattening
+ * direction matrix of an image into a list.\n
  * example: <tt>(Direction -1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 0.1)</tt>\n
  * Default: identity matrix. Elements are sorted as follows: [ d11 d21 d31 d12 d22 d32 d13 d23 d33] (in 3D).
```
N-Dekker, line 103: I see now, here it already says:

    Elements are sorted as follows: [ d11 d21 d31 d12 d22 d32 d13 d23 d33] (in 3D)

N-Dekker:
Going to merge now, but maybe the documentation on Direction and GridDirection should be re-organized somehow, eventually. (`Direction` is now documented in `elastix::TransformBase` while it's not specifically used by that base class. On the other hand, `GridDirection` is documented multiple times, redundantly, for four b-spline transforms.)

PR #1213, opened 2024-08-14
Remove ValidRegion properties from BSpline Transforms 

N-Dekker:


Common/Transforms/itkAdvancedBSplineDeformableTransform.hxx
```diff
@@ -99,7 +99,6 @@ AdvancedBSplineDeformableTransform<TScalarType, NDimensions, VSplineOrder>::SetG
     // Where offset = std::floor(spline / 2 ).
     // Note that the last pixel is not included in the valid region
     // with odd spline orders.
-    // For backward compatibility m_ValidRegion is still created.
```
N-Dekker, line 102: For the record, from commit f621db3acc8e7a3762d83aec3ef21936b2e1981d (Coert Metz, 2009), m_ValidRegion was only there for backward compatibility.

PR #1226, opened 2024-08-30
Use modern C++ range-based for loops in OpenCL/GPU based code

N-Dekker:
Addressed clang-tidy (LLVM 18.1.8) messages, saying

> use range-based for loop instead [modernize-loop-convert]

@dpshamonin Can you please have a look?


dpshamonin (approved):
Looks good

PR #1227, opened 2024-08-30
COMP: Add missing semicolons to ITK macro calls

N-Dekker:
Aims to resolve issue https://github.com/SuperElastix/elastix/issues/1224 "Missing semicolons causing compile issues with some compilers".

ITK has become stricter with respect to those semicolons from ITK pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4706 commit https://github.com/InsightSoftwareConsortium/ITK/commit/55e23393e1d2c2e71d919d33459201fcee364655 "ENH: Default to allow ITK_MACROEND_NOOP_STATEMENT", Hans Johnson, June 4, 2024.

PR #1228, opened 2024-08-30
COMP: Add missing semicolons to ElastixRegistrationMethod and OpenCL

N-Dekker:
- Follow-up to pull request https://github.com/SuperElastix/elastix/pull/1227 commit 7b99c207a222fec33bf4aceaf47a949c0f341afb "COMP: Add missing semicolons to ITK macro calls"


N-Dekker:
@Wildcarde Don't know if this follow-up is still of interest to you. It depends on which optional component of elastix you're compiling.

Wildcarde:
Yea I intially tried to work around this by just disabling things, eventually i ran into an error in a core library tho.

PR #1253, opened 2024-10-16
Transform VTK files in accordance with their format and floating point precision

N-Dekker:
When using transformix command-line option "-def" to transform the points of a VTK file, the output was always an ASCII file, having double-precision (64-bit) floating point numbers.

With this pull request, the format (ASCII or binary) and the floating point precision of the output file will correspond with the input file, instead.

- Inspired by pull request #943, @ChristophKirst, Aug 2, 2023.

---

When reviewing this pull request, it may be convenient to ignore whitespace changes: https://github.com/SuperElastix/elastix/pull/1253/files?w=1

### DIFF
--- a/Core/ComponentBaseClasses/elxTransformBase.hxx
+++ b/Core/ComponentBaseClasses/elxTransformBase.hxx
@@ -916,72 +916,95 @@ template <typename TElastix>
 void
 TransformBase<TElastix>::TransformPointsSomePointsVTK(const std::string & filename) const
 {
-  /** Typedef's. \todo test DummyIPPPixelType=bool. */
-  using DummyIPPPixelType = float;
-  using MeshTraitsType =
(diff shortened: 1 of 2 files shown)


N-Dekker:
@ChristophKirst can you please let us know if this pull request would be helpful to you? Would it be possible/useful for you to use 32-bit floating point precision binary VTK files, both as input and output of elastix/transformix?

ChristophKirst:
@N-Dekker this would be super useful and solve our problems. If you detect binary vtk during reading and automatically write binary back (32 bit floating), this would fully solve our problem and help a lot.

N-Dekker:
@mstaring @stefanklein I would like to merge this pull request, which is triggered by a request from [Christoph Kirst](https://github.com/ChristophKirst)! I think it's very straightforward and intuitive: when the input VTK file is a binary file, let the output VTK file also be a binary file. And likewise: when the input is 32-bit, let the output also be 32-bit. Hope it's OK to you!

mstaring:
sounds good to me!

PR #1270, opened 2024-11-22
STYLE: Remove `sigma4factor` from StochasticGradientDescent Optimizers

N-Dekker:
The local variable `sigma4factor` has been equal to `1.0` for more than 16 years. (Starting with commit a90c96af2cfd9d79cf2b154b74b8ca7be5818550, 16 Jan 2008).

----

@stefanklein Looks like you introduced this constant with 27f26bf48252bf888bbe1107b237b66de3f0fcb5 and then set it to `1.0` with the next commit (a90c96af2cfd9d79cf2b154b74b8ca7be5818550), on the very same day, 16 Jan 2008. Do you agree that it may be removed now?


N-Dekker:
OK! Discussed with Stefan at our weekly internal elastix sprint

PR #1278, opened 2024-12-13
Changes in file

julietam:


N-Dekker:
Thanks you for your pull request, Julieta. Do I understand correctly that you are proposing to add support for weighted masks? Can you please explain how you would want to use them? And also, how they should be taken into account by the image registration?

I'm sorry we cannot merge your pull request in its current form. For one thing, the feature should still be tested. (Some sort of test should be added to our test framework.) But it should also have additional documentation. Then the git commits themselves should be more organized, maybe squashed together, with more detailed commit messages.  

Next week we will discuss if there is enough interest to implement this feature.

N-Dekker (changes requested):
Just marked your pull request "Request changes" to avoid merging accidentally. So it is not really a "change request" to you. It is really up to you if you would like to proceed with this pull request. You may also just close the pull request, if that is what you want!

julietam:
Thank you,
And again sorry for the confusion

PR #1288, opened 2025-02-06
Support the TOML file format for input parameter files

N-Dekker:
Supported TOML as file format for input parameter files (including transform parameter files). For example:

    # Image specific
    FixedImageDimension = 3
    MovingImageDimension = 3
    FixedInternalImagePixelType = "float"
    MovingInternalImagePixelType = "float"
    Size = [115, 166, 131]

Using [the TOML++ library](https://github.com/marzer/tomlplusplus) from Mark Gillard (@marzer).

PR #1289, opened 2025-02-10
ENH: Support `OutputTransformParameterFileFormat` parameter for TOML

N-Dekker:
Supports the parameter `OutputTransformParameterMapFileFormat"` in elastix registration parameter files. Allows specifying the file format of the output transform parameter files.

Possible values:

 - `"txt"`: Use the legacy file format, "TransformParameters.*.txt" (the default)
 - `"TOML"`: Use the TOML file format, "TransformParameters.*.toml"
 - `""` (empty string): Use the default

----

- Complementary to pull request #1288

### DIFF
--- a/Common/GTesting/elxConversionGTest.cxx
+++ b/Common/GTesting/elxConversionGTest.cxx
@@ -325,11 +325,14 @@ GTEST_TEST(Conversion, ToOptimizerParameters)
 
 GTEST_TEST(Conversion, ParameterMapToString)
 {
-  EXPECT_EQ(Conversion::ParameterMapToString({}), std::string{});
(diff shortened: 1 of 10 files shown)


N-Dekker:
Some considerations:
- We might choose to support spelling the parameter value for TOML with lowercase: "toml". I did not propose that, because the standard way appears to be uppercase: https://toml.io
- We might support multiple parameter values, `(OutputTransformParameterMapFileFormat "txt" "TOML")`, to output the files in both file formats. But it's a bit extra work to implement. 🤷 
-  We might change the default output format to TOML. But I would rather do so after the next release, to allow users to gradually move from txt to TOML. 
- In Python, it should be quite easy to convert TOML to text, using Python's standard [tomllib](https://docs.python.org/3/library/tomllib.html) package.

PR #1291, opened 2025-02-11
ENH: Add pre-commit and pixi configuration

thewtex:
Add a new ./tools/pre-commit-setup.bash. After a developer clones
the repository, they should run this script. This script will download
pixi, install python via pixi, and install the python pre-commit package
and set up the Git pre-commit hooks. These hooks will run clang-format
and other checks on commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,50 @@
+# .pre-commit-config.yaml
+# for details see https://pre-commit.com
+# for list of available hooks see https://pre-commit.com/hooks.html
+#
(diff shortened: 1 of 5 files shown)


N-Dekker:
Thanks @thewtex Really great!!!

Just double-checked by doing: 

    ./.git/hooks/pixi/bin/pixi.exe run pre-commit-run

It does remove quite a few trailing line breaks from the end of files, but otherwise, it leaves the C++ source code formatting as it is. Which is fine, because I already manually ran clang-format on the entire source tree 👍

PR #1307, opened 2025-03-21
STYLE: Calculate `centerOfRotationIndex` StackTransform in 64-bit double

N-Dekker:
Sync `EulerStackTransform::InitializeTransform()` with `AffineLogStackTransform::InitializeTransform()`.


N-Dekker:
Note that this automatic calculation of `centerOfRotationIndex` occurs in those two InitializeTransform() member functions, but not in the corresponding ReadFromFile() member functions. ReadFromFile() just throw an exception ("Transform parameter file is corrupt."), when there is no "CenterOfRotation" or "CenterOfRotationPoint".

PR #1309, opened 2025-03-24
Remove parameters from `CheckNumberOfSamples`, let `AdvancedImageToImageMetric` directly access its own data

N-Dekker:
The `found` parameter always had the value `Superclass::m_NumberOfPixelsCounted`, so it might as well be removed.

---

@mstaring Looking at the initial commit, it appears that `CheckNumberOfSamples` always had a `found` parameter that should be equal to `m_NumberOfPixelsCounted`. Or the other way around, `m_NumberOfPixelsCounted` should be equal to `found`:  https://github.com/SuperElastix/elastix/blob/f6d2f4b5f63b54b88f407ff84e4dc3ca48b73ecf/src/Common/CostFunctions/itkAdvancedImageToImageMetric.hxx#L730-L740

In practice they _are_ always equal, because `m_NumberOfPixelsCounted` is always passed as the argument for this parameter. For example, https://github.com/SuperElastix/elastix/blob/0dd663acc9e34e9c627d3bf61d246634854f36cc/src/Components/Metrics/AdvancedNormalizedCorrelation/itkAdvancedNormalizedCorrelationImageToImageMetric.txx#L209-L210

So that's why I think the parameter is unnecessary!

### DIFF
--- a/Common/CostFunctions/itkAdvancedImageToImageMetric.h
+++ b/Common/CostFunctions/itkAdvancedImageToImageMetric.h
@@ -483,7 +483,7 @@ protected:
   /** Check if enough samples have been found to compute a reliable
    * estimate of the value/derivative; throws an exception if not. */
   void
-  CheckNumberOfSamples(unsigned long wanted, unsigned long found) const;
+  CheckNumberOfSamples(unsigned long wanted) const;
 
   /** Methods for image derivative evaluation support **********/
 
--- a/Common/CostFunctions/itkAdvancedImageToImageMetric.h
+++ b/Common/CostFunctions/itkAdvancedImageToImageMetric.h
@@ -483,7 +483,7 @@ protected:
   /** Check if enough samples have been found to compute a reliable
    * estimate of the value/derivative; throws an exception if not. */
   void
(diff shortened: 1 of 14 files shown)


mstaring:
Hmm, that should not be the case. We request a certain number of samples, but that is not always found.

N-Dekker:
> Hmm, that should not be the case. We request a certain number of samples, but that is not always found.

But `m_NumberOfPixelsCounted` _is_ the found number of samples! The requested number is `wanted * m_RequiredRatioOfValidSamples`. Right?

N-Dekker:
Update: it appears that _both_ parameters (`wanted` _and_ `found`) can be removed from `AdvancedImageToImageMetric::CheckNumberOfSamples()`!

- `wanted` was always just the total number of samples, which is `m_ImageSampler->GetOutput()->size()`
- `found` was always just `ImageToImageMetric::m_NumberOfPixelsCounted`

The term `wanted` seems a bit confusing, as we really "want" `m_RequiredRatioOfValidSamples * numberOfSamples`.

So then I find it clearer to just remove these two parameters. Hope it's OK to you!

PR #1316, opened 2025-04-16
ENH: Replace `unsigned long` with uint64_t for non-zero Jacobian indices

N-Dekker:
Aims to ease interaction with LibTorch (specifically when passing such indices as `data` parameter to `torch::from_blob`). `unsigned long` appears problematic, as it is typically 32-bit on Windows, while it's typically 64-bit on Linux.

Issue encountered during the development of the IMPACT Similarity Metric, pull request https://github.com/SuperElastix/elastix/pull/1311 by Valentin Boussot (@vboussot) et al.

### DIFF
--- a/Common/Transforms/itkAdvancedTransform.h
+++ b/Common/Transforms/itkAdvancedTransform.h
@@ -130,7 +130,7 @@ public:
    * Using an itk::FixedArray instead of an std::vector gives a performance
    * gain for the SpatialHessianType.
    */
-  using NonZeroJacobianIndicesType = std::vector<unsigned long>;
+  using NonZeroJacobianIndicesType = std::vector<uint64_t>;
   using SpatialJacobianType = Matrix<ScalarType, OutputSpaceDimension, InputSpaceDimension>;
   using JacobianOfSpatialJacobianType = std::vector<SpatialJacobianType>;
   // \todo: think about the SpatialHessian type, should be a 3D native type
--- a/Common/Transforms/itkRecursiveBSplineTransform.hxx
+++ b/Common/Transforms/itkRecursiveBSplineTransform.hxx
@@ -566,8 +566,8 @@ RecursiveBSplineTransform<TScalar, NDimensions, VSplineOrder>::ComputeNonZeroJac
   }
 
   /** Call the recursive implementation. */
-  unsigned long   currentIndex = totalOffsetToSupportIndex;
-  unsigned long * nzjiPointer = &nonZeroJacobianIndices[0];
+  unsigned long currentIndex = totalOffsetToSupportIndex;
+  uint64_t *    nzjiPointer = &nonZeroJacobianIndices[0];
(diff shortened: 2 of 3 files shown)


Common/Transforms/itkAdvancedTransform.h
vboussot, line 1: @N-Dekker Thanks a lot for the modification! Just a quick note: torch::kLong in PyTorch corresponds to int64_t, so if the data is declared as uint64_t, torch::from_blob(..., torch::kLong) will interpret the underlying memory as signed. In practice, values from NonZeroJacobianIndicesType will likely never exceed INT64_MAX, so it should be fine, but it's something to keep in mind for strict type compatibility.

vboussot: Unfortunately, LibTorch doesn’t support unsigned integer types beyond kUInt8

N-Dekker: Strange... I did find `kUInt64` in the LibTorch 2.6.0 distribution:

libtorch-win-shared-with-deps-debug-2.6.0+cpu\libtorch\include\torch\csrc\api\include\torch\types.h from line 54:
```cpp

constexpr auto kUInt64 = at::kUInt64;
constexpr auto kFloat16 = at::kHalf;
constexpr auto kFloat32 = at::kFloat;
constexpr auto kFloat64 = at::kDouble;

/// Rust-style short dtypes.
constexpr auto kU8 = kUInt8;
constexpr auto kU16 = kUInt16;
constexpr auto kU32 = kUInt32;
constexpr auto kU64 = kUInt64;

```

vboussot: @N-Dekker I tested it because the type torch::kUInt64 does exist, but in practice it’s not fully supported. For example, operations like index_add_() fail with the following error:

    index_add_(): Expected dtype int32/int64 for index but got: UInt64

It seems that torch::from_blob(..., torch::kUInt64) works for reading the data, but a cast to torch::kLong is still required to ensure compatibility with indexing operations. I think I’ll go with that approach, since I’d rather keep uint64_t in elastix for types like indices, where unsigned values make more sense.

N-Dekker: Casting a uint64 pointer to an int64 pointer, and then using its values "as if they are signed int64" technically sounds like undefined behavior. Right? On the other hand, it should be safe to just copy those Jacobian indices (uint64) into a signed int64 buffer, and pass that to Torch. Is it really a large performance issue?

If it's important enough, maybe we can make a pull request for torch to improve their kUInt64 support 😃

vboussot: I found this https://github.com/pytorch/pytorch/issues/58734 it's a fairly recent addition in LibTorch

vboussot: I'm not sure when kUInt64 was added, I think it was with PyTorch 2.3 (April 24, 2024). For now, it's reasonable to use torch::from_blob(..., torch::kLong) with uint64_t in this context. Thank you for the change

Issue #1318, opened 2025-04-25
Proposal: CI testing for ARM macOS systems, deprecate Intel macOS support


thewtex:
### **Proposal**

1. Add **macOS ARM64 (macos-14)** to CI/CD testing matrices.
2. Begin deprecating Intel (x86_64) macOS support in future `elastix` releases.
3. Align with ITK’s roadmap (Intel deprecation in ITK 6) and Apple’s industry shift.

---

### **Rationale**

**1. Apple Silicon Dominance**

- Apple holds **90% of the ARM-based laptop market**[^4], with macOS Sonoma (14) and later prioritizing ARM64.
- New GitHub Actions `macos-latest` runners default to ARM64, and Intel support is emulated (slower, less reliable)[^1].
- macOS 17 (2026) is expected to drop Intel support entirely[^3][^5].

**2. ITK Alignment**

- ITK [plans](https://github.com/InsightSoftwareConsortium/ITK/issues/5326) to deprecate Intel macOS in ITK 6 (2026), retaining LTS support in ITK 5.
- Continuing Intel support risks compatibility gaps as ITK optimizes for ARM64 (e.g., SIMD, Metal acceleration).

**3. Industry Trends**

- Developers increasingly exclude Intel from CI/CD pipelines due to performance and maintenance costs[^1].
- Critical tools (Xcode, Homebrew) now prioritize ARM64, limiting Intel debugging/testing capabilities.

---

### **Implementation Steps**

1. **CI/CD Testing**
    - Replace `macos-13` (Intel) with explicit `macos-14` ARM runners.
---

### **Potential Impact**

- **Performance**: ARM64 builds leverage Apple Silicon optimizations.
- **Legacy Users**: Researchers with Intel Macs can use Docker/Linux or older `elastix` versions.

---

### **Supporting Data**

| Key Point | Source |
| :-- | :-- |
| macOS 14+ runners are ARM64 | [GitHub Issue \#9741][^1] |
| Apple holds 90% ARM laptop market | [Patently Apple][^4] |
| macOS Sonoma drops Intel Macs | [9to5Mac][^3], [Wikipedia][^5] |
<img src="https://r2cdn.perplexity.ai/pplx-full-logo-primary-dark%402x.png" class="logo" width="120"/>

---

**Request for Feedback**

1. Are there critical Intel macOS workflows that require extended support?
2. Should `elastix` maintain an LTS branch for Intel macOS (aligned with ITK 5) or is this unnecessary?

[...]

PR #1321, opened 2025-04-28
ENH: Replace `macos-13` with `macos-14` (amd64) in *.yml

N-Dekker:
- Partially addressed issue https://github.com/SuperElastix/elastix/issues/1318 "Proposal: CI testing for ARM macOS systems, deprecate Intel macOS support", proposed by Matt McCormick
- Related to ITK issue https://github.com/InsightSoftwareConsortium/ITK/issues/5326 "Proposal to Deprecate macOS Intel Support in ITK 6 (Retain in ITK 5)"

YAML Label `macos-14` (for GitHub Actions) corresponds with macOS-14-arm64, according to: https://github.com/actions/runner-images/blob/0e37973a015a27a4a410a4ae37c0aa99bdd63ea8/README.md#available-images

However, YAML VM Image Label `macOS-14` (for Azure Pipeline) still appears to refer to an Intel machine. See also https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

### DIFF
--- a/.github/workflows/ElastixGitHubActions.yml
+++ b/.github/workflows/ElastixGitHubActions.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       max-parallel: 3
       matrix:
-        os: [ubuntu-22.04, windows-2022, macos-13]
+        os: [ubuntu-22.04, windows-2022, macos-14]
         include:
           - os: ubuntu-22.04
             c-compiler: "gcc"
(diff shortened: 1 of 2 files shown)


N-Dekker:
This pull request is postponed, because there is still discussion (at issue https://github.com/InsightSoftwareConsortium/ITK/issues/5326) on whether to drop Intel MacOS. It may be too early now. Maybe later 🤷 (issue https://github.com/InsightSoftwareConsortium/ITK/issues/5326)

PR #1323, opened 2025-05-01
ENH: Make random numbers multithreading registrations deterministic

N-Dekker:
Stopped using the global instance of MersenneTwisterRandomVariateGenerator. Stopped calling its `GetInstance()`.

Added a deterministic default-constructed `m_RandomVariateGenerator` to ElastixBase.

Added pointers to this generator to AdvancedImageToImageMetric, ImageRandomSamplerBase, and CMAEvolutionStrategyOptimizer. Added `SetRandomVariateGenerator` member functions to these classes (using a default-constructed generator when `SetRandomVariateGenerator` is not yet called).

Also removed the one `MersenneTwisterRandomVariateGenerator::New()` call from elastix, which was for a local generator in `ImageRandomSamplerBase::GenerateRandomNumberList()`. Instead, added `m_Seed` to ImageRandomSamplerBase, and used that seed for a default-constructed local generator.

Aims to make the results of running multiple registrations parallel (multi-threaded) within a single process deterministic.

- Triggered by pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5287, Michal Meszaros (@michalmeszaroswork), Mar 21, 2025.

----

@michalmeszaroswork Would it be possible for you to check if this pull request would make your use case deterministic (even without adjusting ITK)?


mstaring:
lgtm

N-Dekker:
@michalmeszaroswork Now that this pull request is merged, can you please check that it has indeed solved the problem you described at https://github.com/InsightSoftwareConsortium/ITK/pull/5287  ?

It is merged to the elastix [main branch](https://github.com/SuperElastix/elastix/tree/main/), with commit 6d2dc936eab777baf5faa5a5ec0b538c20ae372c

PR #1329, opened 2025-05-19
Itk elastix 2025 04 08 fb1daee6eec

thewtex:


thewtex:
To clean up we could:

- Use a dedicated CMake option
- Refactor classes ?

PR #1338, opened 2025-06-24
ENH: HasParameter methods for ParameterObject

thewtex:
Make it easy to determine whether a parameter has been set in a
ParameterObject.

### DIFF
--- a/Core/Main/elxParameterObject.cxx
+++ b/Core/Main/elxParameterObject.cxx
@@ -168,6 +168,39 @@ ParameterObject::GetParameter(const unsigned int index, const ParameterKeyType &
 
 
 /**
+ * ********************* HasParameter *********************
+ */
+
+bool
+ParameterObject::HasParameter(const unsigned int index, const ParameterKeyType & key) const
+{
+  if (index >= m_ParameterMaps.size())
+  {
+    return false;
+  }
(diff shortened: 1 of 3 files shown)


Core/Main/elxParameterObject.cxx
```diff
@@ -167,6 +167,39 @@ ParameterObject::GetParameter(const unsigned int index, const ParameterKeyType &
...
+ */
+
+bool
+ParameterObject::HasParameter(const unsigned int index, const ParameterKeyType & key) const
+{
+  if (index >= m_ParameterMaps.size())
+  {
+    return false;
+  }
+  return m_ParameterMaps[index].find(key) != m_ParameterMaps[index].end();
```
N-Dekker, line 181: Mostly looks fine to me, thanks Matt! In this particular case, I would rather avoid calling `m_ParameterMaps[index]` twice, by doing, instead:

```cpp
    return m_ParameterMaps[index].count(key) > 0;
```
Of course, the ideal would be `m_ParameterMaps[index].contains(key)`, but that's for C++20  😸

thewtex: Good idea, Neils! Added.

N-Dekker (approved):
Thanks Matt! 👍

PR #1342, opened 2025-07-01
ENH: Support initialization from ITK transform files in TransformBase

vboussot:
elastix writes ITK transform files through `ITKTransformOutputFileNameExtension`, but `-t0` cannot read one back: the file is parsed as an elastix transform parameter file and the run stops there.

This makes `-t0` fall back to `TransformIO::Read` when that parse fails, and wrap the result in the matching elastix transform component. Any format ITK reads works (`.itk.txt`, `.h5`, `.tfm`), and a file that is neither still fails with the same error as before.

To be clear about what this is worth: `(Transform "File")` with `TransformFileName` already covers this from a parameter file, and `SetInitialTransform` covers it from C++ and Python. So this is convenience for the command line, not a new capability.

Two points I would rather raise than hide:

- the elastix parser still logs its own error before the fallback succeeds, so a working run prints a confusing ERROR line
- the output carries `(InitialTransformParameterFileName "")`, so the composition cannot be replayed by transformix

Rebased on current main. The previous version keyed off the `.itk.txt` suffix, which missed the other ITK formats.


N-Dekker:
I'll have another look at this PR of yours, thanks for your patience  😃 

I didn't know about ".itk.txt" being used as file extension, is it for the same file format as ".tfm"? (The "legacy" file format implemented by https://github.com/InsightSoftwareConsortium/ITK/tree/main/Modules/IO/TransformInsightLegacy). If so, I would prefer to use the ".tfm" file extension. 🤷

vboussot:
@N-Dekker Yes, same format 😄 I actually just realized .itk.txt is probably a local LTSI convention I picked up when I started my PhD. I’ve been using it ever since and somehow never noticed I might be the only one doing that 😂

.tfm is definitely clearer here.

N-Dekker:
I think this feature is not yet complete. When a user runs elastix with `-t0 <path/to/file>`, the filename is expected to appear in the generated transform file (typically "TransformParameters.0.txt"), as in:

    (InitialTransformParameterFileName "path/to/file")

That's still missing, as far as I can see. Also, transformix should be able to read back the InitialTransformParameterFileName.

N-Dekker (changes requested):
Marked "request changes", just so that we won't merge it _accidentally_!  😃  See my comment at https://github.com/SuperElastix/elastix/pull/1342#issuecomment-5516463002

Core/ComponentBaseClasses/elxTransformBase.hxx
```diff
@@ -454,14 +454,77 @@ TransformBase<TElastix>::ReadInitialTransformFromFile(const std::string & transf
 
   if (configurationInitialTransform->Initialize({ { "-tp", transformParameterFileName } }) != 0)
```
N-Dekker, line 455: I find it a bit cumbersome that this approach depends on a run-time failure of `configurationInitialTransform->Initialize` (including an error message). Maybe it would be nicer to let `configurationInitialTransform->Initialize` take care of a possible TFM or HDF5 file, passed to `-t0`. What do you think? I already have a draft implementation, locally here.

N-Dekker: Here is my draft implementation: 
https://github.com/SuperElastix/elastix/tree/Initialization-from-ITK-transform-files

Please have a look! What do you think?

PS You may ignore whitespace changes, when looking at the code:
https://github.com/SuperElastix/elastix/commit/f7e8ee3ce6555f38ee658927e3e52a7d8278c0ce?w=1

vboussot: You're right, and your version also fixes the InitialTransformParameterFileName you mentioned: going through Configuration::Initialize, ReadFromFile() sets it from -tp on its own. Mine bypasses that path, which is why it's empty. Let's go with yours.

Two things I noticed. .txt is ambiguous, since ITK's TxtTransformIOTemplate::CanReadFile takes it as well as .tfm, and elastix can write ITK transforms with that extension via ITKTransformOutputFileNameExtension, so those end up in the parameter parser and fail. Could we try the parameter parser first for .txt, then the transform reader?

And TransformIO::Read throws rather than returning null, so the else with log::error and return 1 never runs. The exception escapes Configuration::Initialize instead.

vboussot:
Superseded by #1468.

N-Dekker:
@vboussot I'll merge ~~this~~ #1468 now, but please let me know if it doesn't work the way you originally had in mind! 

"this", I mean https://github.com/SuperElastix/elastix/pull/1468.  😃

Issue #1250, opened 2024-10-01
Compile errors when instantiating `OpenCLVector<T>`


N-Dekker:
An attempt to instantiate `OpenCLVector<T>` for a specific template argument `T` triggers compile errors. For example: 

```cpp
// Template instantiation for T = float.
template class itk::OpenCLVector<float>;

```

VS2022 output:
```
ITKimprovements\itkOpenCLVector.hxx(100,42): error C2039: 'map': is not a member of 'itk::OpenCLVector<float>'
ITKimprovements\itkOpenCLVector.hxx(137,43): error C2039: 'GetLength': is not a member of 'itk::Vector<float,3>'

```
Affected lines of code:

https://github.com/SuperElastix/elastix/blob/0852e7a4db5281f1be67f88163e66d350e1590c7/Common/OpenCL/ITKimprovements/itkOpenCLVector.hxx#L100

https://github.com/SuperElastix/elastix/blob/0852e7a4db5281f1be67f88163e66d350e1590c7/Common/OpenCL/ITKimprovements/itkOpenCLVector.hxx#L137

I guess these errors can be fixed easily. But are the affected member functions `operator[]` and `Write`) really necessary? Or can they simply be removed?

N-Dekker:
@AidynAIMedical wrote:
> If so, is there a reason it hasn’t been changed yet?

Thanks for asking! We left this issue unfixed, because it did not seem to bother anyone (other than myself). Until now! We did not know that it would cause compile errors from (newer versions of) Xcode!

I'm preparing a fix now...

N-Dekker:
@AidynAIMedical I just removed the problematic `operator[]`overload. (Merged PR #1347 to the main branch.) Is that sufficient for you to make it work?

AidynAIMedical:
@N-Dekker Sorry for the late reply. Yes, it did build without issues. Thank you very much for your prompt actions!

PR #1347, opened 2025-08-19
COMP: Remove const overload of `operator[]` from OpenCLVector

N-Dekker:
`OpenCLVector::operator[]` does potentially alter the internal state of the object, so it's more const-correct to just provide a non-const overload of this operator.

Moreover, this overload caused compile errors from Xcode (16.4) / Clang (17.0.0), as Aidyn Ubingazhibov (@AidynAIMedical) reported at issue https://github.com/SuperElastix/elastix/issues/1250:

> no member named 'map' in 'OpenCLVector<T>'; did you mean 'Map'?

This suggests that this particular overload was never used anyway.

### DIFF
--- a/Common/OpenCL/ITKimprovements/itkOpenCLVector.h
+++ b/Common/OpenCL/ITKimprovements/itkOpenCLVector.h
@@ -104,12 +104,6 @@ public:
   T &
   operator[](const std::size_t index);
 
-  /** Returns a const reference to the element at \a index in this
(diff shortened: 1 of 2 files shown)


N-Dekker:
@dpshamonin Hi Denis! Just for your information, I'll just remove the const `OpenCLVector::operator[]` overload for now. Let me know if you ever want it back! You know where to find me  😃

PR #1365, opened 2025-09-16
BUG: Initialize GridScheduleComputer ImageDirection as identity matrix

N-Dekker:
`m_ImageDirection` was originally just zero-filled, in the default-constructor of `GridScheduleComputer`. By commit 18e1e8b0d16092fc353b14ccf64195fdcd9b0d70, Coert Metz, 28-06-2013. However `itk::Image` does not allow using a zero-filled direction. For example, the following code would cause an exception:

    ImageType::DirectionType direction;
    direction.Fill(0.0);
    image->SetDirection(direction);

The exception would says something like:

> itkImageBase.hxx:139: in 'void itk::ImageBase::SetDirection(const itk::Matrix &)':
> ITK ERROR: Image: Bad direction, determinant is 0. Refusing to change direction from 1 0 0 1 to 0 0 0 0

Obviously, `itk::Image` _does_ allow using an identity matrix instead.

### DIFF
--- a/Common/Transforms/itkGridScheduleComputer.hxx
+++ b/Common/Transforms/itkGridScheduleComputer.hxx
@@ -40,7 +40,7 @@ GridScheduleComputer<TTransformScalarType, VImageDimension>::GridScheduleCompute
 
   m_ImageOrigin.Fill(0.0);
   m_ImageSpacing.Fill(1.0);
-  m_ImageDirection.Fill(0.0);
+  m_ImageDirection.SetIdentity();
   m_FinalGridSpacing.Fill(0.0);
 
   this->SetDefaultSchedule(3);


Common/Transforms/itkGridScheduleComputer.hxx
```diff
@@ -40,7 +40,7 @@ GridScheduleComputer<TTransformScalarType, VImageDimension>::GridScheduleCompute
 
   m_ImageOrigin.Fill(0.0);
   m_ImageSpacing.Fill(1.0);
-  m_ImageDirection.Fill(0.0);
+  m_ImageDirection.SetIdentity();
```
N-Dekker, line 43: @mstaring @stefanklein Could this be a serious bug? (Zero-filling looks like an incorrect way to initialize an direction matrix.) The fix is easy, anyway, as you see 😃

mstaring: Probably line 161 in that file saves us, right?

mstaring: I see. Ok, then I guess we were saved by:
https://github.com/SuperElastix/elastix/blob/030a80d19cd3c68ad46481b438170ad946917c61/Components/Transforms/AdvancedBSplineTransform/elxAdvancedBSplineTransform.hxx#L165

But a good catch anyway!

N-Dekker: Thanks @mstaring !  I just added the following to the commit text:

> Note that this bug never affected any actual use case of elastix, because the ImageDirection of a GridScheduleComputer is in practice always set, in the PreComputeGridInformation() member function of the transform that uses the GridScheduleComputer. As remarked by Marius Staring.

PR #1371, opened 2025-10-16
ENH: Add MaximumNumberOfIterationsOfDefaultParameterMap GTest unit test

N-Dekker:
Note that the value of "MaximumNumberOfIterations" from `GetDefaultParameterMap` (`256`) is different from the default value of the "MaximumNumberOfIterations" parameter, when a parameter map is created from scratch, or read from a parameter file.

Most optimizers use a default of `500` iterations:

    AdaGrad
    AdaptiveStochasticGradientDescent
    AdaptiveStochasticLBFGS
    AdaptiveStochasticVarianceReducedGradient
    CMAEvolutionStrategy
    FiniteDifferenceGradientDescent
    Powell
    PreconditionedStochasticGradientDescent
    RegularStepGradientDescent
    Simplex
    SimultaneousPerturbation
    StandardGradientDescent

Other optimizers use a default of `100` iterations:

    ConjugateGradient
    ConjugateGradientFRPR
    QuasiNewtonLBFGS
    RSGDEachParameterApart

### DIFF
--- a/Core/Main/elxParameterObject.cxx
+++ b/Core/Main/elxParameterObject.cxx
@@ -462,6 +462,7 @@ ParameterObject::GetDefaultParameterMap(const std::string & transformName,
   parameterMap["FinalBSplineInterpolationOrder"] = { "3" };
   parameterMap["NumberOfResolutions"] = { std::to_string(numberOfResolutions) };
   parameterMap["WriteIterationInfo"] = { "false" };
+  parameterMap["MaximumNumberOfIterations"] = { "256" };
 
   // Image Sampler
   parameterMap["ImageSampler"] = { "RandomCoordinate" };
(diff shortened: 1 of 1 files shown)


N-Dekker:
@mstaring @stefanklein Just an observation (not necessarily a problem): so the default value of "MaximumNumberOfIterations" is either `256` (when using GetDefaultParameterMap, originally by Kasper), `500` (when unspecified, for most optimizers), or `100` (when unspecified, for four other optimizers). Although I find it a bit confusing, I can imagine that for some optimizers need less iterations than others. Right?

Our napari plugin uses `256` as "default default setting" (non-advanced, just by using GetDefaultParameterMap), but `500` as "advanced default setting". That's also a bit confusing to me 🤷 https://github.com/SuperElastix/elastix-napari/blob/eae95c5b9b12842d6166cdff86bbd6bacf37b212/elastix_napari/elastix_registration.py#L106

PR #1396, opened 2026-01-27
ENH: Lazy loading for optional components (IMPACT)

vboussot:
This PR introduces a first, minimal implementation of on-demand loading for elastix components, following the discussion in  https://github.com/SuperElastix/elastix/pull/1394#issuecomment-3801183554.

The goal is to keep the official elastix binaries lightweight and dependency-free, while still allowing optional
components (IMPACT) to be enabled transparently.

This implementation is intentionally small and conservative, and meant as a basis for discussion and extensive testing.

## Changes

- `MainBase::CreateComponent()` now:
 1) Tries to create the component from the existing `ComponentDatabase` (current behavior).
 2) If not found, tries to load a plugin and retries.

- A new `TryLoadComponentPlugin()` is introduced:
  - Triggered only when `"Impact"` is requested.
  - Tries `ImpactMetricCuda` first, then falls back to `ImpactMetric`.
  - Loads the shared library with `dlopen` / `LoadLibrary`.
  - Calls `*InstallComponent` to register the component.
  - Prints clear diagnostics when IMPACT is requested but unavailable, making it explicit whether the failure comes from     a missing plugin or from unresolved runtime dependencies (e.g. LibTorch).

- A mutable accessor to `ComponentDatabase` is added to allow late registration by plugins.

- IMPACT is built as a `MODULE` (plugin) instead of a static component:
  - elastix has no hard dependency on LibTorch.
  - IMPACT becomes optional and is activated only when it is requested via `(Metric "Impact")`.

- On Linux, PIC is forced to allow loadable modules.

- The existing CI remains unchanged: elastix is still fully tested with the IMPACT CPU variant, exactly as before.
- In addition, a dedicated workflow is introduced to produce release artifacts with IMPACT plugins:
  - `ImpactMetric`
  - `ImpactMetricCuda` (Linux/Windows only)

## Result

As a result, elastix always works out of the box, and IMPACT is enabled transparently when the corresponding LibTorch
libraries are available on the runtime search path. When IMPACT is requested but cannot be activated, error messages clearly indicate whether the failure comes from a missing plugin or from a missing runtime dependency (e.g. LibTorch).

### DIFF
--- a/Core/Kernel/elxMainBase.cxx
+++ b/Core/Kernel/elxMainBase.cxx
@@ -50,10 +50,10 @@ namespace elastix
 MainBase::MainBase() = default;
 
 /**
- * ********************* GetComponentDatabaseMutable ****************************
+ * ********************* GetMutableComponentDatabase ****************************
  */
 ComponentDatabase &
-MainBase::GetComponentDatabaseMutable()
+MainBase::GetMutableComponentDatabase()
 {
   // Improved thread-safety by using C++11 "magic statics".
   static const auto staticComponentDatabase = [] {
@@ -69,7 +69,7 @@ MainBase::GetComponentDatabaseMutable()
   }();
 
   return *staticComponentDatabase;
-} // end GetComponentDatabaseMutable
+} // end GetMutableComponentDatabase
 
 /**
  * ********************* GetComponentDatabase ****************************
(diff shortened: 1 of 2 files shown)


vboussot:
@N-Dekker The only side effect I’ve identified so far when switching IMPACT from a static component to a dynamically loaded shared library is the following: on macOS, writing feature maps (.mha) can fail with “There are no registered IO factories”.

This seems related to ITK IO factories not being registered when the component is loaded dynamically. Have you encountered this pattern before, or do you have a preferred way of handling ITK IO registration in this kind of setup?

PR #1415, opened 2026-03-12
BUG: Fix const-correctness of `elx::ParameterObject::GetParameter`

### LINKED ISSUE #1398: `elx::ParameterObject::GetParameter` should not modify the ParameterObject
`ParameterObject::GetParameter` calls `m_ParameterMaps[index][key]`, which potentially modifies the map at `m_ParameterMaps[index]`:

https://github.com/SuperElastix/elastix/blob/9f87f8943ac5c90bf5fffa1b5ac1566a607ad020/Core/Main/elxParameterObject.cxx#L166-L170

When the key is not in the specified map, it will add an empty `std::vector<std::string>` to the map, for that key.

I don't think that's what we want.

N-Dekker:
Let `parameterObject->GetParameter(index, key)` return an empty vector of values when the specified key is not there, without modifying the specified parameter map.

- Fixes https://github.com/SuperElastix/elastix/issues/1398

----

@mstaring @stefanklein @thewtex `parameterObject->GetParameter(index, key)` silently returns an empty vector of values when the key does not exist. (It already did so with commit 0ce6c24a2b74c0132f1a53848321f1c7a92162ea, June 2018, when the function was introduced.) Do we want to keep it like that?

We could instead let it throw an exception, in that case. It would prevent users from having typo's in their key names. But that might _possibly_ break some legacy user code 🤷

### DIFF
--- a/Core/Main/elxParameterObject.cxx
+++ b/Core/Main/elxParameterObject.cxx
@@ -164,9 +164,11 @@ ParameterObject::SetParameter(const ParameterKeyType & key, const ParameterValue
  */
 
 const ParameterObject::ParameterValueVectorType &
-ParameterObject::GetParameter(const unsigned int index, const ParameterKeyType & key)
+ParameterObject::GetParameter(const unsigned int index, const ParameterKeyType & key) const
 {
-  return m_ParameterMaps[index][key];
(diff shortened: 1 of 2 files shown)


thewtex (approved):
👍

PR #1424, opened 2026-03-26
ENH: Replace ImpactWriteFeatureMaps with ImpactFeatureMapOutputDirectory and fix publish-elastix workflow bug

vboussot:
This PR replaces the deprecated ImpactWriteFeatureMaps parameter with ImpactFeatureMapOutputDirectory, following the suggestion by @N-Dekker (see https://github.com/vboussot/ImpactElastix/issues/6).

It also fixes an issue in the publish-elastix GitHub Actions workflow related to the generation of release artifacts.

In addition, for the 5.3.1 release, GPU support could be enabled by manually running the publish-elastix.yml workflow to build the ImpactMetricCuda library, and then adding the generated binaries (ImpactMetricCuda.so for Linux and ImpactMetricCuda.dll for Windows) to the existing release archives alongside the CPU library.

At runtime, when installing the component, elastix first attempts to load ImpactMetricCuda if a compatible GPU environment is detected. Otherwise, it automatically falls back to ImpactMetric, resulting in CPU-only execution.

Currently, the 5.3.1 release only ships the CPU library, meaning that even on systems with GPU support elastix immediately falls back to ImpactMetric, and GPU acceleration cannot be used.

### DIFF
--- a/.github/workflows/publish-elastix.yml
+++ b/.github/workflows/publish-elastix.yml
@@ -139,9 +139,9 @@ jobs:
         shell: cmd
         run: |
           cd elastix-build
-          cmake -DCMAKE_C_COMPILER="${{ inputs.c-compiler }}" ^
-              -DCMAKE_CXX_COMPILER="${{ inputs.cxx-compiler }}" ^
-              -DCMAKE_BUILD_TYPE="${{ inputs.cmake-build-type }}" ^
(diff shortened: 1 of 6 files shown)


vboussot:
Superseded by #1463, which carries the `ImpactFeatureMapOutputDirectory` rename and the publish-elastix workflow fix along with the backend refactor. Closing this one.

PR #1425, opened 2026-03-27
COMP: Update pixi config from project to workspace

thewtex:
This is the new name.

Addresses:

  WARN Encountered 1 warning while parsing the manifest:
  ⚠ The `project` field is deprecated. Use `workspace` instead.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,4 +1,4 @@
-[project]
+[workspace]
 name = "elastix"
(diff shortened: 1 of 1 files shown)


N-Dekker:
Thanks Matt! Can you please tell me where that warning appears?

thewtex:
@N-Dekker it occurs when installing the pre-commit hooks.

PR #1428, opened 2026-04-10
COMP: Restore new GTest targets for ITK > v5.4.5

N-Dekker:
Restored support for new GTest target names, that was already introduced with pull request https://github.com/SuperElastix/elastix/pull/1408 commit b2701168a9f2f1300cf9403cd20af5918a3b3d74 "COMP: Use GTest::gtest and GTest::gtest_main targets for ITK > v5.4.5"

It was accidentally removed with pull request https://github.com/SuperElastix/elastix/pull/1396 commit d3fd241565e4dac79a806b8529239f030bce9c8d "ENH: Enable IMPACT as a lazy-loaded plugin (CPU/CUDA)"

---
@vboussot Please have a look! I _think_ we accidentally removed that piece of CMake code, right?


vboussot:
Nice catch!
My bad, I merged and kept the wrong version =/

PR #1440, opened 2026-05-05
PERF: Use ImageBufferRange in AfterThreadedComputePDFs

N-Dekker:
Made `ParzenWindowHistogramImageToImageMetric::AfterThreadedComputePDFs` much faster.

The registration test case of the benchmark by Nicolas Chiaruttini (https://discourse.itk.org/t/8x-slower-registration-with-itk-elastix-python-api-vs-elastix-cli-minimal-reproducible-example/7736) appears to run more than 30 % faster with the CLI (elastix executable), from ~3 seconds to ~2 seconds on Window 11, at "my" [LKEB/LUMC](https://www.lkeb.nl/) pc (AMD Threadripper, 32-Cores, 4 GHz, 64 logical processors), when using the default number of work units for the metric (`ITK_MAX_THREADS`, which is 128 for the executable), so without specifying "-threads".


thewtex (approved):
🚀 🔥

PR #1448, opened 2026-05-31
Use log level for image writing progress messages.

jakeforster:
* Core/ComponentBaseClasses/elxResamplerBase.hxx (WriteResultImage): Use log::info instead of log::to_stdout.
* Core/ComponentBaseClasses/elxFixedImagePyramidBase.hxx (WritePyramidImage): Likewise.
* Core/ComponentBaseClasses/elxMovingImagePyramidBase.hxx (WritePyramidImage): Likewise.

### DIFF
--- a/Core/ComponentBaseClasses/elxResamplerBase.hxx
+++ b/Core/ComponentBaseClasses/elxResamplerBase.hxx
@@ -395,7 +395,7 @@ ResamplerBase<TElastix>::WriteResultImage(OutputImageType *   image,
   /** Do the writing. */
   if (showProgress)
   {
-    log::to_stdout("  Writing image ...");
+    log::info("  Writing image ...");
   }
 #ifndef ELX_NO_FILESYSTEM_ACCESS
   try
--- a/Core/ComponentBaseClasses/elxFixedImagePyramidBase.hxx
+++ b/Core/ComponentBaseClasses/elxFixedImagePyramidBase.hxx
@@ -177,7 +177,7 @@ FixedImagePyramidBase<TElastix>::WritePyramidImage(const std::string & filename,
   configuration.ReadParameter(doCompression, "CompressResultImage", 0, false);
 
   /** Do the writing. */
-  log::to_stdout("  Writing fixed pyramid image ...");
+  log::info("  Writing fixed pyramid image ...");
 #ifndef ELX_NO_FILESYSTEM_ACCESS
   try
   {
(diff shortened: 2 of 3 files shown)


Core/ComponentBaseClasses/elxResamplerBase.hxx
```diff
@@ -395,7 +395,7 @@ ResamplerBase<TElastix>::WriteResultImage(OutputImageType *   image,
   /** Do the writing. */
   if (showProgress)
   {
-    log::to_stdout("  Writing image ...");
+    log::info("  Writing image ...");
```
N-Dekker, line 398: Thank you for your pull request, @jakeforster However, I'm not sure. These three `log::to_stdout("  Writing image ...")` calls were really there to ensure that the messages are _only_ printed to the console (not the log file).  They are actually the only three use cases of `log::to_stdout`.

The print statement in `ResamplerBase::WriteResultImage` was originally added by commit  517406dd215679563c5df7b16dddc2fd085cbbf2, July 25, 2008 (before I got involved with elastix):

```c++
xl::xout["coutonly"] << "\n  Writing image ..." << std::endl;
```
https://github.com/SuperElastix/elastix/blob/517406dd215679563c5df7b16dddc2fd085cbbf2/src/Core/ComponentBaseClasses/elxResamplerBase.hxx#L297

`"coutonly"` suggests that it should _not_ go to the log file.


However, there could certainly be room for improvement. @jakeforster What problem exactly do you want to solve with your pull request?

jakeforster: Hi Niels, the problem is that these messages do not respect the log level. If the user sets the log level to warn or error, it is presumably because they do not want progress messages like these to be shown.

N-Dekker: @jakeforster Thanks, I see your point.

Your current pull request has the "side effect" that it places those `"  Writing image ..."` messages in the log file, whereas I think they were only meant to inform an interactive user of the progress, while elastix is running.

Instead, as a compromise, we could replace `log::to_stdout` with a new function, `log::info_to_stdout`, which would _only_ use stdout, _and_ only  when the log level is `info` (so the most verbose level). What do you think?

jakeforster: Sounds good to me. Thanks for your time.

N-Dekker: There's another case we may then need to reconsider. `ProgressCommand::PrintProgress` unconditionally prints the progress percentage to `std::cout` as well. I'll have another look later this week. Please remind me, if necessary  😃

N-Dekker:
- Closed, as it is superseded by "my" pull request #1449

Of course, my pull request was very much inspired by yours, so thanks very much @jakeforster !  👍

PR #1450, opened 2026-06-04
ParzenWindowHistogramImageToImageMetric: replace m_JointPDFWindow with m_JointPDFWindowSize

N-Dekker:


Common/CostFunctions/itkParzenWindowHistogramImageToImageMetric.h
```diff
@@ -280,7 +280,6 @@ class ITK_TEMPLATE_EXPORT ParzenWindowHistogramImageToImageMetric
   IncrementalMarginalPDFPointer m_MovingIncrementalMarginalPDFRight{ nullptr };
   IncrementalMarginalPDFPointer m_FixedIncrementalMarginalPDFLeft{ nullptr };
   IncrementalMarginalPDFPointer m_MovingIncrementalMarginalPDFLeft{ nullptr };
-  mutable JointPDFRegionType    m_JointPDFWindow{}; // no need for mutable anymore?
```
N-Dekker, line 283: @mstaring FYI, I think this pull request is just a logical follow-up to your commit 2fc2c495837f64416263fa7c5d1e2bc4e3e061fa from 2013 ("ENH: HUGE merge of the performance branch"). Your 2013 commit made the `Index` component of m_JointPDFWindow unused, so from then, it was only its `Size` that mattered. (And indeed, no need for `mutable` anymore.)

I also noticed that its Size is always equal to `{ m_MovingKernelBSplineOrder + 1, m_FixedKernelBSplineOrder + 1 }`, so it is still a bit redundant, having a data member whose value is based on two other data members  🤷

PR #1451, opened 2026-06-05
WIP: PERF: Accumulate joint histograms in ThreadedComputePDFs

N-Dekker:
Inspired by a suggestion from Bradley Lowekamp (@blowekamp), June 2, 2026, at https://discourse.itk.org/t/8x-slower-registration-with-itk-elastix-python-api-vs-elastix-cli-minimal-reproducible-example/7736/37

----
A major performance improvement (10% to 40%) of elastix.exe was observed on the benchmark data set from https://discourse.itk.org/t/8x-slower-registration-with-itk-elastix-python-api-vs-elastix-cli-minimal-reproducible-example (by @NicoKiaru). However, this PR might cause slightly different results, because it may change the order of additions of floating point numbers, during accumulation.

----

Update, July 15, 2026: I feel a bit uncertain now about the performance claims that I did above here. I would have to redo the benchmark, to double-check. @blowekamp FYI

### DIFF
--- a/Common/CostFunctions/itkParzenWindowHistogramImageToImageMetric.hxx
+++ b/Common/CostFunctions/itkParzenWindowHistogramImageToImageMetric.hxx
@@ -195,7 +195,7 @@ ParzenWindowHistogramImageToImageMetric<TFixedImage, TMovingImage>::InitializeHi
    */
   m_JointPDF = JointPDFType::New();
   m_JointPDF->SetRegions(JointPDFSizeType{ m_NumberOfMovingHistogramBins, m_NumberOfFixedHistogramBins });
-  m_JointPDF->Allocate();
+  m_JointPDF->AllocateInitialized();
 
   if (this->GetUseDerivative())
   {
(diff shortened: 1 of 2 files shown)


Common/CostFunctions/itkParzenWindowHistogramImageToImageMetric.hxx
```diff
@@ -195,7 +195,7 @@ ParzenWindowHistogramImageToImageMetric<TFixedImage, TMovingImage>::InitializeHi
    */
   m_JointPDF = JointPDFType::New();
   m_JointPDF->SetRegions(JointPDFSizeType{ m_NumberOfMovingHistogramBins, m_NumberOfFixedHistogramBins });
-  m_JointPDF->Allocate();
+  m_JointPDF->AllocateInitialized();
```
N-Dekker, line 198: Note to myself: It might not be sufficient to zero-initialize the values here in InitializeHistograms(). Maybe ComputePDFs should (also) do `m_JointPDF->FillBuffer(0.0)` before calling `m_Threader->SetSingleMethodAndExecute`.

PR #1457, opened 2026-07-09
STYLE: Add PUBLIC to target_link_libraries and elastix_link_itk calls

N-Dekker:
Using Notepad++, Replace in Files, doing

  Find what: `(target_link_libraries\([^ \r\n]+) `
  Replace with: `\1 PUBLIC `
  Find what: `(target_link_libraries\([^ \r\n]+)$`
  Replace with: `\1 PUBLIC`
  Find what: `(elastix_link_itk\([^ \r\n]+) MODULES`
  Replace with: `\1 PUBLIC MODULES`
  Filters: `CMakeLists*.txt`
  (*) Regular expression

It appears clearer, and more modern, to explicitly specify the scope (by adding the `PUBLIC` keyword).

Doing so would prevent a CMake error when having multiple target_link_libraries calls on the same target, saying:

    The plain signature for target_link_libraries has already been used with
    the target "...".  All uses of target_link_libraries with a target must
    be either all-keyword or all-plain.

As experienced with CMake 4.3.2

- Follow-up to pull request #1457
----

@blowekamp Please check! Do you agree that the `PUBLIC` keyword is preferred over leaving the scope unspecified (implicitly transitive)?

### DIFF
--- a/Common/GTesting/CMakeLists.txt
+++ b/Common/GTesting/CMakeLists.txt
@@ -37,14 +37,14 @@ target_compile_definitions(CommonGTest PRIVATE
 )
 
 if (ITK_VERSION VERSION_LESS_EQUAL 5.4.6)
-  target_link_libraries(CommonGTest
+  target_link_libraries(CommonGTest PUBLIC
     GTest::GTest
     GTest::Main
     ${ELASTIX_ITK_EXECUTABLE_LIBRARIES}
     elastix_lib
   )
 else()
-  target_link_libraries(CommonGTest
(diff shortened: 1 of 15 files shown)


blowekamp:
Look good.

Ideally further work would include more specific specification on which dependencies libraries are using in the header ( PUBLIC ) vs the libraries whose headers are only used in compiled files ( PRIVATE ). And there is also the potential case where a header/library is only used in a header file but not compiled by the library ( INTERFACE ).

PR #1463, opened 2026-08-11
Take the IMPACT backend from ITKIMPACT

vboussot:
The IMPACT losses, `ModelConfiguration`, online (Jacobian) inference, the vector-image interpolator and the Static feature-map generation come from the [ITKIMPACT](https://github.com/InsightSoftwareConsortium/ITKIMPACT) remote module, a one-way dependency Elastix to ITKIMPACT. The local copies (`ImpactLoss.h`, `ImpactTensorUtils.*`,
`itkBSplineInterpolateVectorImageFunction.*`, `itkImpactModelConfiguration.h`) are removed, and the backend's `ModelConfiguration` source is compiled into the Impact component so its torch-free public header stays Python-wrappable.

**Building.** With `USE_ImpactMetric` ON, `ITKIMPACT_INCLUDE_DIR` must point at the ITKIMPACT `include/` directory. Both the `ImpactMetric` component and `CommonGTest` receive it, so `elxImpactMetricGTest.cxx` resolves the backend headers. `ImpactWriteFeatureMaps` becomes `ImpactFeatureMapOutputDirectory`.

**Patch geometry** follows the backend's. `PatchPoint()` sends each offset through the fixed image's direction matrix, one column per image axis, and the sampling callbacks receive the patch tensor shape from `itk::Impact::PatchTensorShape()`. This is the index-to-physical map ITKIMPACT applies in its v4 threader and in `ImageToTensorFilter`, so both modes describe the same neighbourhood.

**CI** provisions LibTorch with `pip install torch` on all three OSes, the same mechanism the backend uses. torch 2.11 is pinned rather than 2.12 because it is the last series still shipping CUDA cu128, the widest GPU coverage for the release: Blackwell plus older cards, driver floor R570, where 2.12 offers only cu126/cu129/cu130. The `download_libtorch` / `build_libtorch` actions are no longer referenced, and `python-version` moves to 3.11 for the
pip step.


thewtex (approved):

Components/Metrics/Impact/CMakeLists.txt
```diff
@@ -2,17 +2,27 @@ ADD_ELXCOMPONENT(ImpactMetric OFF PLUGIN
...
- ImpactTensorUtils.hxx
- itkBSplineInterpolateVectorImageFunction.h
- itkBSplineInterpolateVectorImageFunction.hxx
  itkImpactImageToImageMetric.h
  itkImpactImageToImageMetric.hxx
- itkImpactModelConfiguration.h
  )
 
 if(USE_ImpactMetric)
     find_package(Torch REQUIRED)
```
N-Dekker, line 10: Maybe for a follow-up: would it be possible to add a minimum version to the `find_package(Torch REQUIRED)` call?

More specifically, I'm trying out your PR locally now, and I wonder if my LibTorch 2.8 is still good enough. (I have no problem upgrading to LibTorch 2.11, just wondering!)

.github/workflows/ElastixGitHubActions.yml
```diff
@@ -67,22 +64,35 @@ jobs:
...
+        # source (which the macOS job did, taking ~1-2h). torch 2.11 is used across all three
+        # OSes so the CI compiles the ITKIMPACT backend against the same torch as the release
+        # (publish-elastix.yml uses 2.11 because it is the last series that still ships CUDA
+        # cu128 -- broad GPU coverage incl. Blackwell). Export Torch_DIR for find_package(Torch)
+        # and TORCH_LIB_DIR for the Windows runtime DLL search (Linux/macOS resolve via RPATH).
+        shell: bash
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install "torch==2.11.*" --index-url https://download.pytorch.org/whl/cpu
+          python - <<'PY' >> "$GITHUB_ENV"
```
N-Dekker, line 91: Very clever way of adding environment variables! 👍 I just learned that this is a way to pass a "here document" to Python 🤓  Instead of 'PY', I guess any arbitrary string would work as delimiter, right? Anyway, OK!

vboussot: Haha, with Claude Code I learn something new every day too 😄 And yes, PY is just the here document delimiter here, so any suitable matching delimiter like EOF or PYTHON would work.

.github/workflows/publish-elastix.yml
```diff
@@ -101,53 +110,62 @@ jobs:
           cmake-build-type: ${{ matrix.cmake-build-type }}
           vcvars: ${{ matrix.vcvars64 }}
 
-      - name: Cleanup libtorch (CPU) (Linux)
+      - name: Cleanup CPU torch (Linux)
         if: matrix.os == 'ubuntu-22.04'
         run: |
           mv elastix-build/bin/libImpactMetric.so elastix-build/bin/libImpactMetricCpu.so
-          rm -rf libtorch
+          python -m pip uninstall -y torch
```
N-Dekker, line 117: Just for my understanding, why would you _uninstall_ torch?

vboussot: I was running into disk-space issues on the runner, which is also why I clear some things at the beginning of the job. The rm -rf libtorch and pip uninstall -y torch here are mainly to free up space before the next build/test step.

N-Dekker: OK, thanks, that makes sense 👍

Components/Metrics/Impact/CMakeLists.txt
```diff
@@ -2,17 +2,27 @@ ADD_ELXCOMPONENT(ImpactMetric OFF PLUGIN
...
+    # the missing setting that caused it.
+    if(NOT EXISTS "${ITKIMPACT_INCLUDE_DIR}/ImpactLoss.h")
+        message(FATAL_ERROR
+            "USE_ImpactMetric is ON, so the ITKIMPACT backend is required, but "
+            "ITKIMPACT_INCLUDE_DIR does not point at it: '${ITKIMPACT_INCLUDE_DIR}'. "
+            "Clone https://github.com/InsightSoftwareConsortium/ITKIMPACT and set "
+            "ITKIMPACT_INCLUDE_DIR to its include/ directory.")
+    endif()
+    target_include_directories(ImpactMetric PRIVATE ${ITKIMPACT_INCLUDE_DIR})
+    target_sources(ImpactMetric PRIVATE ${ITKIMPACT_INCLUDE_DIR}/../src/itkModelConfiguration.cxx)
```
N-Dekker: This works as long as ITKIMPACT only has one CXX file (itkModelConfiguration.cxx), of course. OK for now, but it might need to be extended in the future. No problem  👍

Components/Metrics/Impact/elxImpactMetric.h
```diff
@@ -292,7 +293,7 @@ class ITK_TEMPLATE_EXPORT ImpactMetric
 private:
   elxOverrideGetSelfMacro;
 
-  std::vector<itk::ImpactModelConfiguration>
+  std::vector<itk::ModelConfiguration>
```
N-Dekker: I see, `itk::ImpactModelConfiguration` is renamed to `itk::ModelConfiguration`. OK for now, but in the future we might need to place it in an IMPACT specific namespace, or restore the "Impact" prefix, to avoid naming conflicts in the `itk` namespace. OK for now 👍

vboussot: You’re right. I went back to ModelConfiguration when moving it to the remote module, but the potential naming conflict in the itk namespace is real. I can restore the ImpactModelConfiguration name

N-Dekker: > I can restore the ImpactModelConfiguration name

Do you want to do it now? You may also rename `ModelConfiguration` later, after this PR is merged. No problem.

Components/Metrics/Impact/elxImpactMetric.hxx
```diff
@@ -198,13 +198,17 @@ ImpactMetric<TElastix>::GenerateModelsConfiguration(unsigned int level,
...
+      // overlap is 0; the backend precomputes the online patch index unconditionally
+      // (only meaningful when patch and voxel dimensions match, i.e. non-Static).
       modelsConfiguration.emplace_back(modelsPathVec[i],
                                        modelsDimensionVec[i],
                                        numberOfChannelsVec[i],
                                        patchSizeVecByModel[i],
                                        voxelSizeVecByModel[i],
+                                       /*overlap*/ 0u,
                                        GetBooleanVectorFromString(layersMaskVec[i], false),
-                                       mode == "Static",
```
N-Dekker, line 207: I see, you removed the `isStatic` parameter. Just for my understanding, the ImpactMode can still be either "Jacobian" or "Static", right? But it's no longer relevant to the ModelConfiguration constructor...?

Components/Metrics/Impact/elxImpactMetric.hxx
```diff
@@ -437,9 +441,10 @@ ImpactMetric<TElastix>::BeforeEachResolution()
 
   if (mode == "Static")
   {
-    std::string writeFeatureMapsStr = "false";
-    configuration.ReadParameter(writeFeatureMapsStr, "ImpactWriteFeatureMaps", this->GetComponentLabel(), level, 0);
-    if (writeFeatureMapsStr != "false")
+    std::string writeFeatureMapsStr = "";
+    configuration.ReadParameter(
+      writeFeatureMapsStr, "ImpactFeatureMapOutputDirectory", this->GetComponentLabel(), level, 0);
```
N-Dekker: Minor nitpick: You might consider renaming `writeFeatureMapsStr` (maybe to `featureMapOutputDirectory`), now that its meaning has changed. (No show-stopper to me, no problem.)

Components/Metrics/Impact/elxImpactMetric.hxx
vboussot: Yes, exactly. ImpactMode is still either "Jacobian" or "Static".

The only change is that ModelConfiguration no longer needs the isStatic flag. The mode is still handled on the elastix side

Components/Metrics/Impact/elxImpactMetric.h
vboussot: It’s done now, so that’s one less thing to think about later 😄

.github/workflows/publish-elastix.yml
```diff
@@ -25,8 +23,6 @@ jobs:
             libs:
               - elastix-build/bin/libelx-ANNlib.so
           - os: windows-2022
-            libtorch-cpu-url: "https://download.pytorch.org/libtorch/cpu/libtorch-win-shared-with-deps-2.8.0%2Bcpu.zip"
```
N-Dekker, line 28: I guess, the previous approach of downloading libtorch from https://download.pytorch.org/libtorch/cpu (rather than doing `pip install torch`) still remains useful for the Windows Debug version ("libtorch-win-shared-with-deps-debug-2.11.0+cpu.zip"), when building a Debug configuration of elastix with IMPACT, right? Anyway, no problem, just something to keep in mind  😇

Components/Metrics/Impact/elxImpactMetric.hxx
```diff
@@ -190,21 +190,25 @@ ImpactMetric<TElastix>::GenerateModelsConfiguration(unsigned int level,
...
+      // Backend itk::ImpactModelConfiguration ctor order: (..., voxelSize, overlap, layersMask,
+      // useMixedPrecision). Elastix does not tile via the backend's Static path, so
+      // overlap is 0; the backend precomputes the online patch index unconditionally
+      // (only meaningful when patch and voxel dimensions match, i.e. non-Static).
       modelsConfiguration.emplace_back(modelsPathVec[i],
                                        modelsDimensionVec[i],
                                        numberOfChannelsVec[i],
                                        patchSizeVecByModel[i],
                                        voxelSizeVecByModel[i],
+                                       /*overlap*/ 0u,
```
N-Dekker: For other use-cases, the `overlap` argument would be non-zero, right? I'm just wondering, because when it is always just zero, it doesn't need to be an extra parameter  🤷  No problem, though, just curious.

Components/Metrics/Impact/itkImpactImageToImageMetric.h
```diff
@@ -21,10 +21,17 @@
 
 #include "itkAdvancedImageToImageMetric.h"
 #include "itkBSplineInterpolateImageFunction.h"
-#include "itkBSplineInterpolateVectorImageFunction.h"
+// Vector-image interpolator consumed from the ITKIMPACT backend (same per-component
+// B-spline interpolation as the former local itkBSplineInterpolateVectorImageFunction).
+#include "itkInterpolateVectorImageFunction.h"
```
N-Dekker, line 26: The name "itkInterpolateVectorImageFunction.h" might (also) eventually cause naming conflicts with ITK (which currently just has a file named "itkInterpolateImageFunction.h")

No problem for now, just something I noticed.

Components/Metrics/Impact/itkImpactImageToImageMetric.h
```diff
@@ -21,10 +21,17 @@
...
 #include "itkImpactModelConfiguration.h"
-#include "ImpactTensorUtils.h"
-#include "ImpactLoss.h"
+#include "itkImpactModelConfigurationDetail.h"
```
N-Dekker: No problem, but I would expect a file named "itk...Detail.h" to be an internal implementation detail (of ITKIMPACT, in this case), that does not need to be included _directly_ by client code (elastix, in this case).

Just a nitpick that you may ignore  😃

Components/Metrics/Impact/itkImpactImageToImageMetric.h
```diff
@@ -21,10 +21,17 @@
...
+// Static feature-map generation (itk::ImageToFeaturesMap) and online inference
+// (itk::Impact::GenerateOutputs/AndJacobian) are consumed from the ITKIMPACT backend.
+#include "itkImageToFeaturesMap.h"
+#include "itkImageToFeaturesMapInternals.h"
```
N-Dekker: Similar nitpick here: I would expect a file named "itk...internals.h" to be an internal implementation detail (of ITKIMPACT), that does not need to be included directly by client code (elastix). No problem though.

N-Dekker:
@vboussot I'm ready to merge your PR now! Is it also ready to you? I guess so, otherwise you may convert to draft 😺

vboussot:
@N-Dekker Thanks for these, all good catches 👍

ModelConfiguration is back to itk::ImpactModelConfiguration, and while I was at it I moved the internal tiling machinery (accumulator, patch grid, blend windows) into itk::Impact.

I did leave InterpolateVectorImageFunction and the ImageToTensorFilter / TensorToImageFilter pair alone, though. They're a generic ITK/torch bridge and a generic vector image interpolator rather than IMPACT concepts, so the resemblance to itkInterpolateImageFunction.h is on purpose 🙂

Your nitpick on the internal headers was spot on, and easier to fix than expected. elastix used no symbol from either, and the pimpl is held in a shared_ptr, so the complete type is never needed on this side. Both includes are gone.

And overlap turned out to be a real gap rather than a dead parameter, so thanks for pulling on that thread. It drives the blending when patches are reassembled, and elastix does tile in Static mode, which is what a machine short on VRAM has to fall back to. Hardcoding 0 meant those feature maps were stitched with no blending at all. It's now an ImpactOverlap parameter, per axis like PatchSize, defaulting to a quarter of each patch axis.

Components/Metrics/Impact/itkImpactImageToImageMetric.h
N-Dekker: Resolved by your commit 83b0cdece5753d98ace9af6b3a4bf83406f3dd93, thanks!

Components/Metrics/Impact/itkImpactImageToImageMetric.h
N-Dekker: Also resolved by your commit 83b0cdece5753d98ace9af6b3a4bf83406f3dd93, thanks again!

Components/Metrics/Impact/elxImpactMetric.hxx
```diff
@@ -199,17 +217,19 @@ ImpactMetric<TElastix>::GenerateModelsConfiguration(unsigned int level,
...
                                        numberOfChannelsVec[i],
                                        patchSizeVecByModel[i],
                                        voxelSizeVecByModel[i],
-                                       /*overlap*/ 0u,
+                                       overlapVecByModel[i].front(),
                                        GetBooleanVectorFromString(layersMaskVec[i], false),
                                        useMixedPrecision);
+      // The reassembly reads the per-axis overlap, which the constructor can only broadcast
+      // from its scalar; set it explicitly so an anisotropic patch keeps its own per-axis one.
+      modelsConfiguration.back().SetOverlaps(overlapVecByModel[i]);
```
N-Dekker: This looks somewhat redundant to me: first `overlapVecByModel[i].front()` is passed to the appended configuration, then `overlapVecByModel[i]` is _also_ set into the configuration that is just appended. That's the way you mean it to be, right? Just double-checking  🤓

vboussot:
Good catch, it was redundant. The constructor took a scalar and broadcast it, so the SetOverlaps right after overwrote what it had just done. And nothing ever read the scalar back.

Fixed in the backend instead of here: the configuration now takes the overlap per axis, like patchSize and voxelSize next to it, and elastix passes it in one call.

N-Dekker:
Thanks for addressing my overlap comment, @vboussot. Now your pull request is _really_ ready, right?  😃

vboussot:
Yes, ready 😃 Thanks for the review !

N-Dekker (approved):
🎉

PR #1468, opened 2026-09-07
ENH: Support ITK transform file (txt, tfm, h5) as initial transform file

N-Dekker:
Allowed the user to specify an ITK transform file as initial transform file, both through the library interface (ElastixRegistrationMethod.SetInitialTransformParameterFileName) and the executable command-line interface ("-t0").

- Very much inspired by pull request #1342, by Valentin Boussot (@vboussot)


Core/Configuration/elxConfiguration.cxx
```diff
@@ -252,25 +252,46 @@ Configuration::Initialize(const CommandLineArgumentMapType & _arg)
...
+    /** Connect the parameter file reader to the interface. */
+    m_ParameterMapInterface.SetParameterMap(
+      AddDataFromExternalTransformFile(m_ParameterFileName, parameterFileParser->GetParameterMap()));
   }
-  catch (const itk::ExceptionObject & excp)
+  else
   {
-    log::error(std::ostringstream{} << "ERROR: when reading the parameter file:\n" << excp);
-    return 1;
+    if (const auto transform = TransformIO::Read(m_ParameterFileName))
```
vboussot: TransformIO::Read calls reader->Update() without a try/catch, so it throws
instead of returning null. A missing file, a wrong content and an unknown
extension all throw. The exception leaves Configuration::Initialize and is
caught only in main, while the branch above returns 1.

Core/Configuration/elxConfiguration.cxx
```diff
@@ -252,25 +252,46 @@ Configuration::Initialize(const CommandLineArgumentMapType & _arg)
     return 1;
   }
 
-  const auto parameterFileParser = itk::ParameterFileParser::New();
-
-  /** Read the ParameterFile. */
-  parameterFileParser->SetParameterFileName(m_ParameterFileName);
-  try
+  if (itk::ParameterFileParser::HasParameterFileNameExtension(m_ParameterFileName))
```
vboussot: BasicFileChecking now runs only for .txt and .toml, so the existence check is skipped for any other name. "elastix -p params.prm" used to print "should be a text file (*.txt) or a TOML file (*.toml)" and return 1. Now it prints an HDF5 error block and a terminating exception. This function also serves -p and -tp,
not only -t0.

Would it work to try the parameter file first, and read the transform only if that fails, inside a try/catch?

Common/ParameterFileParser/itkParameterFileParser.cxx
```diff
@@ -551,4 +551,14 @@ ParameterFileParser::ConvertToParameterMap(const std::string & text) -> Paramete
 }
 
 
+bool
+ParameterFileParser::HasParameterFileNameExtension(const std::string & fileName)
+{
+  static constexpr std::string_view parameterFileNameExtensions[] = { ".txt", ".toml" };
```
vboussot: ITK writes the same legacy format to .tfm and .txt. Since .txt is claimed here,
a transform saved as .txt can never be read back by -t0.

Common/ParameterFileParser/itkParameterFileParser.cxx
```diff
@@ -551,4 +551,14 @@ ParameterFileParser::ConvertToParameterMap(const std::string & text) -> Paramete
...
+
+  return std::count(std::begin(parameterFileNameExtensions),
```
vboussot: std::count needs <algorithm>, not included here. std::find() != std::end() would also stop at the first match

Core/Configuration/elxConfiguration.cxx
```diff
@@ -299,17 +299,20 @@ Configuration::InitializeWithInitialTransform(const std::string & initialTransfo
 
   try
   {
-    log::info("Reading the initial transform parameters from file ...\n");
-
+    const auto transform = TransformIO::Read(m_ParameterFileName);
+    m_ParameterMapInterface.SetParameterMap(
+      { { "ITKTransformFixedParameters", elx::Conversion::ToVectorOfStrings(transform->GetFixedParameters()) },
```
vboussot: `TransformIO::Read` can hand back a null pointer: it ends in `return transformList->empty() ? nullptr : transformList->front();`, and the `assert` just above it is compiled out in a release build. So a file that ITK reads without complaining but that holds no transform, say an Insight legacy `.txt` with nothing but its header, gets dereferenced right here. A null check falling through to the parameter file branch would cover it.

Core/Configuration/elxConfiguration.cxx
```diff
@@ -299,17 +299,20 @@ Configuration::InitializeWithInitialTransform(const std::string & initialTransfo
...
+        { "ITKTransformParameters", elx::Conversion::ToVectorOfStrings(transform->GetParameters()) },
+        { "ITKTransformType", { transform->GetTransformTypeAsString() } },
+        { "Transform", { elx::TransformIO::ConvertITKNameOfClassToElastixClassName(transform->GetNameOfClass()) } } });
+  }
+  catch (const std::exception &)
+  {
     // Read the parameter map from the parameter file and add data from a possible external transform file.
     m_ParameterMapInterface.SetParameterMap(AddDataFromExternalTransformFile(
       m_ParameterFileName, itk::ParameterFileParser::ReadParameterMap(m_ParameterFileName)));
```
vboussot: Reading the parameter file from inside the catch also changes what the user sees when both reads fail. Today a bad `-t0` file makes this return false, and `ReadInitialTransformFromFile` turns that into `ERROR: Reading initial transform parameters failed: <file name>`. Now `ReadParameterMap` throws from inside the catch, so the message becomes `the file X should be a text file (*.txt) or a TOML file (*.toml)`, which no longer says the file was the initial transform. The `Reading the initial transform parameters from file ...` log line is gone as well.

The `HasParameterFileNameExtension` you add in this PR would let you pick the reader from the extension and keep those messages. Right now it is only used by `BasicFileChecking`.

Common/ParameterFileParser/itkParameterFileParser.cxx
vboussot: `std::count` needs `<algorithm>`, and only `<iterator>` is included here, so this compiles through a transitive include. For two extensions `ext == ".txt" || ext == ".toml"` would do, or `std::find`, which stops at the first match.

Core/Configuration/elxConfiguration.cxx
vboussot: Falling back inside the catch also changes what the user gets told when the file is neither an ITK transform nor a parameter file. On main this catches the exception, logs `ERROR: when reading the initial transform parameter file:` with it, and returns false, and `TransformBase::ReadInitialTransformFromFile` then throws `ERROR: Reading initial transform parameters failed: <file name>`. Here `ReadParameterMap` throws from inside the catch, so both of those go away and all that is left is whatever the parameter file parser says, for instance `the file X should be a text file (*.txt) or a TOML file (*.toml)` for an `.h5` that ITK could not read. The `Reading the initial transform parameters from file ...` info line is gone too.

`HasParameterFileNameExtension`, which you add in this PR, would let you choose the reader from the extension and keep those messages. At the moment it is only used by `BasicFileChecking`.

Common/ParameterFileParser/itkParameterFileParser.cxx
vboussot: `std::count` needs `<algorithm>`, and only `<iterator>` is included here, so this is relying on a transitive include. For two extensions `ext == ".txt" || ext == ".toml"` would do, or `std::find`, which stops at the first match.

Core/Configuration/elxConfiguration.cxx
```diff
@@ -299,17 +299,19 @@ Configuration::InitializeWithInitialTransform(const std::string & initialTransfo
...
```
vboussot, line 313: When the file is neither an ITK transform nor a parameter file, the user is now told a lot less. On main the catch logs `ERROR: when reading the initial transform parameter file:` together with the exception and returns false, and `TransformBase::ReadInitialTransformFromFile` turns that into `ERROR: Reading initial transform parameters failed: <file name>`. Here `ReadParameterMap` throws from inside the catch, so both messages are gone and what is left is whatever the parameter file parser says, for instance `the file X should be a text file (*.txt) or a TOML file (*.toml)` for an `.h5` that ITK failed to read. The reason the transform read failed is swallowed as well, and the `Reading the initial transform parameters from file ...` info line has disappeared.

Wrapping the fallback in its own try/catch that keeps the old log and `return false` would restore all of that.

